PR #42, opened 2023-09-01
fix: entity_id normative language about how the final metadata is obtained

peppelinux:
This PR improves the normative text regarding the client identifier when it is set to openid federation.
The previous text was misleading as it mentioned the metadata taken from the entity statement, while the final metadata must be derived from the trust chain.

This PR was manually migrated from bitbucket, where the original one is available at the following link:
https://bitbucket.org/openid/connect/pull-requests/578


Sakurann (approved):

selfissued (approved):
LGTM

rohe (approved):

PR #43, opened 2023-09-01
Support for Federations and Trust Schemes, removal of termsOfUse

peppelinux:
Reading the specs I’ve realized that the text pertaining the federation/trust-schemes can be simplified by giving simple indications and normative language to the implementers.

During the review phase, additional comments asked for the removal of `termsOfUse`. This was accomplished in the current PR.

This was manually migrated from bitbucket, where the original one is available at the following link:
https://bitbucket.org/openid/connect/pull-requests/587


openid-4-verifiable-presentations-1_0.md
```diff
@@ -828,40 +828,17 @@ The following is a non-normative example of a set of static configuration values
 }
 ```
 
-## Support for Federations/Trust Schemes
+## Support for Federations and Trust Schemes
 
-Often Verifiers will want to request Verifiable Credentials from a Credential Issuer who is a participant of a federation, or adheres to a known trust scheme, rather than from a specific Credential Issuer, for example, a "BSc Chemistry Degree" Credential from the hypothetical "eduCreds" trust scheme rather than from a specifically named university.
+The Verifier who wants to obtain Verifiable Credentials issued by Credential Issuer who is a participant of a federation, or adheres to a known trust scheme, MUST create a `presentation_definition` that includes a known property within such federation, that allows to distinguish the request according to it.
```
peppelinux, line 833: I definitively have to change this text

peppelinux: Done

openid-4-verifiable-presentations-1_0.md
```diff
@@ -828,40 +828,17 @@ The following is a non-normative example of a set of static configuration values
 }
 ```
 
-## Support for Federations/Trust Schemes
+## Support for Federations and Trust Schemes
```
peppelinux, line 831: I'd change Trust Schemes to Trust Models if you agree

I heard some comments about the concrete meaning of the term trust scheme

openid-4-verifiable-presentations-1_0.md
```diff
@@ -828,40 +828,17 @@ The following is a non-normative example of a set of static configuration values
...
 
-To facilitate this, federations will need to determine how a Credential Issuer can indicate in a Verifiable Credential that they are a member of one or more federations/trust schemes. Once this is done, the Verifier will be able to create a `presentation_definition` that includes this filtering criteria. This will enable the Wallet to select all the Verifiable Credentials that match this criteria and then by some means (for example, by asking the user) determine which matching Verifiable Credential to return to the Verifier. Upon receiving this Verifiable Credential, the Verifier will be able to call its federation API to determine if the Credential Issuer is indeed a member of the federation/trust scheme that it says it is.
+To facilitate this, federations will need to determine how a Credential Issuer can indicate in a Verifiable Credential that they are a member of one or more federations. Once this is done, the Verifier will be able to create a `presentation_definition` that includes this filtering criteria.
```
peppelinux, line 835:
```suggestion
To facilitate this, federations need to determine how a Credential Issuer indicates in a Verifiable Credential that it is a member of one or more federations. Once this is done, the Verifier will be able to create a `presentation_definition` that includes this filtering criteria. 
```

openid-4-verifiable-presentations-1_0.md
Sakurann: why not trust framework? what is the distinction/definition of trust scheme vs trust model vs trust framework?

openid-4-verifiable-presentations-1_0.md
Sakurann, line 835: I think an example of how to use a filter property would be helpful?
so the idea is that the credential contains a claim that expresses a trust framework that the issuer of that credential belongs to and verifier must know that claim and put restrictions on that claim using PE, correct? if so, explaining this a little more would help.

Sakurann (changes requested):
agree with the direction, think the alternatively proposed mechanism and its rationale should be explained in more detail

openid-4-verifiable-presentations-1_0.md
peppelinux: I'm in favor of Trust Framework, I'm doing this change

peppelinux: Done c72b02f0791d0de05d5428a39c3672b1e4f58636

openid-4-verifiable-presentations-1_0.md
peppelinux: I agree, I'll try to put in the form of comments some proposal before the commit for your kindly review

openid-4-verifiable-presentations-1_0.md
```diff
@@ -828,40 +828,17 @@ The following is a non-normative example of a set of static configuration values
...
 
-To facilitate this, federations will need to determine how a Credential Issuer can indicate in a Verifiable Credential that they are a member of one or more federations/trust schemes. Once this is done, the Verifier will be able to create a `presentation_definition` that includes this filtering criteria. This will enable the Wallet to select all the Verifiable Credentials that match this criteria and then by some means (for example, by asking the user) determine which matching Verifiable Credential to return to the Verifier. Upon receiving this Verifiable Credential, the Verifier will be able to call its federation API to determine if the Credential Issuer is indeed a member of the federation/trust scheme that it says it is.
+To facilitate this, federations need to determine how a Credential Issuer indicates in a Verifiable Credential that it is a member of one or more federations. Once this is done, the Verifier will be able to create a `presentation_definition` that includes this filtering criteria.
```
Sakurann, line 835: what if there is no explicit claim in the credential that conveys the federation that a credential belogs to? I am not sure the verifier can communicate using PE that "please return this credential if after obtaining the trust chain using the iss value in the credential, you are certain that the credential belongs to Federation A"?

peppelinux: I'd not mention any trust chain or policy since these devices are not required here but Just a shared and agreed trust framework where many entities may find themselves under the same regulation

Sakurann: the text you have written mandates/assumes that issuers put inside a credential information about which federation that issuer belongs to, I just do not think that's the case. there are cases where `iss` in the credential is HTTPS and a file found by appending a `.well-known` is sufficient. in other places, `iss` is a DID and metadata with the federation information is hosted at the service endpoint obtained from a DID Document.

I suggest you file an issue first. and this PR is refactored to remove dependency on `termsofuse` property.

David-Chadwick:
I think this PR is detrimental to implementors because it is highly likely now that the VC DMv2 will contain the Terms of Use to hold information such as trusted federations.
I believe this PR was first proposed at the time when it was being proposed to remove ToU from VCDMv2 because it was not thought that sufficient people were implementing it. But since then, it has been reported that at least 11 wallets and 3 issuers are using the ToU to store trust information and there is currently a PR (authored by myself) which includes an example from ESBI (who use it in their eco-system).
I therefore suggest that this PE is deleted and that the ToU text remains

tlodderstedt:
I see the addition of data about the trust framework as a workaround to ease the processing of a credential. I agree with @Sakurann that the trust framework should be determined from the issuer. I think we need to discuss this topic further. Is there a corresponding issue?

David-Chadwick:
Just to confirm that the ToU property is now part of the VCDMv2 draft, and the text describing the ToU property indicates that it is inserted by the issuer to provide information to verifiers about the trust framework(s) that the issuer is a member of.

Note that there are various ways that an issuer can provide information to verifiers:

1. It can be published in the issuer's metadata
2. It can be in the credential itself
3. It can be in the protocol communication with the issuer

We currently use all three methods, and sometimes we use more than one method for the same data.

Note that the trust framework confirms the veracity of the information that the issuer releases, although it could in principle publish all the issuer's information, so that the issuer does not need to release anything!

tlodderstedt:
> OID4VP is format-agnostic. Whatever we do, it should work regardless of the credential format.

I agree. Format specific rules must go into credential specific profiles.

David-Chadwick:
I also agree. All credentials should be extensible. For W3C VCs, the ToU is an extension point. Other credential formats can also add extensions to indicate the trust frameworks the issuer is a member of.

David-Chadwick:
There is value in an informative annex giving an example of how trust frameworks can be added to a specific credential format. Using W3C VCs as the example will provide guidance to other credential formats how they might achieve this.

jogu:
I think both Torsten & Kristina have suggested we should have an issue (possibly more than one issue?) to discuss these issues and agree on the approach first.

I'd hence propose closing this PR in a week unless anyone strongly objects, and we can continue any discussion on issues.

David-Chadwick:
Yes please open the other issues first, and then we can close this PR and point to the new issues.

jogu:
Closing as per previous comments. I don't feel sufficiently well informed to raise issues about the different problems, please feel free to raise them as issues yourselves.

PR #50, opened 2023-09-26
Remove `client_id_scheme` from Verifier Metadata

cobward:
There is currently some ambiguity around `client_id_scheme`. This parameter appears in the request object AND the client metadata, which also appears in the request object. There is no text in the specification to define what a wallet should do if it receives a rquest object where these two values do not match.

As I see it there are 3 ways forwards:
1. Remove `client_id_scheme` from Verifier Metadata.
2. Remove `client_id_scheme` from the authz request object, AND make it so that either `client_metadata` or `client_metadata_uri` are always provided in the request object.
3. Add text to describe what a wallet should do if it receives a request object with mismatching `client_id_scheme`.

I have raised this PR with option 1.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -762,9 +762,6 @@ This specification defines the following new metadata parameters according to [@
 : REQUIRED. An object defining the formats and proof types of Verifiable Presentations and Verifiable Credentials that a Verifier supports. For specific values that can be used, see (#alternative_credential_formats).
 Deployments can extend the formats supported, provided Issuers, Holders and Verifiers all understand the new format.
 
-`client_id_scheme`:
-: OPTIONAL. JSON String identifying the Client Identifier scheme. The value range defined by this specification is `pre-registered`, `redirect_uri`, `entity_id`, `did`. If omitted, the default value is `pre-registered`. 
-
 # Verifier Attestation JWT {#verifier_attestation_jwt}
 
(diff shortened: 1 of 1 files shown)


Sakurann (approved):
problem statement and option 1 as a solution makes sense to me.

tlodderstedt (approved):

awoie (approved):

bc-pi (approved):

peppelinux:
> Remove client_id_scheme from the authz request object, AND make it so that either client_metadata or client_metadata_uri are always provided in the request object.

This is in conflicts with OpenID Federation or any other trust model that requires the metadata be provided by a trusted third party

This PR introduce an important breaking change to some current implementations that uses OpenID Federation for the wallet ecosystem

With this comment I would not say that client_id_scheme must be keep, if removed the text/PR should give guidance about the implementations that requires the metadata with different mechanisms than the authz request object

`client_metadata_uri` is a strategy where its response cannot be just `application/json` but a JWS as well, where required by the trust model

peppelinux (changes requested):
See https://github.com/openid/OpenID4VP/pull/50#issuecomment-1749341269

Sakurann:
@peppelinux if wallet is able to obtain verifier's metadata using openid federation, it means that wallet already trusted client_id_scheme `entity_id` in the authorization request. so I do not think you need `client_id_scheme` verifier metadata parameter in the openid federation files.

peppelinux:
Thank you for clarification @Sakurann , having not  build the preview I was Just worries about the potential impact, but as you said this PR removes the client id scheme just from the metadata

Approved!

PS = Is there any CD to have the preview a of the build for each PR?

peppelinux (approved):

PR #51, opened 2023-09-27
clarify client id scheme description

Sakurann:
clarify that when certain client id schemes mention redirect_uri, it is only when response mode is not direct_post and when response mode is direct_post, those restrictions apply to response_uri instead of redirect_uri.


openid-4-verifiable-presentations-1_0.md
```diff
@@ -417,7 +417,7 @@ This specification defines the following values for the `client_id_scheme` param
 
 * `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed, the Verifier MAY omit the `redirect_uri` Authorization Request parameter, and all Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
+* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). When Response Mode `direct_post` is used, these restrictions apply to the `response_uri` value instead of `redirect_uri` value.
```
jogu, line 420: I'm not sure the 'MAY' clause is technically a restriction so a different word might be better. I'm struggling to come up with a really good word but 'clauses' may be better?

```suggestion
* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). When Response Mode `direct_post` is used, these clauses apply to the `response_uri` value instead of `redirect_uri` value.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -452,9 +452,9 @@ Body
...
+* `x509_san_dns`: When the Client Identifier Scheme is `x509_san_dns`, the Client Identifier MUST be a DNS name and match a `dNSName` Subject Alternative Name (SAN) [@!RFC5280] entry in the leaf certificate passed with the request. The request MUST be signed with the private key corresponding to the public key in the leaf X.509 certificate of the certificate chain added to the request in the `x5c` JOSE header [@!RFC7515] of the signed request object. The Wallet MUST validate the signature and the trust chain of the X.509 certificate. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter. If the Wallet can establish trust in the Client Identifier authenticated through the certificate (e.g. because the Client Identifier is contained in a list of trusted Client Identifiers) it may allow the client to freely choose the `redirect_uri` value. If the Wallet cannot establish such trust, the FQDN of the `redirect_uri` value MUST match the Client Identifier. When response mode `direct_post` is used, these restrictions apply to `response_uri` value, instead of `redirect_uri` value.
```
Sakurann, line 455:
```suggestion
* `x509_san_dns`: When the Client Identifier Scheme is `x509_san_dns`, the Client Identifier MUST be a DNS name and match a `dNSName` Subject Alternative Name (SAN) [@!RFC5280] entry in the leaf certificate passed with the request. The request MUST be signed with the private key corresponding to the public key in the leaf X.509 certificate of the certificate chain added to the request in the `x5c` JOSE header [@!RFC7515] of the signed request object. The Wallet MUST validate the signature and the trust chain of the X.509 certificate. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter. If the Wallet can establish trust in the Client Identifier authenticated through the certificate (e.g. because the Client Identifier is contained in a list of trusted Client Identifiers) it MAY allow the client to freely choose the `redirect_uri` value. If the Wallet cannot establish such trust, the FQDN of the `redirect_uri` value MUST match the Client Identifier. When response mode `direct_post` is used, these restrictions apply to `response_uri` value, instead of `redirect_uri` value.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -452,9 +452,9 @@ Body
...
+* `x509_san_uri`: When the Client Identifier Scheme is `x509_san_uri`, the Client Identifier MUST be a URI and match a `uniformResourceIdentifier` Subject Alternative Name (SAN) [@!RFC5280] entry in the leaf certificate passed with the request. The request MUST be signed with the private key corresponding to the public key in the leaf X.509 certificate of the certificate chain added to the request in the `x5c` JOSE header [@!RFC7515] of the signed request object. The Wallet MUST validate the signature and the trust chain of the X.509 certificate. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter. If the Wallet can establish trust in the Client Identifier authenticated through the certificate, e.g. because the Client Identifier is contained in a list of trusted Client Identifiers, it may allow the client to freely choose the `redirect_uri` value. If the Wallet cannot establish such trust, the `redirect_uri` value MUST match the Client Identifier. When response mode `direct_post` is used, these restrictions apply to `response_uri` value, instead of `redirect_uri` value.
```
Sakurann, line 457:
```suggestion
* `x509_san_uri`: When the Client Identifier Scheme is `x509_san_uri`, the Client Identifier MUST be a URI and match a `uniformResourceIdentifier` Subject Alternative Name (SAN) [@!RFC5280] entry in the leaf certificate passed with the request. The request MUST be signed with the private key corresponding to the public key in the leaf X.509 certificate of the certificate chain added to the request in the `x5c` JOSE header [@!RFC7515] of the signed request object. The Wallet MUST validate the signature and the trust chain of the X.509 certificate. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter. If the Wallet can establish trust in the Client Identifier authenticated through the certificate, e.g. because the Client Identifier is contained in a list of trusted Client Identifiers, it MAY allow the client to freely choose the `redirect_uri` value. If the Wallet cannot establish such trust, the `redirect_uri` value MUST match the Client Identifier. When response mode `direct_post` is used, these restrictions apply to `response_uri` value, instead of `redirect_uri` value.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: omitting something is a technical restriction, no?

jogu: I'm not sure. I think allowing someone to optionally omit something is technically an addition rather than a restriction. [addition probably isn't quite the right word, hopefully the point still makes sense :) ]

I didn't want to give anyone any wiggle room to argue they they don't have to allow verifiers to omit response_uri.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -417,7 +417,7 @@ This specification defines the following values for the `client_id_scheme` param
 
 * `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed, the Verifier MAY omit the `redirect_uri` Authorization Request parameter, and all Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
+* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). When Response Mode `direct_post` is used, these clauses apply to the `response_uri` value instead of `redirect_uri` value.
```
peppelinux, line 420: this statement is not clear

it is not clear why the client uid must (or is) the same of the redirect uri.
it is not clear why the authorization must not be signed.

tlodderstedt: "it is not clear why the client uid must (or is) the same of the redirect uri."

there is no way to link the client_id to a redirect_uri in this scheme then making both the same

"it is not clear why the authorization must not be signed."

I think we can drop that requirement. It was used in the past (before client id schemes were introduced) to differentiate redirect_uri as client id and other options. With the client id scheme `redirect_uri`, it is no longer needed.

tlodderstedt (changes requested):
suggest to drop the requirement to sign requests for redirect_uri scheme

peppelinux (changes requested):
I think that we should follow Torsten's points here: https://github.com/openid/OpenID4VP/pull/51/files#r1456848468

jogu:
I think https://github.com/openid/OpenID4VP/pull/84/files has solved the main issue this PR was trying to solve.

I suggest closing this PR.

Issue #45, opened 2022-01-20
Advanced flow for OpenID4VP


OIDF-automation:
### Original Reporter: tlodderstedt

The current cross device flow works fine, however it forces the RP to make assumptions about the SIOP/wallet the user might choose to process the SIOP request. 

I propose to investigate an advanced flow that involves an advertisement/discovery step and allows Rp and SIOP to better adjust the flow to each others capabilities. 

The RP could render a QR code that give rise to its capabilities and endpoints. The user then scans that QR code with the wallet of her choice. 

The wallet uses the QR code data and send a request to the RP containing its capabilities and endpoints/identifiers in a direct HTTPS POST request.

In the HTTPS POST response, the RP directly sends with the authentication request data \(tailored for the particular SIOP\). 

As already noted, this flow would allow the RP and the SIOP to tailor the requests to each others capabilities. Also, the QR code could potentially be static and concise.

Sakurann:
at OSW we discussed an alternative approach, which builds up on RFC9101 (passing request by reference):
1) verifier sends an authorization request with a request_uri to the wallet (custom url scheme, etc.)
2) when the wallet obtains the request object at a received request_uri, it sends a POST request with the parameters describing wallet capabilities, maybe wallet attestation etc.
3) flow continues

but I prefer @tlodderstedt 's proposal[ in this comment](https://github.com/openid/OpenID4VP/issues/45#issuecomment-1707474498), mainly because:
- it addresses the wallet invocation problem because verifier advertisement (to borrow Torsten's terms) might use a custom url scheme, but the actual authorization request that matters can be tailored to a wallet specific endpoint (app link/universal link)
-  is cleaner than building up on rfc9101.

Sakurann:
@alenhorvat thank you - I am trying to better understand the flow you linked to: does the flow you point to work only when the wallet initiates the flow? is the assumption that the wallet knows verifier endpoint out of band? how does cross device flow look like? why is there token request from the wallet - what is access token scoped for? why does the description text talk about credential req-res if this is about presentation? Is the Self-issued ID Token of the wallet sent in the first authorize request to the verifier without verifier provided nonce?

alenhorvat:
I believe there are 2 possible scenarios
a) dynamic wallet discovery - in this case a simple POST to an ?authorisation? endpoint with wallet metadata could solve the problem (same behaviour as redirect_uri except that it can accept the wallet metadata and adjust the request object accordingly)
[...]

PR #52, opened 2023-09-27
Advanced flow

tlodderstedt:
This is the first revision of the advanced flow (issue #45). Based on the feedback on the issue, I called the wallet request `Discovery Request`. The verifier may require the wallet to authenticate. It will do so be responding with a HTTP status code 401, the response also provides a nonce to be included in the proof of possession. 

Open Topic: Do we want to support plain JSON presentation requests?


peppelinux:
Plain JSON request doesn't seems to satisfy the eudi wallet requirements but I think that openid4vp Will be used also for other Networks and use cases

The wallet that send a plain JSON request to an RP that doesn't support it should obtain a 400 http error with the error identifier and an error_description

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+```JSON
+{
+    "presentation_request_uri": "verifier.example.com/presentation_request",
+    "context": "register_kyc"
+}
+```
+
+The Discovery Request is either send the Discovery Endpoint of the Wallet or it is rendered as a QR Code and scanned with the Wallet or a camera app. 
+
+Note: the Discovery Request intentionally does not use OAuth parameters, such as a `client_id`. The idea is to allow the Verifier to select the Client Identifier it wants to use for interacting with the Wallet after it has determined the trust mechanisms and domains the wallet supports.
```
peppelinux, line 303:
```suggestion
Note: the Discovery Request intentionally does not use OAuth parameters, such as a `client_id`. The idea is to allow the Verifier to select the Client Identifier it wants to use for interacting with the Wallet after it has determined the trust mechanisms and domains the Wallet supports.   
```

peppelinux: Not in favour of this because the RP must provide it's unique identifier since a wallet solution, or a trust model, may require to not send the wallet instance attestation before having established the trust to the RP

A client_id allows this in both oidc federation and eidas trusted list

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+}
+```
+
+The Discovery Request is either send the Discovery Endpoint of the Wallet or it is rendered as a QR Code and scanned with the Wallet or a camera app. 
+
+Note: the Discovery Request intentionally does not use OAuth parameters, such as a `client_id`. The idea is to allow the Verifier to select the Client Identifier it wants to use for interacting with the Wallet after it has determined the trust mechanisms and domains the wallet supports.   
+
+## Discovery Endpoint
+
+This is an HTTPS endpoint offered by the Wallet. The Discovery Request MUST be sent as a HTTPS GET request or an HTTP redirect to the Discovery Endpoint URL defined in (#wallet-metadata)
```
peppelinux, line 307:
```suggestion
This is an HTTPS endpoint offered by the Wallet. The Discovery Request MUST be sent as an HTTPS GET request or an HTTP redirect to the Discovery Endpoint URL defined in (#wallet-metadata)
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+The HTTPS URL contains a single URI query parameter `discovery` where the value is the JSON object as defined above.
```
peppelinux, line 309:
```suggestion
The HTTPS URL MUST contain a single URI query parameter `discovery` where the value is the JSON object as defined above. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+Presentation Requests MUST be HTTPS POST requests with the "application/json" media type.
+
+The follwowing parameters are defined: 
+
+`context`:
+: A JSON String containing the value of the corresponding Discovery Request's `context` parameter.
+
+`issuer`:
+: A JSON containg a HTTPS URL designating the Issuer URL of the Wallet.
```
peppelinux, line 325:
```suggestion
: A JSON containing an HTTPS URL designating the Issuer URL of the Wallet. 
```

peppelinux: I didn't get the Issuer URL of the Wallet

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+: A JSON String containing the value of the corresponding Discovery Request's `context` parameter.
+
+`issuer`:
+: A JSON containg a HTTPS URL designating the Issuer URL of the Wallet. 
+
+`w_nonce`:
+: A JSON String containg a cryptographic nonce the Verifier MUST use when creating the signed presentation request object. 
+
+`client_assertion`
+: A JSON String containing a wallet attestation along with a proof of posession of the confiration key as defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. This assertion is used to authenticate the Wallet towards the Verifier.
```
peppelinux, line 331:
```suggestion
: A JSON String containing a Wallet attestation along with a proof of possession of the configuration key as defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. This assertion is used to authenticate the Wallet towards the Verifier. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+The follwowing parameters are defined: 
+
+`request`:
+: A JSON String containing the signed request object as defined in [@RFC9101]. It MUST fulfill the requirements as defined in (#vp_token_request).
+
+
+### Presentation Error Response
+
+The error code `401` signals to the wallet that it needs to authenticate to the Verifier. In this case, the error response SHOULD contain a `WWW-Authenticate` header for every attestation method the Verifier supports.
```
peppelinux, line 345:
```suggestion
The error code `401` signals to the Wallet that it needs to authenticate to the Verifier. In this case, the error response SHOULD contain a `WWW-Authenticate` header for every attestation method the Verifier supports.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+This is an example for the wallet attestation method as specified above. The `WWW-Authenticate` contains the nonce value the Wallet MUST use in the calculation of the proof of possession of the respective wallet attestation.
```
peppelinux, line 347:
```suggestion
Below a non-normative example for the Wallet attestation method as specified above. The `WWW-Authenticate` contains the nonce value the Wallet MUST use in the calculation of the proof of possession of the wallet attestation. 
```

peppelinux (changes requested):
A more accurate revision Is needed, there are some details to be further Explorer and consolidates but I like the main line, in particular the way the nonce is provided. At a first glance!

examples/discovery/discover.json
```diff
@@ -0,0 +1,4 @@
+{
+    "presentation_request_uri": "verifier.example.com/presentation_request",
+    "context": "register_kyc"
+}
```
Sakurann, line 4: could these kind of small examples be in the text and not as a separate file..?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -126,7 +126,7 @@ This specification supports any Credential format used in the Issuer-Holder-Veri
 
 Implementations can use any pre-existing OAuth 2.0 Grant Type and Response Type in conjunction with this specification to support different deployment architectures.
 
-OpenID for Verifiable Presentations supports scenarios where the Authorization Request is sent both when the Verifier is interacting with the End-User using the device that is the same or different from the device on which requested Credential(s) are stored.
+OpenID for Verifiable Presentations supports scenarios where the Authorization Request is sent both when the Verifier is interacting with the End-User using the device that is the same or different from the device on which requested Credential(s) are stored. It also allows the Verifier to first discover the Wallet's capabilities before the actual presentation request is created and passed to the wallet.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
OpenID for Verifiable Presentations supports scenarios where the Authorization Request is sent both when the Verifier is interacting with the End-User using the device that is the same or different from the device on which requested Credential(s) are stored. 

This specification also provides the Verifier an option to first discover the Wallet's capabilities before the actual presentation request is created and passed to the Wallet. 
```

this point is significant enough it should have a separate paragraph.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
 
 (3) The Wallet prepares the Verifiable Presentation(s) of the Verifiable Credential(s) that the End-User has consented to. It then sends to the Verifier an Authorization Response where the Verifiable Presentation(s) are contained in the `vp_token` parameter.
 
+## Advanced Flow with Wallet Capabilities Discovery {#adv_flow}
```
Sakurann, line 218:
```suggestion
## Cross Device Flow with Wallet Capabilities Discovery {#adv_flow}
```
Not sure I am a big fan of "advance flow" term - feels too abstract.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
+
+Below is a diagram of a flow where the End-User presents a Credential to a Verifier interacting with the End-User on a different device as the device the Wallet resides on. In this case, the Verifier starts the flow by discovering the Wallet's capabilities through an additional message exchange. This is especially useful, if the Verifier starts the process using a URL serving as alias for a group of wallets (e.g. a custom scheme) but would like to tailor the presentation request to the particular wallet.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
Below is a diagram of a flow where the Verifier starts the flow by discovering the Wallet's capabilities through an additional message exchange. This is especially useful, if the Verifier starts the process using a URL serving as alias for a group of Wallets (e.g. a custom scheme) but would like to tailor the presentation request to the particular Wallet. This example uses cross-device flow where the End-User presents a Credential to a Verifier interacting with the End-User on a different device as the device the Wallet resides on. 
```
I would emphasize the wallet capability selection part as opposed to starting with cross-device description.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
 
+## Advanced Flow with Wallet Capabilities Discovery {#adv_flow}
+
+Below is a diagram of a flow where the End-User presents a Credential to a Verifier interacting with the End-User on a different device as the device the Wallet resides on. In this case, the Verifier starts the flow by discovering the Wallet's capabilities through an additional message exchange. This is especially useful, if the Verifier starts the process using a URL serving as alias for a group of wallets (e.g. a custom scheme) but would like to tailor the presentation request to the particular wallet. 
+
+In this flow, the Verifier first prepares a discovery request and sends it to the Wallet. The Wallet answers this request by sending an HTTP POST request to the Verifier, providing the Verifier with its Issuer URL and further parameters. The Verifier then creates a presentation request tailored to the capabilities and trust domain of the partiular Wallet and passes the request object in the HTTP POST response. The rest of the flow works like described in (#same_device). The discovery request can also be used in cross device scenarios, the Verifier would render the disovery message in a QR code.
```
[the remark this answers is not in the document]
Sakurann: +1 I had the same feedback as Giuseppe. probably something like `providing the Verifier with an app link/universal link specific to that Wallet`. 
though is it only app/universal link that the Wallet can return? does it make sense returning cloud wallet's backend endpoint?

openid-4-verifiable-presentations-1_0.md
peppelinux: Yes, I'd tend to not use comparative terms like advanced or High or anything that might produce a relation or reference to something lower or feature-less

peppelinux:
@tlodderstedt 

Here are the elements that come to mind:

1. An RP or a Wallet Instance acting as a Verifier, should provide wallet discovery endpoint, be this a www https url (RP) or a universal-link/url-scheme (Wallet)
2. wallet discovery endpoint should issue a discovery nonce, called `d_nonce`
3. the wallet discovery endpoint may be not supported by RP that doesn't require this assurance level related to the wallet capabilities

I would like to simplify and at the same time give more guidance to the implementers, then I would remove `presentation_uri`, `v_nonce`, `w_nonce`, `context` giving in addition details about cross-device and same-device flows

[This is the flow](http://www.plantuml.com/plantuml/uml/jLPRRzj64ttdhnZevPFyqRMTlWck4IENDcq2gvRJl1WGbhj3QUFLBhiNsMwR_lREBfSYnTZ8Gz470mOvS-QSCnVnfVFC-h1L1GlUwB2jq1P5PHvWyj6X_P-3rxYJ7DsaQ2XGSjamxM7oYIc57jvfojSS9y0St6yFOX80NAvH-_Gw5AW547J6_msjh2NQ4V9RQxIdo9HZ3z-ziiQvbX3SwgLdFZXueqLZP2xYy4cUhwNZPeVsaT8Euk_5AE4QVm_eF0Ghnx6s6OSAQP4_7LlrQei0nnTWv_3JpSq2Se5R3Oqrdj9Hq3DdWkLOCExbZdaauJOd-Gr2TF8uEGdIWJOUMA3dchWmoa-nKSeSZ9LhgPcNUWqsIx6es2D52Hp0zwMY1VItXn0OMK7ocl1WoRO8sVHQcVk3O6unOZ3mbcaNCOq-beBWLGSnJ8MKIyaXkl7XFJHiJGoaty3Fot840amBylNNQtXbH1jrgpjLgsWwPijBu4gI7IifpckZFN2ZZFsXKW4lIX1TysFEgiyHAsqqn_EJc79HG06z9zk2_fBAxHok6jIf3wEfBucTCfofOJk8EpskRNIDg7EYFiHQSgOXIqe2yS5RgZi_bFOlP0Wya560CgQ1ZyHEKR56siSIQYOLYjXT5pX7v-gWgBnhZ7P45zB4EkhlvMVsaEQs8vVQC6xKzCStDtyjhfOtCyfrs2Np3V8xQ0YNl6cTSUtEodfE9fdACwbHbB1ZAlSxEBIZwIR8VcY8oidPTsVmulGKhduXa3ydDkzXz6KoXuttZODWrPSY3aY1ok5uYCUJIFhssppz_lJ5BCcC7F-xSGlTI3JMxAJ0biPR4yW5PTf9AgA08a-Fm4prgwBH-hvknpFtAdDbpg6Dc7FuzExoxyNLOfePpJfAmyL6QaaW-zruw86pXbLIII_HaK7Joylt9Nn0pmJpR4PZ71f1YMbX73eNsrubLTwaX_TO4Nbp9_FToJPyfIfgJNtE_61umIYpCyOssHUQmbXZsBBaG7VJhwhFiS7_7n2mJOUGBzdKoRL6CJiSssxuJKA8glSJK48oiTolJwmRByCGO2Mmbbh6Vykq0pXBfRFqewEswgZEqH5yiywtglINPxXKSTNI0sFb7sshhXU5DlJ6olN6fuIqYjpeY9yI72haoGzJft5JnX7fsFT1dMpaBlsyiqhXePLJUYyY3wOwThQY9J3ExwN1TDUilBb3FHBeGhMLpWq5zhzfLM6h3YuT5EP7jQbloKIef1PIhsU3Z7xWSmeAwAd2y1TDf3j2kTq1EQykqg5CbvPveeODj1zJS7PsVb9TG7VhuFbxbVTZy5rWytV1boS_8PwFxs_Ua6jgJtz1wSn1dVPosZ5w7d7MNYA1GxjVqhFuwVWF) that I have in mind:

immagine

peppelinux:
WDYT about changinc the name from "Advanced flow" to "Wallet Capabilities Discovery flow"?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
+|     User     |   |   Verifier   |                                    |    Wallet    |
++--------------+   +--------------+                                    +--------------+  
+        |                 |                                                   |
+        |    Interacts    |                                                   |
+        |---------------->|                                                   |
+        |                 |  (1) Discovery Request                            |
+        |                 |  (Presentation Request URI)                       |
+        |                 |-------------------------------------------------->|
+        |                 |                                                   |
+        |                 |  (2) Request the Request Object                   |
```
Sakurann, line 241: I think it should be clearer that this (2) is also a "Discovery Response", correct?

Sakurann:
```suggestion
        |                 |  (2) Discovery Response                                    |
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
+        |                 |-------------------------------------------------->|
+        |                 |                                                   |
+        |   User Authentication / Consent                                     |
+        |                 |                                                   |
+        |                 |  (3)   Authorization Response                     |
+        |                 |  (VP Token with Verifiable Presentation(s))       |
+        |                 |<--------------------------------------------------|
+~~~
+!---
+Figure: Cross Device Flow
```
Sakurann, line 255:
```suggestion
Figure: Wallet Discovery Flow
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
+
+(1) The Verifier sends to the Wallet a Discovery Request that contains a Presentation URI from where to obtain the Request Object containing the actual Authorization Request object.
```
Sakurann, line 257:
```suggestion
(1) The Verifier sends to the Wallet a Discovery Request that contains a Presentation URI to where Wallet can send information about itself and obtain the Request Object containing the actual Authorization Request. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
+        |                 |  (3)   Authorization Response                     |
+        |                 |  (VP Token with Verifiable Presentation(s))       |
+        |                 |<--------------------------------------------------|
+~~~
+!---
+Figure: Cross Device Flow
+
+(1) The Verifier sends to the Wallet a Discovery Request that contains a Presentation URI from where to obtain the Request Object containing the actual Authorization Request object. 
+
+(2) The Wallet sends an HTTPS POST request to the Presentation URI to retrieve the Request Object.
```
```suggestion
(2) The Wallet sends an HTTPS POST request to the Presentation URI that can contain information about wallet's endpoint, capabilities and attestation.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +215,53 @@ Figure: Cross Device Flow
...
+
+(2.5) The HTTPS POST response returns the Request Object containing Authorization Request parameters. It especially contains a Presentation Definition as defined in [@!DIF.PresentationExchange] that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Request Object and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers her consent to present the requested Credentials.
```
Sakurann, line 261:
```suggestion
(2.5) The HTTPS POST response returns the Request Object containing Authorization Request parameters based on the information specific to the Wallet that it received in step (2). It especially contains a Presentation Definition as defined in [@!DIF.PresentationExchange] that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Request Object and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers her consent to present the requested Credentials. 
```
do we need to repeat `it especially contains...` part which is duplicative across all examples? guess we do in case reader start reading one example without reading others..

openid-4-verifiable-presentations-1_0.md
Sakurann: I think the way this section is written right now does not exactly convey why this is an "advanced" flow. maybe i am misunderstanding something but I would expect more focus on the new ability for the wallet to communicate its capabilities/endpoint/attestation

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
 
 Presentation of Verifiable Credentials using OpenID for Verifiable Presentations can be combined with the user authentication using [@SIOPv2], and the issuance of OAuth 2.0 Access Tokens.
 
+# Discovery Request
+
+The Discovery Request allows Verifers to discover the Wallet's capabilities before they create the actual presentation request. It also allows the Wallet to authenticate towards the Verifier and to provide the Verifier with additional data used in the creation of the request signature and to encrypt the request at the application layer (if needed).
```
Sakurann, line 280:
```suggestion
The Discovery Request allows Verifers to discover the Wallet's capabilities before they create the actual presentation request. It also allows the Wallet to authenticate towards the Verifier and to provide the Verifier with additional data used in the creation of the request signature and to encrypt the request at the application layer (if needed).
```
in my mind, wallet's endpoint, wallet's capabilities, and wallet attestation are three different things, and I am not sure "wallet capabilities" can be used to express them all

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+The Discovery Request message has the following parameters:
```
Sakurann, line 282:
```suggestion
This specification defines the following new parameters for the Discovery Request message:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+`presentation_request_uri`:
+: A string containing an HTTPS URL pointing to a resource under the control of the Verifier where the Wallet is supposed to obtain the presentation request object.
```
Sakurann, line 285: I don't think `presentation_request_uri` parameter name could be more explicit and differentiate it more clearly from `request_uri` in rfc9101..? something like `discovery_and_request_uri` (yes, I know i am not brilliant at names)?

I would also borrow some language from https://www.rfc-editor.org/rfc/rfc9101#section-5-2.4 like 
```suggestion
`presentation_request_uri`:
: The absolute URI, as defined by [RFC 3986](https://www.rfc-editor.org/rfc/rfc9101#RFC3986) [[RFC3986](https://www.rfc-editor.org/rfc/rfc9101#RFC3986)], that receives data from the Wallet such as endpoint, capabilities and attestation and returns the [Request Object](https://www.rfc-editor.org/rfc/rfc9101#name-request-object) containing the authorization request parameters defined in [Section 5](https://openid.bitbucket.io/connect/openid-4-verifiable-presentations-1_0.html#section-5).
```
not sure we need ` If this parameter is present in the authorization request, request MUST NOT be present.`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+The Discovery Request is represented as a JSON object, where all parameters are top level JSON claims. 
+
+The following is an example request: 
+
+```JSON
+{
+    "presentation_request_uri": "verifier.example.com/presentation_request",
+    "context": "register_kyc"
+}
+```
```
Sakurann, line 299: just to note that parameter names might change based on the comments above.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+The Discovery Request message has the following parameters:
+
+`presentation_request_uri`:
+: A string containing an HTTPS URL pointing to a resource under the control of the Verifier where the Wallet is supposed to obtain the presentation request object. 
+
+`context`:
+: A string identifying the context of the discovery request from the perspective of the Verifier. The value is opqaue to the Wallet, it MUST pass this value to the Presentation Endpoint Request (see below). 
+
+The Discovery Request is represented as a JSON object, where all parameters are top level JSON claims.
```
Sakurann, line 290:
```suggestion
The Discovery Request MUST be represented as a JSON object with all parameters as top-level JSON claims. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+    "presentation_request_uri": "verifier.example.com/presentation_request",
+    "context": "register_kyc"
+}
+```
+
+The Discovery Request is either send the Discovery Endpoint of the Wallet or it is rendered as a QR Code and scanned with the Wallet or a camera app. 
+
+Note: the Discovery Request intentionally does not use OAuth parameters, such as a `client_id`. The idea is to allow the Verifier to select the Client Identifier it wants to use for interacting with the Wallet after it has determined the trust mechanisms and domains the wallet supports.   
+
+## Discovery Endpoint
```
Sakurann, line 305: I think this section needs to be in the Verifier Metadata section https://openid.bitbucket.io/connect/openid-4-verifiable-presentations-1_0.html#name-wallet-metadata-authorizati and not here

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+The following is an example request: 
+
+```JSON
+{
+    "presentation_request_uri": "verifier.example.com/presentation_request",
+    "context": "register_kyc"
+}
+```
+
+The Discovery Request is either send the Discovery Endpoint of the Wallet or it is rendered as a QR Code and scanned with the Wallet or a camera app.
```
Sakurann, line 301:
```suggestion
The Discovery Request is either sent to the Discovery Endpoint of the Wallet or is rendered as a QR Code.
```
i think reference to the section where discovery endpoint (not sure about the name...) is defined would be nice

openid-4-verifiable-presentations-1_0.md
Sakurann: i can see both sides, but if the value of this mechanism is for the verifier to be able to tailor the request based on the POST from the wallet, this sentence makes sense. to avoid sending data to random verifiers, wallet can have the list of trusted `presentation_request_uri`? or append .well-known to the presentation_request_uri? (thinking out loud here)

openid-4-verifiable-presentations-1_0.md
Sakurann, line 309: I don't understand this.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+This is an HTTPS endpoint offered by the Wallet. The Discovery Request MUST be sent as a HTTPS GET request or an HTTP redirect to the Discovery Endpoint URL defined in (#wallet-metadata)
+
+The HTTPS URL contains a single URI query parameter `discovery` where the value is the JSON object as defined above. 
+
+## Presentation Endpoint
+
+The Presentation Endpoint is an HTTPS endpoint exposed by the Verifier. 
+
+### Presentation Request
```
Sakurann, line 315: I am not comfortable calling this `presentation request`. requests in OID4VP are from the verifier. it is confusing to introduce a `request` coming from the wallet. why can't this be a `discovery response` to `discovey request` from the verifier? or `wallet capability offer` to learn from VCI

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+Presentation Requests MUST be HTTPS POST requests with the "application/json" media type.
+
+The follwowing parameters are defined: 
+
+`context`:
+: A JSON String containing the value of the corresponding Discovery Request's `context` parameter.
```
Sakurann, line 322:
```suggestion
`verifier_state`:
: A JSON String containing the value of the corresponding Discovery Request's `verifier_state` parameter.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: +1 same as a thread here https://github.com/openid/OpenID4VP/pull/52/files?diff=unified&w=0#r1339310421

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+## Presentation Endpoint
+
+The Presentation Endpoint is an HTTPS endpoint exposed by the Verifier. 
+
+### Presentation Request
+
+Presentation Requests MUST be HTTPS POST requests with the "application/json" media type.
+
+The follwowing parameters are defined:
```
Sakurann, line 319:
```suggestion
This specification defines the following new parameters:
```

Sakurann, line 319: please indicate which parameters is REQUIRED vs OPTIONAL - hard to review without that

openid-4-verifiable-presentations-1_0.md
Sakurann, line 325: also this PR needs to define how wallet capability/endpoint parameters can be obtained by appending `/.well-known/oauth-authorization-server` to this issuer URL. also I think it should define how wallet can pass those parameters by value in this POST message - i can totally see wallets doing that.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+: A JSON String containing a wallet attestation along with a proof of posession of the confiration key as defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. This assertion is used to authenticate the Wallet towards the Verifier. 
+
+### Presentation Response
+
+Presentation Response MUST be HTTPS POST response with the "application/json" media type.
+
+The follwowing parameters are defined: 
+
+`request`:
+: A JSON String containing the signed request object as defined in [@RFC9101]. It MUST fulfill the requirements as defined in (#vp_token_request).
```
Sakurann, line 340: I think this is confusing. why is there suddenly a need to define a top level claim? usually the entire object in the response is a request object without any top level claims. the parameter name also conflicts with the `request` parameter defined in rfc 9101 to pass the request by value

Sakurann (changes requested):
I think there are a lot of details to be brushed up.

openid-4-verifiable-presentations-1_0.md
bc-pi: "different device as the device the Wallet resides on" -> "different device than the device that the Wallet resides on" maybe reads more clearly (or maybe it's just me).

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
bc-pi: also unsure what "Issuer URL" means here

openid-4-verifiable-presentations-1_0.md
bc-pi: JSON doesn't have claims.

openid-4-verifiable-presentations-1_0.md
bc-pi: ietf-oauth-attestation-based-client-auth is a method of client authentication to the AS. I don't think it works to just reference here as a way to authenticate the Wallet (the AS in this context) towards the Verifier (the client in this context).

openid-4-verifiable-presentations-1_0.md
bc-pi: does {#wallet-metadata} exist anywhere? Seems to be causing problems in the "build" https://github.com/openid/OpenID4VP/actions/runs/6324297591/job/17173526145?pr=52 

```2023/09/27 09:52:29 Empty 'stream' in [seriesInfo], resulting XML may fail to parse.
./openid-4-verifiable-presentations-1_0-20.xml(0): Error: IDREF attribute target references an unknown ID "wallet-metadata", at None
/data/openid-4-verifiable-presentations-1_0-20.xml(3): Error: Invalid document before running preptool.
Unable to complete processing ./openid-4-verifiable-presentations-1_0-20.xml
Error: Process completed with exit code 1.```

**Comment by bc-pi** [comment]
My review was requested on this pull request and so I tried to do a review but what I was able to do was not at all comprehensive or complete. Not sure what I'm trying to say here, provide a disclaimer of sorts, but seemed like something that needed saying.

File: openid-4-verifiable-presentations-1_0.md (hunk 16 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:259]: wallet's endpoint -> Wallet's endpoint

File: openid-4-verifiable-presentations-1_0.md (hunk 18 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:280]: I assume that the wallet instance attestation contains the wallet's capabilities, materially defined with:

- cnf.jwk (any other cnf.*)
- NiST's AAL (the equivalence of the italian `attested_security_contexts`)
- Authorization Server metadata
- other wallet instance metadata, as client_metadata when a wallet instance acts like a Verifier.

File: openid-4-verifiable-presentations-1_0.md (hunk 22 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:290]: according to rfc8259

```
An object structure is represented as a pair of curly brackets
   surrounding zero or more name/value pairs (or members).
```

**Comment by tlodderstedt** [comment]
> A more accurate revision Is needed, there are some details to be further explored and consolidated but I like the main line, in particular the way the nonce is provided. At a first glance!

The nonce for the wallet attestation pop is provided in the presentation error response.

File: openid-4-verifiable-presentations-1_0.md (hunk 11 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:222]: I made a mistake, as the flow obviously is same device. 
The sentence (and with that the order of things) was copied from the other flows. It just spells out the fact that this is a same device flow. 
Changed the text to refer to the same device flow instead of duplicating text.

File: openid-4-verifiable-presentations-1_0.md (hunk 12 above)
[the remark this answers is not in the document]
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:226]: The Issuer URL ist the URL identifying the Wallet as an AS. This URL is used to fetch the wallet's metadata.

File: openid-4-verifiable-presentations-1_0.md (hunk 11 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md]: please check

File: openid-4-verifiable-presentations-1_0.md (hunk 4 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:329]: by relying on RFC8414 for wallet metadata (Section 8.1) the wallet as AS needs to have an Issuer URL

File: openid-4-verifiable-presentations-1_0.md (hunk 5 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:335]: happy to extend the text if we have basically agreed to use the mechanism

File: openid-4-verifiable-presentations-1_0.md (hunk 10 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:220]: Advanced is the title of the PR, now it seems it will be a discovery (pre-step). Changed title and first sentence.

File: openid-4-verifiable-presentations-1_0.md (hunk 13 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:245]: there is discovery, but the result is a request object

File: openid-4-verifiable-presentations-1_0.md (hunk 14 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:259]: changed to "Same Device Flow with Wallet Capabilities Discovery"

File: openid-4-verifiable-presentations-1_0.md (hunk 16 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md]: I adopted the text to convey what I perceive was the objective of the proposal, I used different text since there is no capabilities nor endpoint directly be conveyed in the POST request as it is specified now.

File: examples/discovery/discover.json (hunk 8 above)
Comment (tlodderstedt) [reply examples/discovery/discover.json:4]: it is inline, I just used to file to format the example - can remove it once we merge

File: openid-4-verifiable-presentations-1_0.md (hunk 18 above)
Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:284]: those three things are differentiated in the current text. wallet capabilities is in the first sentence, authentication in the second sentence

Comment (tlodderstedt) [reply openid-4-verifiable-presentations-1_0.md:284]: @peppelinux I assume the capabilities are conveyed using wallet metadata. I agree, there can be other ways. Let's discuss how we can support them.

**Comment by peppelinux** [comment]
@tlodderstedt 's

> The nonce for the wallet attestation pop is provided in the presentation error response.

Regarding the way the nonce would be provided, I'm looking for any kind of harmonizations with the issuance flow and in particular for the solution you have provided, there are sensibilities within some groups of implementers to have a specialized nonce endpoint with a fallback strategy with the error response, as remarked here

https://github.com/openid/OpenID4VCI/issues/71#issuecomment-1744352204

File: openid-4-verifiable-presentations-1_0.md (hunk 13 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:245]: this reminds me the italian flow where we have mixed the provisioning of the wia with DPoP to the request_uri endpoint.

We are looking for a better specialization of the endpoints, then the proposed, alternative, flow was provided in this comment: https://github.com/openid/OpenID4VP/pull/52#issuecomment-1741179550

File: openid-4-verifiable-presentations-1_0.md (hunk 14 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:259]: I would like to cover both cross device and same device within this PR

an alternative flow was provided aiming to satisfy this: https://github.com/openid/OpenID4VP/pull/52#issuecomment-1741179550

File: openid-4-verifiable-presentations-1_0.md (hunk 20 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:289]: I'm not in favor of discovery AND request, since I assume that there might be cases where a wallet solution doesn't support the discovery feature then the request_uri endpoint should be untouched

I assume, according to my implementations, that the RP that can't assert the wallet capabilities should attests to the wallet instance the lowest level of security, if supported for the requirements of the LoA

File: examples/discovery/discover.json (hunk 8 above)
Comment (Sakurann) [inline examples/discovery/discover.json:4]:
```suggestion
```

Sakurann: thank you

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +217,55 @@ Figure: Cross Device Flow
 
 (3) The Wallet prepares the Verifiable Presentation(s) of the Verifiable Credential(s) that the End-User has consented to. It then sends to the Verifier an Authorization Response where the Verifiable Presentation(s) are contained in the `vp_token` parameter.
 
+## Same Device Flow with Wallet Capabilities Discovery {#adv_flow}
+
+Below is a diagram shows a flow a describe in ({#same_device}) with an additional pre-step used to discover the Wallet's capabilities and tailor the actual request to it.
```
Sakurann, line 222:
```suggestion
Below is a diagram that shows a flow described in (#same_device), but with an additional pre-step used that allows the Verifier to discover the Wallet's capabilities and tailor the actual request.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +217,55 @@ Figure: Cross Device Flow
 
 (3) The Wallet prepares the Verifiable Presentation(s) of the Verifiable Credential(s) that the End-User has consented to. It then sends to the Verifier an Authorization Response where the Verifiable Presentation(s) are contained in the `vp_token` parameter.
 
+## Same Device Flow with Wallet Capabilities Discovery {#adv_flow}
```
Sakurann, line 220: I think {#xxx_xxx} shows in the URL to this section, so I think we should start cleaning it up something like {#capability_discovery_flow}. (this can be a separate issue/PR)

openid-4-verifiable-presentations-1_0.md
tlodderstedt: The idea in the PR is to have a new endpoint (the presentation endpoint) that bundles discovery and request creation. We could also split this into discovery and request_uri. However I would like to understand the consequences. What directly comes to mind is that the state must be kept at the verifier between the two requests whereas the current design allows the verifier to create the request and directly release it to the wallet.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +275,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+The Discovery Request allows Verifers to discover the Wallet's capabilities before they create the actual presentation request. It also allows the Wallet to authenticate towards the Verifier and to provide the Verifier with additional data used in the creation of the request signature and to encrypt the request at the application layer (if needed).
+
+The Discovery Request message has the following parameters:
+
+`presentation_request_uri`:
+: A string containing an HTTPS URL pointing to a resource under the control of the Verifier where the Wallet is supposed to obtain the presentation request object. 
+
+`context`:
+: A string identifying the context of the discovery request from the perspective of the Verifier. The value is opqaue to the Wallet, it MUST pass this value to the Presentation Endpoint Request (see below).
```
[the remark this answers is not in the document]
tlodderstedt: I intentionally did not designate it as state as it is intended to be used to label the context ("entrance to east wing") and not carry a certain state of a certain transaction. This allows to use the same discovery request for multiple transactions (print it out and put it onto the wall).

openid-4-verifiable-presentations-1_0.md
tlodderstedt: @bc-pi what do you suggest?

openid-4-verifiable-presentations-1_0.md
tlodderstedt: This endpoint is not the typical static "get some metadata" endpoint. It is part of the application logic of the wallet. I therefore added it here. I think this is the best solution for readability and understandability.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: added text re 'oauth-authorization-server' we-known location

we need to discuss the security implications of passing metadata as parameters. I guess that is @peppelinux has some ideas about it

obtaining the data through /.well-known/oauth-authorization-server at least binds the data to the issuer URL

openid-4-verifiable-presentations-1_0.md
tlodderstedt: I changed the label of this section

David-Chadwick:
@paulbastian I would further add that on the open internet sending any attestations about the wallet instance is frowned upon by the IAB, see
https://www.iab.org/documents/correspondence-reports-documents/2023-2/iab-statement-on-the-risks-of-attestation-of-software-and-hardware-on-the-open-internet/
Nevertheless, if the attestation does not identify the wallet software, but simply attests to a standardised set of capabilities that any RP may require from any wallet, so that the wallet software remains anonymous, then this is OK from my perspective.

danielfett (changes requested):
I think this goes into the right direction. 

I don't immediately see security issues with this that we wouldn't have with similar solutions (SSRF comes to mind).

But there need to be privacy considerations going along with this. Right now, this opens up the possibility for fingerprinting, as malicious websites can just send a discovery request to the wallet and receive a presentation request that can potentially leak a lot of information about the user's device.

One option would be that there should be a user interaction before sending the presentation request. That's probably not ideal, so we might need to think about other options as well.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +217,55 @@ Figure: Cross Device Flow
 
 (3) The Wallet prepares the Verifiable Presentation(s) of the Verifiable Credential(s) that the End-User has consented to. It then sends to the Verifier an Authorization Response where the Verifiable Presentation(s) are contained in the `vp_token` parameter.
 
+## Same Device Flow with Wallet Capabilities Discovery {#same_device_with_discovery}
+
+Below is a diagram that shows the flow as described in (#same_device), but with an additional pre-step used that allows the Verifier to discover the Wallet's capabilities and tailor the actual request.
+
+In this case, the Verifier starts the flow by discovering the Wallet's capabilities through an additional message exchange. This is especially useful, if the Verifier starts the process using a URL serving as alias for a group of wallets (e.g. a custom scheme) but would like to tailor the presentation request to the particular wallet.
```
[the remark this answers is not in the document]
David-Chadwick: We should add the following Note somewhere
Note. In order to prevent a Verifier from fingerprinting the wallet, it is strongly recommended that one or more standardised sets of capabilities are defined that may be commonly supported by wallets. This will allow the Verifier to request conformance to one or more sets, and the wallet to return an assertion about one or more sets, thereby protecting the privacy of the wallet implementation.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -126,7 +126,9 @@ This specification supports any Credential format used in the Issuer-Holder-Veri
 
 Implementations can use any pre-existing OAuth 2.0 Grant Type and Response Type in conjunction with this specification to support different deployment architectures.
 
-OpenID for Verifiable Presentations supports scenarios where the Authorization Request is sent both when the Verifier is interacting with the End-User using the device that is the same or different from the device on which requested Credential(s) are stored.
+OpenID for Verifiable Presentations supports scenarios where the Authorization Request is sent both when the Verifier is interacting with the End-User using the device that is the same or different from the device on which requested Credential(s) are stored.
```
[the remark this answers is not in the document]
peppelinux: yes, let's try to make it simpler.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
peppelinux: I fully agree with Kristina, `context` is something too much related to other specs and schemas.
I propose `discovery_state` with the will to keep the words always on the same direction, the good semantic help us to reduce the brain stress

openid-4-verifiable-presentations-1_0.md
peppelinux: I'm for `discovery_response_uri` where the HTTP response could be 302 Found with the discovery_state that binds the discovery to the request issuance

then we have divided the endpoints by using the request_uri endpoint with the sole addition of the discovery_state that allows the RP to do its lookups. This value could be also encrypted (depending by what would cost more in an infrastructure: CPU or Storage!)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+This specification defines the following new parameters for the Discovery Request message:
+
+`presentation_request_uri`:
+: A string containing an HTTPS URL pointing to a resource under the control of the Verifier where the Wallet is supposed to obtain the presentation request object. 
+
+`context`:
+: A string identifying the context of the discovery request from the perspective of the Verifier. The value is opqaue to the Wallet, it MUST pass this value to the Presentation Endpoint Request (see below). 
+
+The Discovery Request MUST be represented as a JSON object with all parameters as top-level JSON claims.
```
peppelinux, line 294:
```suggestion
The Discovery Request MUST be a JSON object with all members in the top-level. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+`context`:
+: A string identifying the context of the discovery request from the perspective of the Verifier. The value is opqaue to the Wallet, it MUST pass this value to the Presentation Endpoint Request (see below). 
+
+The Discovery Request MUST be represented as a JSON object with all parameters as top-level JSON claims. 
+
+The following is an example request: 
+
+```JSON
+{
+    "presentation_request_uri": "verifier.example.com/presentation_request",
```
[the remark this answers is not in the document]
peppelinux: according to my rpevious comments, it would be

`"discovery_response_uri": "https://verifier.example.com/discovery-response"`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+}
+```
+
+The Discovery Request is either sent to the Discovery Endpoint of the Wallet or is rendered as a QR Code.
+
+Note: the Discovery Request intentionally does not use OAuth parameters, such as a `client_id`. The idea is to allow the Verifier to select the Client Identifier it wants to use for interacting with the Wallet after it has determined the trust mechanisms and domains the Wallet supports.   
+
+## Discovery Endpoint
+
+This is an HTTPS endpoint offered by the Wallet. The Discovery Request MUST be sent as an HTTPS GET request or an HTTP redirect to the Discovery Endpoint URL defined in (#wallet-metadata).
```
peppelinux, line 311: `This is an HTTPS endpoint offered by the Wallet.`

please don't, we should assume that the wallet should not require any online HTTPs endpoint

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+## Presentation Endpoint
+
+The Presentation Endpoint is an HTTPS endpoint exposed by the Verifier. 
+
+### Presentation Request
+
+Presentation Requests MUST be HTTPS POST requests with the "application/json" media type.
+
+The follwowing parameters are defined:
```
peppelinux, line 323:
```suggestion
The following parameters are defined: 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+Presentation Requests MUST be HTTPS POST requests with the "application/json" media type.
+
+The follwowing parameters are defined: 
+
+`context`:
+: A JSON String containing the value of the corresponding Discovery Request's `context` parameter.
+
+`issuer`:
+: A JSON containing an HTTPS URL designating the Issuer URL of the Wallet. The Verifier can obtain the Wallet's metadata by adding the well-know location `oauth-authorization-server` as specifid in [@!RFC8414].
```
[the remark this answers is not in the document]
peppelinux: this term `Issuer URL of the Wallet` or `Issuer URL` is misleading, would it mean Wallet Provider?

I'm strongly not in favour of requiring the entities to be online and that the wallet might exposes its metadata thought the wallet providers endpoints

the wallet instance would be the best place, attestable and not-repudiable, where the wallet metadata can be provided.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+
+`w_nonce`:
+: A JSON String containg a cryptographic nonce the Verifier MUST use when creating the signed presentation request object.
```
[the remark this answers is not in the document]
peppelinux: where the reader is able to distinguish between nonce and cryptographic nonce?
I would suggest a ref where the definition of cryptographic nonce is provided and the way to issue it

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,84 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+: A JSON String containing the value of the corresponding Discovery Request's `context` parameter.
+
+`issuer`:
+: A JSON containing an HTTPS URL designating the Issuer URL of the Wallet. The Verifier can obtain the Wallet's metadata by adding the well-know location `oauth-authorization-server` as specifid in [@!RFC8414].
+
+`w_nonce`:
+: A JSON String containg a cryptographic nonce the Verifier MUST use when creating the signed presentation request object. 
+
+`client_assertion`
+: A JSON String containing a Wallet attestation along with a proof of possession of the configuration key as defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. This assertion is used to authenticate the Wallet towards the Verifier.
```
peppelinux, line 335:
```suggestion
: A JSON String containing a Wallet attestation along with a proof of possession of the public key as defined in [@!I-D.ietf-oauth-attestation-based-client-auth]. This assertion is used to authenticate the Wallet towards the Verifier. 
```

peppelinux (changes requested):

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
peppelinux: I agree, since there may be cases where the wallet instances doesn't support discovery, then the RP has to rely on a standard set. 

@David-Chadwick do you think htat this set should be recommended in the current spec or that it would be out of scope and then it should be defined within each impl profile? (eg: EUDI Wallet should have its standard set ... and so on)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +217,55 @@ Figure: Cross Device Flow
...
+        |   User Authentication / Consent                                     |
+        |                 |                                                   |
+        |                 |  (3)   Authorization Response                     |
+        |                 |  (VP Token with Verifiable Presentation(s))       |
+        |                 |<--------------------------------------------------|
+~~~
+!---
+Figure: Cross Device Flow
+
+(1) The Verifier sends to the Wallet a Discovery Request that contains a Presentation URI to where Wallet can send information about itself and obtain the Request Object containing the actual Authorization Request.
```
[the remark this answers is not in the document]
peppelinux: @David-Chadwick ^

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: Since David is not concerned about interoperability here, but privacy, this could go into the privacy considerations I mentioned in my comment on this PR. However, I think that just using some set of default capabilities is not sufficient: There are many, many ways in which information about the wallet can be leaked. Browser vendors are working quite hard to make browser fingerprinting harder, and we need to take a page or two out of their book if we allow this request to happen without user consent, to implement similar protections. For example: Just the presence of a particular wallet might leak the nationality of a user, and that particular wallet might send parameters in a different order than all other wallets. Or: The wallet might support a different version of HTTP than all other wallets.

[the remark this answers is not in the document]
David-Chadwick: I agree that privacy protection is extremely difficult (and maybe its impossible for it to be perfect) but we should at least add guidelines or recommendations into how the capabilities detection feature should be used/implemented.
I am happy for my proposed text to go into the privacy considerations section, supplemented with additional cautionary text from Daniel.
@peppelinux Certainly profiles of the spec will define their own set of required capabilities for all implementors of the profile, but if the WG determine that there is a useful common subset that is widely applicable, this could go into an informative annex. (An analogy is the X.500 spec that defined a common set of useful attributes that most people implemented, whilst acknowledging that other groups of users would define their own additional attributes.)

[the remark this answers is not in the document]
peppelinux: @danielfett by tracking Wallet Instance IPv4 with GeoIP it would be enough to track nationality and also mobile broadband provider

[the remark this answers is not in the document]
peppelinux: @danielfett even if a default capabilities is not sufficient (I agree) specs like OIDC requires the sigalg RS256 as a strong recommendation to the entities, SAML2 made something similar in the real-world with sha-1 and then SHA256. I mentions these just to continuing debating on the requirement of giving more guidance to implementers

peppelinux:
@tlodderstedt [WDYT](https://www.planttext.com/?text=jLLDJzmm4BtxLupsqjreeQtK4uSAfBOgXwfRe71PQUNOiukBrqxzaT3--euTPqduaBYK2oAPUM_UcpVXr0VkGjnhncEm9kvhTAoszp3xwT6zyV09MoNGprX3TKgeXfi0inkkDGOuDzHk1Cw0U-Yc9QaVpdPeGduR6He9XDoZNwDJMuNk4UuNPqsWmjpYkitqxSgX9pmUb3NmsSZ6geBkccdb1Vh66e-Fgkh34C_AMJdRAab3PJaDOn6EZZwcEU04k20RF0GB71fd0ug04cWY6vr0naLGBG-OfquTgJEUG5L9vOLjoGo7loFwK5Lhy-C2X9K8ri45IkK8R6tc1zXDATr4fvOm0FpPaDi15qpYXAsVidk9wzMmhElbDWdfwzNLYcRBzR3wVda5dGgt3qr9URQugc1jo9LdMOXTf8tUXoKexyUanb9mbjJRzURp4D3tkrqI_zEdqDXctlPzYwfQC4E4aDN6PPAVHqx5qVVkVdZt7hpQ6Quz6AHjfUN5S4iPKe8aCyPrWB9ymHjUAws2GWzRP_TmStx6N0F54bi7hemVohzXu9A7d8KK6ldGmr1Jybw6dTuEO8aeosJz9dTKK1pO5wBHTXw_UHtXfTKn-LWqTQ_NDBqEEEvGwwCxOpl3kkOmLMaQfgFBfniNmv_ZiT8Lif6rmc7ACgVRJEiYfD6TbpJHhByeNZ0NMj6xZPBFXAdy6gBPvtnjndWRYLjbL1-p4lm7twig3P3lUinA6VyVl1tqhEDC9XVM9Z-oJlr56BwKJ-8UqmTsr9hCIXQIcp8zuXeyQfgUNhobAzmgjyzL1zxYUviPUArnEZ9fFa0jXlZLj6XBnWnQIUQyRJR1tY49coZsiTxJqK-Shg60j5mhcTWdZ0IcZD37lQDGAoELsOtxIFNGWXEjkQRNvSedbVQW6VsSKjxeNzm_)?

immagine

here the changes

- `context` renamed to `discovery_state` (since it is opaque to the wallet instance) 
- `presentation_uri` renamed to `protected_request_uri` (since it is protected and ruterns 401 if WIA is not provided)
- wallet instance metadata can be obtained in several ways -> optional, depends by trust and interop model

peppelinux:
Following the last discussion, it seems resonable providing the discover request as signed, to le t the Wallet Instace to establish the trust to the RP before providing any information about itself

@tlodderstedt @paulbastian ^

paulbastian:
I agree, but then why not use the signed authorisation request that we have today and send the wallet attestation afterwards in the authorisation response?

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
tlodderstedt: suggestions welcome

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
tlodderstedt: Please suggest text that I can add to the privacy considerations section.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
tlodderstedt: This level of detail is used in same device section, too. I suggest to file an issue to change both.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -215,6 +217,55 @@ Figure: Cross Device Flow
...
+
+In this case, the Verifier starts the flow by discovering the Wallet's capabilities through an additional message exchange. This is especially useful, if the Verifier starts the process using a URL serving as alias for a group of wallets (e.g. a custom scheme) but would like to tailor the presentation request to the particular wallet. 
+
+In this flow, the Verifier first prepares a discovery request to be provided to the Wallet. The Wallet answers this request by sending an HTTP POST request to the Verifier, providing the Verifier with its Issuer URL (identifying the wallet provider and used to fetch wallet metadata) and further parameters. The Verifier then creates a presentation request tailored to the capabilities and trust domain of the particular Wallet and passes the request object in the HTTP POST response. The rest of the flow works like described in (#same_device). The discovery request can also be used in cross device scenarios where the Verifier would render the discovery message in a QR code.
```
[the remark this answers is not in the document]
tlodderstedt: the verifier receives a request via HTTPS, I don't see how this works for offline scenarios

openid-4-verifiable-presentations-1_0.md
tlodderstedt: you are absolutely right, I rewrote the section. please have a look

tlodderstedt:
> context renamed to discovery_state (since it is opaque to the wallet instance) or discovery_session, or discovery_context 

I would like to avoid parameter names causing the impression this parameter should be transaction specific. My idea is to allow for static values so this request could be printed out and put on a wall. The context shall then identify the place where the verifier wants to use the VP. I changed the name to `interaction`. 

> presentation_uri renamed to protected_request_uri (since it is protected and ruterns 401 if WIA is not provided)

protected is pretty generic and holds true for a lot of HTTPS resources

> wallet instance metadata can be obtained in several ways -> optional, depends by trust and interop model

I added text saying metadata MAY be obtained from the AS metadata, but they MAY also be obtained by other means, e.g. as part of the wallet attestation

tlodderstedt:
@peppelinux @paulbastian @David-Chadwick @danielfett 

It seems the most significant issue right now is the question whether the wallet should send data to the verifier before the verifier is authenticated. 

I see the following options:

Option 1: sign the discovery request: that is feasible from a technical standpoint. However, what scheme, what trust framework, and which key should the verifier use? The design assume the verifier will have the require data provided in the presentation request. This allows multi homing verifiers, those connected to a US, a EU, and a JP trust framework. 

Option 2: leave the protocol as is and add privacy considerations recommending a user gesture before the presentation request is sent to the verifier (e.g. the unlock of the wallet). That would at least prevent silent finger printing. 

Option 3: send minimal data to the presentation endpoint, e.g. static profile data. Wallet authentication might happen at the direct post. 

Please let me know what you think.

peppelinux:
@tlodderstedt 

>  I changed the name to interaction.

at least `discovery_interaction`! :-)

> protected is pretty generic and holds true for a lot of HTTPS resources

`presentation_uri` is generic as well since it represents a single endpoint of the entire presentation flow.
I would say that `protected_request_uri` refers to the presentation request, would it be then `presentation_protected_request_uri`?

> I added text saying metadata MAY be obtained from the AS metadata, but they MAY also be obtained by other means, e.g. as part of the wallet attestation

That's good. Because we assume that a specific wallet instance installed on a specific hardware may support a subset of algs supported by wallet solution. Therefore, the Wallet Solution may issue a specific Wallet Instance Attestation containing the tailored capabilities regarding a specific wallet instance.

> Option 1: sign the discovery request

I'm in favor of this. The issue of selecting the trust framework is common to all the signed attestations defines in the OpenID4VC suite. Then I would not see this issue for this flow in particular and I would suggest to keep this out of the scope of this PR.

> Option 2

Too weak from my perspective, since users must be "protected" by their wallet instance.

> Option 3

I would say that this would be too late for the RP, because it should not sign nothing or evaluate nothing before having attested the reliability of the wallet instance, considering also that the wallet capabilities may be provided within the wallet instance attestation, that's the scope of the current PR (Providing the wallet instance attestation before the issuance of RP's signed request).

openid-4-verifiable-presentations-1_0.md
decentralgabe, line 129:
```suggestion
OpenID for Verifiable Presentations supports the scenario in which the Verifier sends an Authorization Request to  an End-User using either the same or different device from which the requested Credential(s) are stored. 
```

how's this?

tlodderstedt:
@David-Chadwick what is your opinion on the three options I proposed?

David-Chadwick:
@tlodderstedt My opinion is that none of the solutions are optimum because they are not addressing the critical issue, which is "is the verifier trusted". So none of your solutions are what is actually needed. I gave you a much longer answer above explaining why authenticating the verifier is not sufficient, because an authenticated crook is still a crook.

paulbastian:
> @peppelinux @paulbastian @David-Chadwick @danielfett
> 
> It seems the most significant issue right now is the question whether the wallet should send data to the verifier before the verifier is authenticated.
> 
> I see the following options:
> 
> Option 1: sign the discovery request: that is feasible from a technical standpoint. However, what scheme, what trust framework, and which key should the verifier use? The design assume the verifier will have the require data provided in the presentation request. This allows multi homing verifiers, those connected to a US, a EU, and a JP trust framework.
> 
> Option 2: leave the protocol as is and add privacy considerations recommending a user gesture before the presentation request is sent to the verifier (e.g. the unlock of the wallet). That would at least prevent silent finger printing.
> 
> Option 3: send minimal data to the presentation endpoint, e.g. static profile data. Wallet authentication might happen at the direct post.
> 
> Please let me know what you think.

Option 1 looks pretty ugly to me and we have 2 steps of Verifier authentication.
Option 2 is likely to be implemented wrong or may result to double user consent.
I favor Option 3. It changes as few things as possible and gives enough benefits in my opinion while ensuring privacy.

tlodderstedt:
Hi all, in the WG call yesterday we felt it is best to go one step back and decide whether we want to fulfill the following requirement.

**Should the protocol allow the verifier to pick client id scheme and credentials based on the data received from the wallet in the presentation request (step 4 in the sequence diagram)?**

This requirement determines whether authentication (and further steps) can be done in the initial discovery request or not. 

Please share your opinion.

David-Chadwick:
My opinion is that the discovery request message should provide enough information for the user's device to determine if the verifier's asserted name/id can be trusted or not, based on information provided in the discovery request message (which does not need to be authenticated).

tlodderstedt:
@David-Chadwick would you assume the verifier provide different identifers? I'm asking since the client could use different identifiers in different trust frameworks.

bc-pi:
On the Thurs call Torsten suggested that wallet authentication/attestation wasn't need in the presentation flow (which I think I agree with) but that some kind of attestation might be necessary with SIOP for user authentication. That suggests that whatever attestation is needed there should be added in SIOP (maybe part of the id token) but not be part of this work/design.

David-Chadwick:
@tlodderstedt The verifier should know what trust frameworks it is registered in (at least this is a reasonable assumption to make, because there will be some rules that the verifier will have to conform to in order to be recognised as trusted by each framework). So the verifier can send its registered id/names in the discovery request message, as well as the IDs of the trust frameworks (which will need to be globally unique, so URLs could be mandated for these).
The wallet then looks at the sent trust frameworks, see which one(s) it recognises, and then ask it(them) if the provided verifier id/name is trusted. Once the wallet has its answer(s) it can progress from there. This is what we implemented.

peppelinux:
@bc-pi following the friday call with @paulbastian and Franceco Marino I bring here the points about the requirement, from the RP perspective, of obtaining the wallet instance attestation and capabilities before issuing the request object

1. With the will of using both openidvp and siopv2 in a single implementation I suggest to work on a flow that would not introduce too many differences between the two specs. I take into account that the pseudonym authentication should be made with siopv2 id_token and that a user would be able to get authenticated using a pseudonym, instead of presenting a digital credential. For this reason I suggest to keep siopv2 and openid4vp as close as possible, reducing any implementation delta;

2. Considering that the wallet instance supports sign alg B and the RP supports sign algs A, B and C, the RP should know the wallet instance capabilities, through the discovery flow, to sign the request object using the alg B, otherwise it would use by default A and then B and C would not never used, or randomly used. This explain well why the RP should know the capabilities of the wallet instance it is interacting with. I would exclude in this analysis the elusive opportunity to not sign the request object, since the signature of the request object is required by several wallet ecosystem trust models (at least EUDI Wallet);

3. The RP that provides to the wallet instance a URL, within a qrcode for cross device or within a redirect (302 Location URL) within the same device flow, should know the wallet instance's authorization_endpoint. The authorization_endpoint could be a custom url-scheme or an universal link registered within the user mobile device and using general purpose values could raise cross-app security issues; if the user uses the device camera for framing the qr-code and then the request_uri would be processed with a wrong (or colliding) mobile application that's not the one used for the digital identities or specific purposes. For this reason the RP should know the authorization_endpoint of the wallet instance to be sure to ask the interaction with the proper mobile application installed into the user's mobile device;

4. A wallet instance that not provides the wallet instance attestation or an expired wallet instance attestation in the presentation response would fail the authentication to the RP. Why we're designing a flow that introduce this possibility? I believe it is valuable for a good user experience to have the discovery of capabilities before the authentication and presentation of personal data takes place, this allows us to save time and resources for both the user and the RP. Think of attacks based on the use of fake encrypted presentations. The RP would find itself consuming resources, decrypting and subsequently failing to validate the signature and contents, when rather discovery concentrates the resources between the parties in a single objective, that of establishing trust and capabilities, before authentication occurs . This also offers a good division of discovery and authentication purposes, which if it were united and monolithic in the presentation response would bring greater risks to the RP, such as attacks based on resource consumption (CPU) as described in this point.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -228,6 +279,78 @@ OpenID for Verifiable Presentations extends existing OAuth 2.0 mechanisms as fol
...
+: A string identifying the context of the discovery request from the perspective of the Verifier. The value is opqaue to the Wallet, it MUST pass this value to the Presentation Endpoint Request (see below). 
+
+The Discovery Request MUST be represented as a JSON object with all parameters as top-level JSON claims. 
+
+The following is an example request: 
+
+```JSON
+{
+    "presentation_request_uri": "verifier.example.com/presentation_request",
+    "context": "register_kyc"
```
bc-pi, line 301: the parameter name changed from context to interaction, i think? 
```suggestion
    "interaction": "register_kyc"
```

jruizaranguren:
Following the call last wednesday 25th, in some of the proposals there was a need to communicate capabilities or features by on of the participants in an exchange:

* the verifier publishes its capabilities / features over trust frameworks, so a holder can decide if it can fullfil the requirements of that verifier per feature. 
* the holder publishes its capabilites so a verifier can offer presentation definitions tailored to the capabilities of the wallet.

With independence of which option satisfies privacy or functional concerns, in both cases the specification of capabilities / features   must be advertised by one or both participants. Is there any previous attempt to describe such a thing? _I'm verifier X, registered in Y and Z trust framework using W means, and offer the following A, B, C, exchanges using these methods, algorithms, credential formats, etc._

tlodderstedt:
Hi all, I filed a new PR with an alternative design for the advanced flow. It is less sophisticated but simpler and more privacy preserving. I hope it can be the basis of a consensus. Have a look at PR https://github.com/openid/OpenID4VP/pull/59.

awoie:
> Hi all, I filed a new PR with an alternative design for the advanced flow. It is less sophisticated but simpler and more privacy preserving. I hope it can be the basis of a consensus. Have a look at PR #59.
> 
> @peppelinux @paulbastian @David-Chadwick @danielfett @Sakurann

In general I'm supportive of the flow described in #59. Personally, I think we should also include the approach that was described in #65 to allow wallets to detect phishing attacks.

peppelinux:
@tlodderstedt may we close this PR since it is overseeded by https://github.com/openid/OpenID4VP/pull/59?

PR 59 references this PR as well

tlodderstedt:
Closing this PR as the WG is pursuing a different design proposed in PR #59

PR #76, opened 2023-12-29
redirect_uri section, added fragment and small editorials

peppelinux:
This PR fixes some terminology and adds the fragment in the description of the response_code, since the previous text mentioned only url path and param while the non-normative examples contained a fragment

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -592,7 +592,7 @@ The following new parameter is defined for use in the response from the endpoint
 
 Note: Response Mode `direct_post` without the `redirect_uri` could be less secure than Response Modes with redirects. For details, see ((#session_fixation)).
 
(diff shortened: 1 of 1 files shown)


jogu (approved):

tlodderstedt (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -606,9 +606,9 @@ The following is a non-normative example of the response from the Verifier to th
   }
 ```
 
-If the response does not contain a parameter, the Wallet is not required by this specification to perform any further steps.
+If the response does not contain the `redirect_uri`, the Wallet is not required to perform any further steps.
```
Sakurann, line 609:
```suggestion
If the response does not contain the `redirect_uri` parameter, the Wallet is not required to perform any further steps.
```

Sakurann (approved):

Issue #75, opened 2023-12-27
Facilitating Submission of Referenced Claims with Integrity Support using multipart/form-data


jruizaranguren:
## Problem Statement

Certain scenarios necessitate the inclusion of sizable file-based claims within credentials, to be presented during verification. Examples of such claims include scanned documents, subject portraits conforming to ISO 18013-5, and various biometric records.

Verifiers may demand such claims to support situations where legacy processes intersect with digital credential procedures. This is especially relevant for Biometrics-based Holder Binding, as outlined in the [OIDC4VP terminology section.](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-terminology)

The current definition of OIDC4VP allows for the presentation of these claims within the VP token, either through a redirect call or encoded as `application/x-www-form-urlencoded` when Response Mode is set to `direct_post`. However, these options pose challenges when handling large-sized claims, a concern addressed in multiple discussions (see references below).

## Proposal

To address the challenge of managing large claims, I propose enabling the content type `multipart/form-data ` for the submission of Authorization Response in the `direct_post` Response Mode.

By doing so, various mechanisms can be employed to efficiently reference and submit files from verifiable credentials. Different parts of the multipart content can be utilized for the `presentation_submission`, `vp_token`, and additional evidence, ensuring referential integrity extracted from credentials in the `vp_token` using `hashlinks`, `SRI`, or other means. The filenames of each part can be linked to identifiers in the credentials.

This would enable a convenient and efficient to manage claims consisting in large sized files both in issuance and presentation, facilitating different means to retrieve and store such files.

## References

* [Integrity of related resources in W3C VCDM 2.0](https://w3c.github.io/vc-data-model/#integrity-of-related-resources).  Example:

```json
"credentialSubject": {
  "id": "did:example:ebfeb1f712ebc6f1c276e12ec21",
  "image": {
    "id": "https://university.example.org/images/58473",
    "digestSRI": "sha384-ZfAwuJmMgoX3s86L7x9XSPi3AEbiz6S/5SyGHJPCxWHs5NEth/c5S9QoS1zZft+J",
    "mediaType": "application/svg+xml",
  },
  ...
}
```
* [W3C subresource integrity](https://www.w3.org/TR/SRI/)
* [W3C VCDM implementation guide, referencing credentials and evidence recommendations](https://w3c.github.io/vc-imp-guide/#referencing-other-credentials)
[...]

PR #77, opened 2024-01-05
Facilitating Submission of Referenced Claims with Integrity Support using multipart/form-data

piraces:
This PR aims for the proposal in #75.

As stated in this issue, in order to address the challenge of managing large claims, the proposal of enabling `multipart/form-data` for the submission of the Authorization Response in the `direct_post` Response mode is added.

By adding this `Content-Type`, we can efficiently submit and reference files from Verifiable Credentials.

In the changes an example of `POST` request with this `Content-Type` is included to let the users know how they may perform this request.

I also "tried" to add/reference the different scenarios in #75 which are for example the use of `hashlinks` or `SRI`, which (as can be read in the references of the issue) is addressed in multiple discussions.

_**Note:** first contributor to a spec, any kind of feedback is greatly appreciated. Probably I'm missing something or doing something wrong but I'm committed with this._

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -537,7 +537,7 @@ It has been defined to address the following use cases:
 The Response Mode is defined in accordance with [@!OAuth.Responses] as follows:
 
 `direct_post`:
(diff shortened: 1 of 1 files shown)


Sakurann:
thank you for doing the PR to illustrate your proposal. we have not discussed the issue/proposal in the WG, yet - putting `do-not-merge` label. (I have a feeling this example could have fit into the issue description..)

jogu:
I think we need to discuss this further on the issue; I've added a comment there. I propose we close this PR in a week; we can always reopen it once a conclusion is reached on the issue.

jogu:
Closing as per previous comment; let's continue discussion on the issue and potentially reopen once we reach a conclusion.

PR #84, opened 2024-01-24
Enable direct_post* for x509_san_* without redirect_uri in the authz request

### LINKED ISSUE #83: Enable x509_san_dns for direct_post without redirect_uri
For `client_id_scheme=redirect_uri`, we already made an exception for `direct_post` where the `client_id` MUST match the `response_uri` instead. A similar exception should be made for `x509_san_dns` and alike, otherwise those cannot be used with `direct_post` that does not use a final redirect.

awoie:
This PR fixes #83 :
- added similar language for x509_san_* client ID schemes as we already had for redirect_uri client ID scheme
- expanded exceptions for direct_post.jwt


cobward:
Agree with the sentiment of this PR, but I think there are a couple of problems.

1. It is missing the nuance of the requirements for those client_id_schemes: 
> If the Wallet can establish trust in the Client Identifier authenticated through the certificate... **_If not, the FQDN of the `redirect_uri` value MUST match the Client Identifier._**

2. It does not remove the restriction on `redirect_uri`, which no longer serves a purpose.

For this reason I wonder if it is better to amend the restriction in the client_id_scheme definition itself, something like:

> If not, the FQDN of **_the URI where the Wallet submits the response_** MUST match the Client Identifier**_, where "the URI where the Wallet submits the response" is `redirect_uri` or `response_uri` depending on the `response_mode`_**.

I think that wording is bad, but something to that effect?

awoie:
> > Agree with the sentiment of this PR, but I think there are a couple of problems.
> > 
> > 1. It is missing the nuance of the requirements for those client_id_schemes:
> > 
> > > If the Wallet can establish trust in the Client Identifier authenticated through the certificate... **_If not_**, the FQDN of the `redirect_uri` value MUST match the Client Identifier.
> 
> Good catch, I can update the PR to say that:
> 
> ```
> In case the `redirect_uri` is used in conjunction with the `client_identifier`, the `response_uri` MUST be used instead.
> ```
> 
> That should fix that problem.
> 
> > 2. It does not remove the restriction on `redirect_uri`, which no longer serves a purpose.
> 
> There is no restriction on the `redirect_uri` value of the Response URI response parameter in this case. This also applies to `client_id_scheme=redirect_uri` using `direct_post` in conjunction with `response_uri`. Perhaps I misunderstood, what restriction are you missing?
> 
> > For this reason I wonder if it is better to amend the restriction in the client_id_scheme definition itself, something like:
> 
> We already have a similar exception in Response Mode `direct_post` in that section, so I would like to keep it in the same place.

@cobward I updated the PR, please review again.

tlodderstedt (approved):

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -548,6 +548,8 @@ Note: The Verifier's component providing the user interface (Frontend) and the V
 
 Note: If the Client Identifier scheme `redirect_uri` is used in conjunction with the Response Mode `direct_post`, and the `response_uri` parameter is present, the `client_id` value MUST be equal to the `response_uri` value.
 
+Note: If the Client Identifier scheme `x509_san_dns` or `x509_san_uri` is used in conjunction with the Response Mode `direct_post` or `direct_post.jwt`, and the `response_uri` parameter is present, the `response_uri` MUST be used instead of the `redirect_uri` to compute the `client_id` value when applicable.
```
jogu, line 551: I'm not sure 'compute' is the best word to use here. Doesn't the wallet explicitly state the `client_id` and hence it is not necessary to 'compute' a `client_id`?

The "and the `response_uri` parameter is present" part is unnecessary as `response_uri` is mandatory when using direct_post / direct_post.jwt.

I think we should avoid mentioning `direct_post.jwt` here as direct_post.jwt builds on top of direct_post. (If we are going to mention direct_post.jwt in the direct_post section we need to add "or direct_post.jwt" in most of the preceding 6 paragraphs.)

```suggestion
Note: If the Client Identifier scheme `x509_san_dns` or `x509_san_uri` is used in conjunction with the Response Mode `direct_post`, the `response_uri` MUST be a value that the client would be permitted to use as a `redirect_uri` when following the rules defined in (#client_metadata_management).
```

jogu (changes requested):

openid-4-verifiable-presentations-1_0.md
jogu: Although as an alternative how about removing both these notes and simply saying something that is generically applicable to all client id schemes:

> `response_uri` MUST be a value that the client would be permitted to use as a `redirect_uri` when following the rules defined in (#client_metadata_management).

tlodderstedt: That sounds reasonable.

awoie: > Although as an alternative how about removing both these notes and simply saying something that is generically applicable to all client id schemes - I can't imagine any case where response_uri would have a different restriction than redirect_uri does?
> 
> i.e.:
> 
> > `response_uri` MUST be a value that the client would be permitted to use as a `redirect_uri` when following the rules defined in (#client_metadata_management).

I think the above proposal is also not 100% accurate. What about?

> The `response_uri` value MUST be used instead of the `redirect_uri` Authorization Request parameter when following the rules defined in (#client_metadata_management).

IMO, this makes it clear that `redirect_uri` in the Response URI endpoint response can still be different from the `response_uri` Authz Request parameter but the `response_uri` Authz Request parameter has to be used for the `client_id` when applicable as defined in the client ID schemes section.

tlodderstedt: I re-checked the spec. There is no need to state "if the response_uri is present" as it MUST be present in case of response mode "direct_post".

tlodderstedt (changes requested):
I support Joseph's proposal.

openid-4-verifiable-presentations-1_0.md
awoie: You are right. I updated the PR. Please check again.

jogu (approved):

bc-pi (approved):

tlodderstedt (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,12 +542,10 @@ The Response Mode is defined in accordance with [@!OAuth.Responses] as follows:
...
+: OPTIONAL. MUST be present when the Response Mode `direct_post` is used. The Response URI to which the Wallet MUST send the Authorization Response using an HTTPS POST request as defined by the Response Mode `direct_post`. The Response URI receives all Authorization Response parameters as defined by the respective Response Type. When the `response_uri` parameter is present, the `redirect_uri` Authorization Request parameter MUST NOT be present. If the `redirect_uri` Authorization Request parameter is present when the Response Mode is `direct_post`, the Wallet MUST return an `invalid_request` Authorization Response error. The `response_uri` value MUST be a value that the client would be permitted to use as a redirect_uri when following the rules defined in (#client_metadata_management).
```
danielfett, line 545:
```suggestion
: OPTIONAL. MUST be present when the Response Mode `direct_post` is used. The Response URI to which the Wallet MUST send the Authorization Response using an HTTPS POST request as defined by the Response Mode `direct_post`. The Response URI receives all Authorization Response parameters as defined by the respective Response Type. When the `response_uri` parameter is present, the `redirect_uri` Authorization Request parameter MUST NOT be present. If the `redirect_uri` Authorization Request parameter is present when the Response Mode is `direct_post`, the Wallet MUST return an `invalid_request` Authorization Response error. The `response_uri` value MUST be a value that the client would be permitted to use as `redirect_uri` when following the rules defined in (#client_metadata_management).
```

yaromin (approved):

PR #101, opened 2024-02-12
Add one more option to invalid_request if none of the valid options is provided

### LINKED ISSUE #100: Option missing for non-valid OpenID4VP requests
The spec currently says:
> - The request contains more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.

IMO, it is missing that at least one of the options have to be contained.

awoie:
This PR does:
- Covers the case where neither scope, nor one of the presentation_definition/_uri parameters are present.
- Fixes #100 

This PR is non-breaking since we already have the requirement that one of the options must be present.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -659,7 +659,8 @@ The error response follows the rules as defined in [@!RFC6749], with the followi
 
 `invalid_request`:
 
-- The request uses the `vp_token` Response Type but contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
+- The request contains more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
+- The request uses the `vp_token` Response Type but does not request a Credential using any of the three options
 - Requested Presentation Definition does not conform to the DIF PEv2 specification [@!DIF.PresentationExchange].
 - The Wallet does not support the `client_id_scheme` value passed in the Authorization Request.
 - The Client Identifier passed in the request did not belong to the Client Identifier scheme indicated in the Authorization Request, or requirements of a certain scheme was violated, for example an unsigned request was sent with Client Identifier scheme `entity_id`.
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -659,7 +659,7 @@ The error response follows the rules as defined in [@!RFC6749], with the followi
 
 `invalid_request`:
 
-- The request contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
+- The request uses the `vp_token` Response Type and contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
 - Requested Presentation Definition does not conform to the DIF PEv2 specification [@!DIF.PresentationExchange].
 - The Wallet does not support the `client_id_scheme` value passed in the Authorization Request.
 - The Client Identifier passed in the request did not belong to the Client Identifier scheme indicated in the Authorization Request, or requirements of a certain scheme was violated, for example an unsigned request was sent with Client Identifier scheme `entity_id`.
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -659,7 +659,7 @@ The error response follows the rules as defined in [@!RFC6749], with the followi
 
 `invalid_request`:
 
-- The request contains more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
+- The request contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
```
jogu, line 662: We may want to make it clearer that a wallet isn't *required* to return an error in the 'none of' case (but is required to return an error in the "more than one" case). [As the wallet may support other protocols, e.g. SIOPv2]

awoie: In the context of OpenID4VP it is required to provide one of the options, otherwise it is not an OpenID4VP request. Especially, I would expect an error if response_type is vp_token. I agree that for response_type code, the situation is different. If the request is also a SIOPv2 request using vp_token id_token, then I would still expect an error.

awoie: Perhaps we should think about a dedicated scope value that clearly defines that the request is an OpenID4VP request?

awoie: I updated the language in the PR. Would you agree with the new language @jogu ?

Sakurann (changes requested):
`vp_token` response type is not mandatory when response type is `code`. Also, why `none or more`? one of these three options has to be provided for the wallet to know what credential is being requested

awoie:
> `vp_token` response type is not mandatory when response type is `code`. Also, why `none or more`? one of these three options has to be provided for the wallet to know what credential is being requested

The paragraph is about raising an error in case none or more than one option is provided. I agree that there is an exception for response_type code. That is why I added the "The request uses the `vp_token` Response Type and contains none or more than one out of ... " which should read: (vp_token and none) or (>1 options) -> error. I could probably rephrase this.

awoie:
> > `vp_token` response type is not mandatory when response type is `code`. Also, why `none or more`? one of these three options has to be provided for the wallet to know what credential is being requested
> 
> The paragraph is about raising an error in case none or more than one option is provided. I agree that there is an exception for response_type code. That is why I added the "The request uses the `vp_token` Response Type and contains none or more than one out of ... " which should read: (vp_token and none) or (>1 options) -> error. I could probably rephrase this.

Does this now make sense? @Sakurann

openid-4-verifiable-presentations-1_0.md
```diff
@@ -659,7 +659,7 @@ The error response follows the rules as defined in [@!RFC6749], with the followi
 
 `invalid_request`:
 
-- The request contains more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
+- The request uses the `vp_token` Response Type and contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
```
Sakurann, line 662:
```suggestion
- The request uses the `vp_token` Response Type but contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
```
this feels more natural to me..?

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -659,7 +659,7 @@ The error response follows the rules as defined in [@!RFC6749], with the followi
 
 `invalid_request`:
 
-- The request contains more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
+- The request uses the `vp_token` Response Type but contains none or more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
```
jogu, line 662: I think we've ended up in the situation where response=code + presentation_definition + presentation_definition_uri is no longer defined to return invalid_request. I suggest splitting the bullet points up:

```suggestion
- The request contains more than one out of the following three options to communicate a requested Credential: a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a scope value representing a Presentation Definition.
- The request uses the `vp_token` Response Type but does not request a Credential using any of the three options
```

jogu (approved):

PR #102, opened 2024-02-12
Change title to editor's draft

### LINKED ISSUE #70: clarify this is an editor's draft in the title
See https://github.com/openid/OpenID4VCI/pull/126

awoie:
Fixes #70

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1,5 +1,5 @@
 %%%
-title = "OpenID for Verifiable Presentations - draft 20"
+title = "OpenID for Verifiable Presentations - Editor's draft"
 abbrev = "openid-4-vp"
 ipr = "none"
 workgroup = "connect"


c2bo (approved):

paulbastian (approved):

Sakurann:
@tlodderstedt @jogu we need to agree that this is a convention. I think Torsten had some concerns about having to re-generate html before publishing a new draft to openid.net. 
I had concerns that when we published VCI to openid.net @selfissued said that before being published to openid.net, the .md with a latest version number needs to be checked in to the GH. Meaning, every time we want to publish a new draft number to openid.net, there will be two PRs, changing `editors draft` to `- <draft number>` and vice versa. if we agree there is no need to do that, it could maybe work.

jogu:
Having to do two pull requests to update the titles when publishing a numbered revision is potentially an acceptable overhead given there is a definitive advantage of it being clear that versions of the specs generated from the pipelines are not stable versions despite being numbered.

The ultimate aim of OIDF is to automate the publishing process for numbered drafts, so that may then remove the need for manual steps to change the title.

jogu (approved):

tlodderstedt (changes requested):
I don't see any benefit in not having the revision numbers in the markdown. So I would like to stick to the approach of having draft numbers and cutting new draft numbers once we publish a stable WG draft at openid.net.

paulbastian:
The problem that I see is that people will see different versions of the spec over time with the number in the title but with different contents.

Also, people reported that they did not understand what the latest efforts draft is and how to distinguish from stable versions.

selfissued:
I agree with @tlodderstedt that it's cleaner to keep the draft numbers in the spec at all times.

@paulbastian, people will only see the spec changing within a draft number if we point them to the editors' draft.  That's fine within the working group but we shouldn't be sharing those transient drafts outside the working group.  We should be pointing people outside the working group to the stable versions published at openid.net/specs/.

jogu:
> @paulbastian, people will only see the spec changing within a draft number if we point them to the editors' draft. That's fine within the working group but we shouldn't be sharing those transient drafts outside the working group. We should be pointing people outside the working group to the stable versions published at openid.net/specs/.

We do not limit access to the https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html version of the spec to working group members, and people who are not working group members frequently end up there. It is not at all clear to them what the status of the document is. Having it look exactly the same as a version published on openid.net/specs seems like it will cause a lot of confusion.

paulbastian:
I made this suggestion because people from the ISO community told me that they were confused about this and they rely on OpenID4VP, so this is actual, real feedback.

This is my last point on this issue, feel free to do what you think is right.

Sakurann:
I think it is cleaner to have "editor's draft" in openid.bitbucket.io version. My only concern was doing two PRs every single time, but Joseph's comment here convinced me it's ok (guess I am also volunteering him to help with these PRs ;) )

> Having to do two pull requests to update the titles when publishing a numbered revision is potentially an acceptable overhead given there is a definitive advantage of it being clear that versions of the specs generated from the pipelines are not stable versions despite being numbered.
> 
> The ultimate aim of OIDF is to automate the publishing process for numbered drafts, so that may then remove the need for manual steps to change the title.

Sakurann (approved):

tlodderstedt (approved):

jogu:
Editorial, open for many weeks and 5 approvals, merging.

PR #103, opened 2024-02-12
Add well-known client id scheme

### LINKED ISSUE #82: Add client_id_scheme based on .well-known
It would be beneficial to introduce a `client_id_scheme` that
- enables the client to sign Request Objects using a key indirectly trusted via TLS/browser PKI.
- allows the client to provide its metadata via a `.well-known` endpoint rather than in the request itself.

Currently, the only method for implementing request signing leveraging browser PKI is by using `client_id_scheme=x509_san_dns`, which can be quite cumbersome to implement. Clients would need to acquire another certificate from a browser CA for request signing. Additionally, on some platforms, accessing the list of trusted browser CAs outside of a user agent might involve more friction.

A new client ID scheme based on `.well-known` that includes the `jwks`/`jwks_uri` client metadata offers a more straightforward option for implementing the verification of the Request Object's signature.

**Proposal:**
- Introduce a new Client ID scheme value: `well-known`, e.g., `client_id_scheme=well-known`.
- Define it as follows:
  - `client_id` MUST be an HTTPS URI controlled by the client.
  - The wallet MUST fetch the client metadata from `<client_id>/.well-known/client-metadata`.
  - Client metadata from `.well-known` takes precedence over metadata found in the `client_metadata` parameter.
  - Authorization requests can be signed or unsigned. If signed, then the verification keys MUST be provided in the `.well-known` client metadata.

awoie:
This PR:
- Defines a new Client ID scheme `well-known`.
- Introduces no breaking changes.
- Fixes #82


Sakurann:
what happened here?

awoie:
> what happened here?

PR didn't reflect consensus I figured. Will need to put a bit more time in and create a new one.

PR #114, opened 2024-02-22
Relax language on PE

### LINKED ISSUE #113: Relax language around PE evaluation and processing
In certain cases when the PE syntax is known and/or profiled, full PE evaluation and processing is not needed. We introduced very strict language in -20, this language should be relaxed.

awoie:
This PR does:
- relax rules for evaluating and processing PE in case for profiles using PE
- fixes #113 
- contains normative changes to language introduced after ID-2


openid-4-verifiable-presentations-1_0.md
```diff
@@ -234,7 +234,7 @@ The Authorization Request follows the definition given in [@!RFC6749] taking int
...
 
-The Verifier articulates requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. Wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange].
+The Verifier articulates requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. Wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
```
David-Chadwick, line 237: MUST should now be relaxed to SHOULD, because the additional text says when the MUST does not apply

David-Chadwick (changes requested):

openid-4-verifiable-presentations-1_0.md
jogu: > MUST should now be relaxed to SHOULD, because the additional text says when the MUST does not apply

The fact that it's a conditional must doesn't mean it needs to be downgraded to only a recommendation ('should') when the exception doesn't apply. I believe it should stay as a MUST.

Sakurann: +1 to Joseph. the goal here is to say that when there is no credential profile, use as defined in PE, if there is (like 18013-7), use rules defined there.

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
awoie: @David-Chadwick does that work for you?

David-Chadwick: Yes sure

jogu (approved):

David-Chadwick (approved):

c2bo (approved):

PR #115, opened 2024-02-25
Add SD-JWT VC Profile

### LINKED ISSUE #74: Add SD-JWT VC Profile
we need to add the presentation part of the SD-JWT VC profile to OID4VP (https://openid.github.io/oid4vc-haip-sd-jwt-vc/draft-oid4vc-haip-sd-jwt-vc.html#name-presentation-definition) before we can remove the profile from HAIP

javereec:
PR to resolve #74


examples/client_metadata/sd_jwt_vc_verifier_metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "vp_formats": {
+    "vc+sd-jwt": {
+      "sd-jwt_alg_values": ["ES256", "ES384"],
+      "kb-jwt_alg_values": ["ES256", "ES384"]
```
Sakurann, line 5: I think we should keep using `alg` parameter as in other credential formats, so probably something like this:
```suggestion
  "vp_formats": {
    "vc+sd-jwt": {
      "alg": ["ES256", "ES384"]
      },
    "kb+jwt": {
      "alg": ["ES256", "ES384"]
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,90 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+`LTAxLTAxIl0`
+ * Contents:
+`["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
+
+### Verifier Metadata
+
+The Verifier SHOULD add a `vp_formats` element to its metadata (e.g. in the `client_metadata` authorization request parameter) to let the wallet know what protection algorithms it supports in conjunction with SD-JWT VCs. The format element MUST have the key `vc+sd-jwt`, the value is an object consisting of the following elements:
+
+* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for protection of a SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented SD-JWT MUST match one of the array values.
+* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for protection of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
```
Sakurann, line 1602: per the comment above (https://github.com/openid/OpenID4VP/pull/115/files#r1501877810), I think we should use `alg` parameter defined by PE and have two separate objects for `kb+jwt` and `vc+sd-jwt`.

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
Sakurann: > `alg` parameter as in other credential format

I took a deeper look and it looks like despite section 9.1 saying `For specific values that can be used, see [Appendix A](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#alternative_credential_formats).`, appendix A does not define values from client metadata `vp_formats` property. but at the same time, [example in section 5.7](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-5.7-10) contains `alg` claim. we should probably need to decide what approach we are taking across different credential formats.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,90 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+The following is a non-normative example of `client_metadata` request parameter value in a request to present a SD-JWT VC.
+
+<{{examples/client_metadata/sd_jwt_vc_verifier_metadata.json}}
+
```
Sakurann, line 1607: please see the comment above: https://github.com/openid/OpenID4VP/pull/115/files#r1501877810

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
jogu: We currently have this text in VP:

> vp_formats_supported: REQUIRED. An object containing a list of name/value pairs, where the name is a string identifying a Credential format supported by the Wallet. Valid Credential format identifier values are defined in Appendix A of [[OpenID.VCI](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#OpenID.VCI)].

So I think the names under `vp_formats` are required to be Credential format identifiers. "kb+jwt" is not a Credential format identifier (nor should it be) so I don't think @Sakurann's suggestion can be used.

I think my temptation would be to merge this PR as it is, and open a new issue to discuss both this problem and the existing problem other appendix A not defining vp_formats properties.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,90 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+### Presentation Request
+
+The following is a non-normative example of an Authorization Request:
+
+<{{examples/request/request.txt}}
```
bc-pi, line 1612: The data (e.g., nonce, client_id) in this request doesn't align with the data in the respective response/vp_token examples. The examples are non-normative, of course, but they should be reasonably correct. Especially because much of that isn't explicitly defined in text so one has to infer from examples and text elsewhere. 

Related, shouldn't a credential profile explicitly define credential profile specific things? For example, I would have expected to see something saying the nonce and aud in the KB-JWT have to be the nonce and client_id values from the authz request.

danielfett: ...and while discussing this with @bc-pi I noticed that `client_id_scheme` needs to be in the KB-JWT as well and actually in a lot of other places: https://github.com/openid/OpenID4VP/issues/124

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
Sakurann: I am ok externalizing my comment as an issue and merging this PR as-is

openid-4-verifiable-presentations-1_0.md
javereec: I have updated nonce and aud values of the kb-jwt to those of the request. The payload looks now as follows

```
{
  "nonce": "n-0S6_WzA2Mj",
  "aud": "https://client.example.org/cb",
  "iat": 1709839446,
  "sd_hash": "d-m8CkhGgx0KgSgNJvK5mR58GTR2hjhaHbBifUaPhKc"
}
```

I've also added a note in the spec to further clarify values of `aud` and `nonce`.

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
javereec: Should we at least change it to this then before moving. 

```
"vp_formats": {
    "vc+sd-jwt": {
      "alg": ["ES256", "ES384"]
    }
}
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,92 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+<{{examples/response/ps_sd_jwt_vc.json}}
+
+The following is a non-normative example of the `vp_token` parameter provided in the same response and referred to by the `presentation_submission` above:
+
+<{{examples/response/token_response_vp_token_sd_jwt_vc.txt}}
+
+In this example the `vp_token` contains only the disclosures for the claims specified in the `presentation_submission`, along with a Key Binding JWT.
+
+Note: The Key Binding JWT `nonce` claim contains the value of the `nonce` from the presentation request, and the `aud` claim contains the Client Identifier of the Verifier.
```
bc-pi, line 1636:
```suggestion
Note: The Key Binding JWT `nonce` claim contains the value of the `nonce` from the authorization request, and the `aud` claim contains the Client Identifier of the Verifier.
```

openid-4-verifiable-presentations-1_0.md
bc-pi: > I am wondering if it is worth adding some clarification for sd_hash as well? 

Don't think that's needed here. sd_hash is fully specified in SD-JWT itself whereas the values for aud and nonce need to be specified by the protocol using  SD-JWT.

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
awoie: Not sure if this helps but I believe you could also do the following to be more explicit:

```
"vp_formats": {
    "vc+sd-jwt": {
      "kb-jwt": { "alg": [ "ES256", "ES384" ] },
      "issuer-signed-jwt": { "alg": ["ES256", "ES384"] }
    }
}
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,92 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+The following is a non-normative example of an Authorization Request:
+
+<{{examples/request/request.txt}}
+
+The following is a non-normative example of the contents of a presentation_definition parameter that contains the requirements regarding the Credential to be presented:
+
+<{{examples/request/pd_sd_jwt_vc.json}}
+
+The presentation of a SD-JWT VC is requested by adding an object named `vc+sd-jwt` to the `format` object of an `input_descriptor`. The object is empty.
```
awoie, line 1618: Why is the `vc+sd-jwt` parameter always empty?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,92 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+The Verifier SHOULD add a `vp_formats` element to its metadata (e.g. in the `client_metadata` authorization request parameter) to let the wallet know what protection algorithms it supports in conjunction with SD-JWT VCs. The format element MUST have the key `vc+sd-jwt`, the value is an object consisting of the following elements:
+
+* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for protection of a SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented SD-JWT MUST match one of the array values.
+* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for protection of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
+
+The following is a non-normative example of `client_metadata` request parameter value in a request to present a SD-JWT VC.
+
+<{{examples/client_metadata/sd_jwt_vc_verifier_metadata.json}}
+
+### Presentation Request
```
awoie, line 1608: Although obvious but what is missing is that the JSON document for the JSONPath in the `input_descriptor` is the "Unsecured SD-JWT VC".

awoie: Could you please include one sentence that describes this.

Sakurann: I don't understand your point.

openid-4-verifiable-presentations-1_0.md
Sakurann: yeah, shouldn't be always empty.

```suggestion
The presentation of a SD-JWT VC is requested by adding an object named `vc+sd-jwt` to the `format` object of an `input_descriptor`.
```

openid-4-verifiable-presentations-1_0.md
awoie: The Issuer-Signed JWT has the _sd claims in it. You won't be able to run a JSONPath on it. You would need to use the "Unsecured SD-JWT VC" payload instead. This is the term we use in SD-JWT VC.

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
jogu: Discussed on today's WG Call - we agreed to keep the text as Jan currently has it, which is the same as it was in SD JWT VC spec.

awoie (changes requested):
Direction generally looks good but imo, the matcher algorithm is missing. You cannot apply JSONPath using PE to the SD-JWT (i.e., issuer-signed JWT plus disclosures). But the wallet can apply the matcher using the JSONPath to the "Unsecured Payload of an SD-JWT VC" (see https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-03.html#section-1.4). After this was added, I'm fine approving it.

You actually can apply a JSONPath on the SD-JWT VC but it would not give you the expected result since it contains all the `_sd` claims and not the actual disclosures in one JSON object.

Sakurann:
please also review this PR in HAIP that removes sd-jwt vc profile from HAIP https://github.com/openid/oid4vc-haip-sd-jwt-vc/pull/96

Sakurann:
> But the wallet can apply the matcher using the JSONPath to the "Unsecured Payload of an SD-JWT VC"

it might be beneficial to adopt for sd-jwt vc a similar processing requirements as in 18013-7 for mdocs: ""path is a JSON array where each entry is a JSON String containing a requested data element from the requested document type as follows: $['<namespace>']['<data element identifier>']. For example, to request a data element with an data element identifier family_name from the namespace org.iso.18013.5.1, the following JSON String is used: $['org.iso.18013.5.1']['family_name']." cc @bc-pi @danielfett

awoie:
> > But the wallet can apply the matcher using the JSONPath to the "Unsecured Payload of an SD-JWT VC"
> 
> it might be beneficial to adopt for sd-jwt vc a similar processing requirements as in 18013-7 for mdocs: ""path is a JSON array where each entry is a JSON String containing a requested data element from the requested document type as follows: [″][″].Forexample,torequestadataelementwithandataelementidentifierfamilynamefromthenamespaceorg.iso.18013.5.1,thefollowingJSONStringisused:['org.iso.18013.5.1']['family_name']." cc @bc-pi @danielfett

That would work too. Or just say that the root of the JSONPath is the Unsecured SD-JWT VC Payload. I'm ok with both options as long it is defined.

openid-4-verifiable-presentations-1_0.md
jogu: resolving as per discussion on Tuesday's WG call

openid-4-verifiable-presentations-1_0.md
javereec: This was also taken from HAIP. 
If not empty, what type of information can be put in here?

javereec:
@awoie @Sakurann I like the clarity that the unsecured payload brings in terms of finding the disclosures to be added during presentation. I have made some changes, let me know what you think.

openid-4-verifiable-presentations-1_0.md
Sakurann: basically this
```
"sd-jwt_alg_values": ["ES256", "ES384"],
"kb-jwt_alg_values": ["ES256", "ES384"]
```

danielfett:
> > But the wallet can apply the matcher using the JSONPath to the "Unsecured Payload of an SD-JWT VC"
> 
> it might be beneficial to adopt for sd-jwt vc a similar processing requirements as in 18013-7 for mdocs: ""path is a JSON array where each entry is a JSON String containing a requested data element from the requested document type as follows: $['<namespace>']['<data element identifier>']. 
> For example, to request a data element with an data element identifier family_name from the namespace org.iso.18013.5.1, the following JSON String is used: $['org.iso.18013.5.1']['family_name']." cc @bc-pi @danielfett 

That definition is quite confusing and took me a while to process (part of which is caused by the use of "JSON" here). 
If I understood correctly, the expectation is that the JSONPath matching happens on a structure like the following:

```json
{
"org.iso.18013.5.1": {"family_name": "Doe", "given_name": "John"},
"domestic_namespace": {"foo": "bar"}
}
```
Is that right?

How would that fit SD-JWT VC?

Sakurann (approved):

jogu (approved):

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
paulbastian: @jogu do you remember the reason this direction was chosen?

bc-pi:
There are a few places in the Introduction, Terminology, and Overview sections that have text along the lines of this:

> Verifiable Credentials and Verifiable Presentations can be of any format, including, but not limited to W3C Verifiable Credentials Data Model [[VC_DATA](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#VC_DATA)], ISO mdoc [[ISO.18013-5](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#ISO.18013-5)], and AnonCreds [[Hyperledger.Indy](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#Hyperledger.Indy)]. 

which should probably be updated to also mention SD-JWT VC with this PR that adds a profile for it.

examples/client_metadata/sd_jwt_vc_verifier_metadata.json
jogu: Errr, I can't entirely remember, but I believe some combination of:

1. There was no agreement on an idea that was obviously better
2. Without any good suggestion keeping it consistent with HAIP made sense

I think it'd be fine to open an issue if you have a better suggestion, but it doesn't seem worth holding up this PR and an issue allows a more focussed discussion.

examples/credentials/sd_jwt_vc_unsecured.json
```diff
@@ -0,0 +1,6 @@
+{
+  "vct": "https://credentials.example.com/identity_credential",
+  "given_name": "John",
+  "family_name": "Doe",
+  "birthdate": "1940-01-01"
+}
```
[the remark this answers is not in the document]
c2bo: Hmm, true :/ I always like to be sure to add a newline at the end of every line to avoid weird things from happening (some tools break without newline at the end of a line), but let's just keep things how they were in the repo.

Sakurann:
@bc-pi can that be a separate PR..?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+The Credential format identifier is `vc+sd-jwt`.
+
+#### Example Credential
+
+The following is a non-normative example of the unsecured payload of an IETF SD-JWT VC that will be used throughout this section:
+
+<{{examples/credentials/sd_jwt_vc_unsecured.json}}
+
+The following is a non-normative example of an IETF SD-JWT VC using the unsecured payload above, containing claims that are selectively disclosable.
```
paulbastian, line 1566:
```suggestion
The following is a non-normative example of an Issuer-Signed JWT of an IETF SD-JWT VC using the unsecured payload above, containing claims that are selectively disclosable.
```

jogu:
> There are a few places in the Introduction, Terminology, and Overview sections that have text along the lines of this:
> 
> > Verifiable Credentials and Verifiable Presentations can be of any format, including, but not limited to W3C Verifiable Credentials Data Model [[VC_DATA](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#VC_DATA)], ISO mdoc [[ISO.18013-5](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#ISO.18013-5)], and AnonCreds [[Hyperledger.Indy](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#Hyperledger.Indy)].
> 
> which should probably be updated to also mention SD-JWT VC with this PR that adds a profile for it.

As per Kristina's comment opened an issue here to track this: https://github.com/openid/OpenID4VP/issues/143

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+ * SHA-256 Hash: `tiTngp9_jhC389UP8_k67MXqoSfiHq3iK6o9un4we_Y`
+ * Disclosure:\
+`WyI2SWo3dE0tYTVpVlBHYm9TNXRtdlZBIiwgImJpcnRoZGF0ZSIsICIxOTQw`\
+`LTAxLTAxIl0`
+ * Contents:
+`["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
+
+### Verifier Metadata
+
+The Verifier SHOULD add a `vp_formats` element to its metadata (e.g. in the `client_metadata` authorization request parameter) to let the wallet know what protection algorithms it supports in conjunction with IETF SD-JWT VCs. The format element MUST have the key `vc+sd-jwt`, the value is an object consisting of the following elements:
```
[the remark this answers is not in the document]
c2bo: If we change this, then we should also adjust lines 1605+1606

paulbastian (approved):
Approved when my remaining two suggestions are considered

bc-pi (approved):

danielfett (approved):

tplooker (approved):
Approved on the basis we resolve the duplication of supported algorithms as described in #136 separately.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for protection of a SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented SD-JWT MUST match one of the array values.
+* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for protection of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
```
Sakurann, line 1606:
```suggestion
* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for signing of a SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented SD-JWT MUST match one of the array values.
* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for signing of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
Sakurann: made suggestion.

openid-4-verifiable-presentations-1_0.md
javereec: This suggestion introduces a tautology imho as a VC (Verifiable Credential) is already defined as

> An Issuer-signed Credential whose authenticity can be cryptographically verified.

paulbastian: I was trying to differentiate between SD-jwt, disclosures and KB-jwt. The terminology is hard here

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+ * SHA-256 Hash: `tiTngp9_jhC389UP8_k67MXqoSfiHq3iK6o9un4we_Y`
+ * Disclosure:\
+`WyI2SWo3dE0tYTVpVlBHYm9TNXRtdlZBIiwgImJpcnRoZGF0ZSIsICIxOTQw`\
+`LTAxLTAxIl0`
+ * Contents:
+`["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
+
+### Verifier Metadata
+
+The Verifier SHOULD add a `vp_formats` element to its metadata (e.g. in the `client_metadata` authorization request parameter) to let the wallet know what signing algorithms it supports in conjunction with IETF SD-JWT VCs. The format element MUST have the key `vc+sd-jwt`, the value is an object consisting of the following elements:
```
awoie, line 1603: Keep in mind that `vp_formats` is REQUIRED in Verifier metadata.

```suggestion
The `format` value in the `vp_formats` parameter of the Verifier metadata MUST have the key `vc+sd-jwt`, and the value is an object consisting of the following name/value pairs:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for signing of a SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented SD-JWT MUST match one of the array values.
```
awoie, line 1605:
```suggestion
* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the Verifier supports for signing of an Issuer-signed JWT of an SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the Issuer-signed JWT of the presented SD-JWT MUST match one of the array values.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for signing of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
```
awoie, line 1606:
```suggestion
* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the Verifier supports for signing of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+<{{examples/client_metadata/sd_jwt_vc_verifier_metadata.json}}
+
+### Presentation Request
+
+The following is a non-normative example of an Authorization Request:
+
+<{{examples/request/request.txt}}
+
+The following is a non-normative example of the contents of a presentation_definition parameter that contains the requirements regarding the Credential to be presented:
```
awoie, line 1618:
```suggestion
The following is a non-normative example of the contents of a `presentation_definition` parameter that contains the requirements regarding the Credential to be presented:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+
+<{{examples/request/pd_sd_jwt_vc.json}}
+
+The presentation of an IETF SD-JWT VC is requested by adding an object named `vc+sd-jwt` to the `format` object of an `input_descriptor`.
```
awoie, line 1622: I was missing some text that tells the reader where to apply the input_descriptor. Something as follows:

```suggestion
The presentation of an IETF SD-JWT VC is requested by adding an object named `vc+sd-jwt` to the `format` object of an `input_descriptor`. The `input_descriptor` value is applied to the unsecured payload of the IETF SD-JWT VC which correspond to the disclosures of the presented SD-JWT VC.
```

openid-4-verifiable-presentations-1_0.md
javereec: I can see where you are coming from, but when defining it like that the `sd-jwt_alg_values` key as a result makes less sense. Why isn't that then called `issuer-jwt_alg_values`. I really believe that instead of just being able to use SD-JWT, we would now every time need to say which part. "Issuer-signed JWT of an SD-JWT, Key Binding JWT of an SD-JWT, etc).

From the core SD-JWT specification 

> An SD-JWT, at its core, is a digitally signed JSON document containing digests over the selectively disclosable claims with the Disclosures outside the document

it should be clear that when using the term SD-JWT, we mean the Issuer-signed JWT.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1549,6 +1549,96 @@ Note: The reason hashes of the user claims are included in the `issuerAuth` item
...
+`LTAxLTAxIl0`
+ * Contents:
+`["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
+
+### Verifier Metadata
+
+The Verifier SHOULD add a `vp_formats` element to its metadata (e.g. in the `client_metadata` authorization request parameter) to let the wallet know what signing algorithms it supports in conjunction with IETF SD-JWT VCs. The format element MUST have the key `vc+sd-jwt`, the value is an object consisting of the following elements:
+
+* `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the verifier supports for signing of a SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented SD-JWT MUST match one of the array values.
+* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the Verifier supports for signing of a KB-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
```
javereec, line 1606:
```suggestion
* `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms the Verifier supports for signing of a Key Binding JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented Key Binding JWT MUST match one of the array values.
```

openid-4-verifiable-presentations-1_0.md
awoie: I'm actually fine with this as long it is consistent.

openid-4-verifiable-presentations-1_0.md
bc-pi: agree the terminology is hard but think `IETF SD-JWT VC` works here

awoie (approved):
Thanks

PR #118, opened 2024-03-01
Clarifies that alg_values_supported is optional

### LINKED ISSUE #16: what is the point of alg_values_supported in wallet metadata?

Issuer had already signed over holder’s key, so technically there is only one algorithm that the holder can use to sign the Verifiable Presentations \(regardless of the credential format\).

so I don’t think we need alg\_values\_supported in wallet metadata vp\_formats\_supported.

awoie:
This PR fixes #16.

This pull request clarifies that `alg_values_supported` is optional. In certain cases, for example, if there is no prior knowledge of wallet metadata, `alg_values_supported` can be omitted. In such cases, without prior knowledge, the verifier cannot make any assumptions about the cryptographic suites supported by the wallet. The wallet may or may not specific cryptographic suites.


awoie:
Need to create a new PR.

PR #119, opened 2024-03-01
Clarify that alg_values_supported is optional

### LINKED ISSUE #16: what is the point of alg_values_supported in wallet metadata?

awoie:
This PR fixes #16.

This pull request clarifies that `alg_values_supported` is optional. In certain cases, for example, if there is no prior knowledge of wallet metadata, `alg_values_supported` can be omitted. In such cases, without prior knowledge, the verifier cannot make any assumptions about the cryptographic suites supported by the wallet. The wallet may or may not support specific cryptographic suites.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -734,7 +734,7 @@ This specification defines new metadata parameters according to [@!RFC8414].
 
 * `presentation_definition_uri_supported`: OPTIONAL. Boolean value specifying whether the Wallet supports the transfer of `presentation_definition` by reference, with true indicating support. If omitted, the default value is true.
 * `vp_formats_supported`: REQUIRED. An object containing a list of name/value pairs, where the name is a string identifying a Credential format supported by the Wallet. Valid Credential format identifier values are defined in Appendix A of [@!OpenID.VCI]. Other values may be used when defined in the profiles of this specification. The value is an object containing a parameter defined below:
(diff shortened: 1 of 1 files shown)


jogu (approved):

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -734,7 +734,7 @@ This specification defines new metadata parameters according to [@!RFC8414].
...
-    * `alg_values_supported`: An object where the value is an array of case sensitive strings that identify the cryptographic suites that are supported. Parties will need to agree upon the meanings of the values used, which may be context-specific. For specific values that can be used depending on the Credential format, see (#alternative_credential_formats).
+    * `alg_values_supported`: OPTIONAL. An object where the value is an array of case sensitive strings that identify the cryptographic suites that are supported. Parties will need to agree upon the meanings of the values used, which may be context-specific. For specific values that can be used depending on the Credential format, see (#alternative_credential_formats). If `alg_values_supported` is omitted, the wallet MAY or MAY NOT support specific cryptographic suites.
```
Sakurann, line 737: sorry, `MAY or MAY NOT` felt slightly off to me.

```suggestion
    * `alg_values_supported`: OPTIONAL. An object where the value is an array of case sensitive strings that identify the cryptographic suites that are supported. Parties will need to agree upon the meanings of the values used, which may be context-specific. For specific values that can be used depending on the Credential format, see (#alternative_credential_formats). If `alg_values_supported` is omitted, it is unknown what cryptographic suites the wallet supports.
```

Sakurann (approved):
one nit, but rationale makes sense

PR #126, opened 2024-03-05
Require state/nonce to only contain URL safe characters

### LINKED ISSUE #95: non-ASCII/non-url safe characters characters in nonce/state/etc
I feel like we should probably take a position on the use of non-ASCII, and possibly even non-URL-safe characters, in state & nonce - at a minimum recommending that only URL safe characters are used.

state is already limited to ASCII by https://www.rfc-editor.org/rfc/rfc6749#appendix-A however this is no similar limitation for nonce (OpenID Connect just defines it as a "a string".)

I think in practice most experienced OAuth2 practitioners tend to restrict state & nonce to the URL safe character set.

The guidance VP currently gives on creating nonce is:

> (1) The Verifier selects a nonce value as fresh, cryptographically random number with sufficient entropy and associates it with the session.

Something like this would probably be better:

> (1) The Verifier generates 16 fresh, cryptographically random bytes with sufficient entropy, associates it with the session and base64url encodes it for use in the `nonce` parameter.

The advantage of limiting the character set as that it makes implementations/interop easier, as common mistakes in handling character sets (e.g. using Latin1 instead of UTF-8) or incorrectly handling URL escaping/unescaping, have no practical effect.

In contrast if we didn't impose a limitation then the conformance tests probably need to test the full set of characters allowed by the specification, and this is likely to mean the tests are significantly more difficult for implementations to pass. e.g. just yesterday I saw a node.js implementation that was encoding a non-ASCII character in nonce into a JWS in something other than the required UTF-8 and it wasn't obvious why it was doing that or how to fix it, and I've seen characters like ';' or '&' cause issues when used in url queries.

(I'm not sure if there any other places - at least other than the credentials themselves - where non-URL-safe characters are likely to appear in OID4VP?)

jogu:
As discussed in https://github.com/openid/OpenID4VP/issues/95 implementations often make mistakes handling characters that require escaping, character encoding conversion, etc. There is no advantage to using characters outside of the standard URL safe characters, and most people using base64url, UUIDs, JWTs, etc that are by default URL safe, so just make that restriction clear.

closes #95

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -951,7 +951,7 @@ The design is illustrated in the following sequence diagram:
 !---
 Figure: Reference Design for Response Mode `direct_post`
 
-(1) The Verifier generates at least 16 fresh, cryptographically random bytes with sufficient entropy, associates it with the session and base64url encodes it for use in the nonce parameter.
+(1) The Verifier produces a `nonce` value by generating at least 16 fresh, cryptographically random bytes with sufficient entropy, associates it with the session and base64url encodes it.
 
 (2) The Verifier initiates a new transaction at its Response Endpoint. 
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -264,7 +264,7 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
 The following additional considerations are given for pre-existing Authorization Request parameters:
 
 `nonce`:
-: REQUIRED. Defined in [@!OpenID.Core]. It is used to securely bind the Verifiable Presentation(s) provided by the Wallet to the particular transaction. See (#preventing-replay) for details. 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -951,7 +951,7 @@ The design is illustrated in the following sequence diagram:
 !---
 Figure: Reference Design for Response Mode `direct_post`
 
-(1) The Verifier selects a `nonce` value as fresh, cryptographically random number with sufficient entropy and associates it with the session.
+(1) The Verifier generates at least 16 fresh, cryptographically random bytes with sufficient entropy, associates it with the session and base64url encodes it for use in the nonce parameter.
```
Sakurann, line 954:
```suggestion
(1) The Verifier produces a `nonce` value by generating at least 16 fresh, cryptographically random bytes with sufficient entropy, associates it with the session and base64url encodes it.
```

Sakurann (approved):
minor nit - thank you for the PR!

tplooker (approved):

bc-pi (approved):

c2bo (approved):

awoie (approved):
approving again

jogu:
Open for a few weeks, quite a few approvals and mentioned on a few working group calls without any objections been raised - merging.

PR #127, opened 2024-03-05
A minor correction for an example in the Verifier Metadata Management section

adeinega:
This property called as `response_type`.


jogu:
I think this is the same as https://github.com/openid/OpenID4VP/pull/121 - somehow we both looked at the same example today and spotted the same thing...!

adeinega:
Yes, it is. I'm closing this one.

PR #128, opened 2024-03-06
Fixes mdoc/mdl section

### LINKED ISSUE #62: Align with ISO/IEC 18013-7 Annex B
The examples for mdoc/mDL are currently not aligned with the latest Annex B text. We should align as soon as Annex B was published. I'd suggest referencing Annex B instead of redefining the flow for `mso_mdoc`. We can still provide some examples I believe.

awoie:
This PR fixes the mso_mdoc sections by referring to the relevant documents that define the profile.

Fixes #62

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1567,7 +1567,7 @@ See ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 Annex C [@ISO.
 
 ### Presentation Response
 
-The VP Token contains a `DeviceResponse` CBOR structure as defined in ISO/IEC 18013-5:2021 or ISO/IEC 23220-4. `DeviceResponse` is included in the resulting VP Token and signs over the `SessionTranscript` profile defined in ISO/IEC TS 18013-7 [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4].
+The VP Token contains the base64url encoded `DeviceResponse` CBOR structure as defined in ISO/IEC 18013-5:2021 or ISO/IEC 23220-4. `DeviceResponse` signs over the `SessionTranscript` profile defined in ISO/IEC TS 18013-7 [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4].
 
 See ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 Annex C [@ISO.23220-4] for the latest examples on how to use the `presentation_submission` parameter and how to generate the Authorizaton Response for presenting Credentials in the mdoc format.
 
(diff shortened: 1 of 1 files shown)


Sakurann (changes requested):
I don't think there was an agreement in issue #62 to remove examples and only point to ISO specs.

also if you want to mention both 18013 and 23220, much more explanation is needed about the relationship between them. I am afraid the current text would confuse those not familiar.

awoie:
> I don't think there was an agreement in issue #62 to remove examples and only point to ISO specs.
> 
> also if you want to mention both 18013 and 23220, much more explanation is needed about the relationship between them. I am afraid the current text would confuse those not familiar.



> I don't think there was an agreement in issue #62 to remove examples and only point to ISO specs.
> 
> also if you want to mention both 18013 and 23220, much more explanation is needed about the relationship between them. I am afraid the current text would confuse those not familiar.

Ok, let's discuss in the issue first.

On the PR, I can give an intro sentence on what mdocs are similar to what we did for anoncreds. No need to mention more. For that reason the references for interested readers.

awoie:
@Sakurann I tried to incorporate your feedback in the PR. Could you please review again.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1694,6 +1697,8 @@ The technology described in this specification was made available from contribut
 # Document History
 
    [[ To be removed from the final specification ]]
+   -21
```
Sakurann, line 1700:
```suggestion

   -21
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1514,59 +1544,32 @@ The following is the content of the `presentation_definition` parameter:
 
 <{{examples/response/ac_vp_sd.json}}
 
-## ISO mobile Driving License (mDL)
+## Mobile Documents or mdocs (ISO/IEC 18013 and ISO/IEC 23220)
```
Sakurann, line 1547:
```suggestion
## Mobile Documents or mdocs (ISO/IEC 18013 and ISO/IEC 23220 series)
```

Sakurann (approved):
looks pretty good to me. one of the cleanest explanations of the relationship btw 18013, 23220 series and oid4vp that i have seen

jogu (approved):

awoie:
Had to fix some of the references. SessionTranscript is defined in 18013-5/23220-4. OID4VPHandover is defined in 18013-7 and 23220-4. SessionTranscript is always the same for all protocols. OID4VP is just a specific protocol in ISO. Had to adjust the language. For that reason, please review again @Sakurann @jogu .

jogu (approved):

Sakurann (approved):
thank you for requesting re-review

tplooker (approved):

paulbastian (approved):
Looks good to me. Are we sure we want to exclude all examples?

Sakurann:
Yeah, I actually had the same question in the back of my mind... 
examples are super useful but they also depend on 18013-7, so keeping them in oid4vp means dependency on 18013-7 (updating them whenever 18013-7 changes), so probably better to remove them from oid4vp

c2bo (approved):

jogu:
> Looks good to me. Are we sure we want to exclude all examples?

Aside from the very good reason Kristina mentioned, we also don't actually have any valid examples (there's an error in the ones currently in 18013-7) and they'd be a massive pain to generate.

I think omitting them is the correct thing to do at the current time. Whether we might at some future point want to add them back in is a slightly separate discussion.

PR #138, opened 2024-03-27
ed: clarify the usage of response uri with response mode direct_post

### LINKED ISSUE #93: redirect_uri client_id_scheme: is response_uri optional?
When using `client_id_scheme=redirect_uri`, you can omit redirect uri as per https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#client_metadata_management :

> the Verifier MAY omit the redirect_uri Authorization Request parameter

It would be logical to assume that when using `direct_post`  + `client_id_scheme=redirect_uri` you can also omit `response_uri`, however I think the current text is not clear. We should probably be explicit.

Sakurann:
resolves #93.

cc @javereec


jogu (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -441,7 +441,7 @@ This specification defines the following values for the `client_id_scheme` param
 
 * `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed, the Verifier MAY omit the `redirect_uri` Authorization Request parameter, and all Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
+* `redirect_uri`: This value indicates that the Verifier's redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request).
```
javereec, line 444:
```suggestion
* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -629,6 +629,8 @@ The following new Authorization Request parameter is defined to be used in conju
 `response_uri`:
 : REQUIRED when the Response Mode `direct_post` is used. The Response URI to which the Wallet MUST send the Authorization Response using an HTTP POST request as defined by the Response Mode `direct_post`. The Response URI receives all Authorization Response parameters as defined by the respective Response Type. When the `response_uri` parameter is present, the `redirect_uri` Authorization Request parameter MUST NOT be present. If the `redirect_uri` Authorization Request parameter is present when the Response Mode is `direct_post`, the Wallet MUST return an `invalid_request` Authorization Response error. The `response_uri` value MUST be a value that the client would be permitted to use as `redirect_uri` when following the rules defined in (#client_metadata_management).
 
+Note: When the specification text refers to the usage of redirect URI in the Authorization Request, that part of the text also applies when Response URI is used in the Authorization Request with Response Mode `direct_post`.
```
javereec, line 632:
```suggestion
Note: When the specification text refers to the usage of Redirect URI in the Authorization Request, that part of the text also applies when Response URI is used in the Authorization Request with Response Mode `direct_post`.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: the rest of the spec actually uses lower-case `redirect URI`, so if want to change it, we should do it everywhere in another PR.

Sakurann: well, there are 3 occurrences, and you fixed 2 out of them, so i just accepted..

openid-4-verifiable-presentations-1_0.md
```diff
@@ -441,7 +441,7 @@ This specification defines the following values for the `client_id_scheme` param
 
 * `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed, the Verifier MAY omit the `redirect_uri` Authorization Request parameter, and all Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
+* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request).
```
awoie, line 444: In case of direct_post, `response_uri` cannot be omitted, right? Otherwise, you would need to add similar text.

awoie (commented):
Looks good but I had one question and I just want to make sure we are not missing anything.

openid-4-verifiable-presentations-1_0.md
Sakurann: In case of direct_post, `response_uri` can be omitted, only for `client_id_scheme` redirect_uri. what text you are suggesting to add..?

peppelinux, line 444:
```suggestion
* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. Verifier metadata parameters can be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
```

I would not mandate this mechanism, as there could be more advanced methods for providing/fetching verifier's metadata, such as using verifiable attestations or OpenID Federation. Using MUST could restrict the specification by making implementations non-compliant if they adopt alternative approaches.

peppelinux (changes requested):

openid-4-verifiable-presentations-1_0.md
awoie: In that case, I would add the following...

```suggestion
* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
```

selfissued: I agree with Giuseppe's proposed change.

selfissued (dismissed):
I agree with @peppelinux's proposed change.

openid-4-verifiable-presentations-1_0.md
bc-pi: But this is just in the context of the `redirect_uri` client_id_scheme - wouldn't "more advanced methods for providing/fetching verifier's metadata" be defined under a different client_id_scheme? 

Even so, the use of "All" to start that sentence could be read as meaning more than I think it's supposed to mean. Here's a slightly different suggestion:

```suggestion
* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter. Verifier metadata parameters MUST are passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
```

peppelinux: I admit that I never understood why `The Authorization Request MUST NOT be signed`

Sakurann: @peppelinux @selfissued Brian is right. this is only the text for client_id == redirect_uri case, so oid federation and verifier attestation cannot apply. are you ok with Brian's suggestion?

peppelinux: To avoid confusion, I'd suggest to pre-prepend and additional remark, like 

`When `redirect_uri` is used` before ` the Verifier metadata parameters are passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). `

Sakurann: but whether client_metadata parameters are used or not does not depend on redirect_uri but on the client_id_scheme. per PR #165

jogu: @peppelinux re: appending "When redirect_uri is used"

This text is inside a bullet point that's solely for the redirect_uri client scheme. I don't think we need to clarify that we're talking about the redirect uri client scheme, and as Kristina says doing say may actually introduce new confusion.

I'm good with Brian's suggested text, when combined with Oliver's clarification.

peppelinux: in this text we're saying the the client_metadata MUST be passed within the request, while this is not true since there might be several ases where the RP (client) metadata MUST/SHOULD be passed through other method, like thematic trusted lists, openid federation, metadata published at .well-known endpoint and so on depending by the trust framework used.

the most relevant evidence about this is the fact that we cannot pretend that the RP's metadata would not be attested by a trusted third party. This is up to the trust framework used, definitively out of the scopes of the openid4vp specs but necessarly implementable in it.

jogu: I'm struggling to follow. This text is only applicable when `client_id_scheme` is `redirect_uri`, which it's fairly likely no one will use in production, and explicitly means "OpenID Federation is not being used" (because in that case `client_id_scheme` would be `entity_id`).

It is correct to say that the metadata must be in the request for this case. And once #17 is resolved the metadata will be clearly limited to be a pretty small set of metadata.

peppelinux (approved):

peppelinux:
I approve this PR 

Probably, in another PR, we should say something about the value of configuring client_id_scheme differently

it turns out that redirect_uri refers to a "direct trust model"

therefore, as mentioned in other issues, the client_id_scheme parameter is not related about how to interpret the the client id value but to the trust evaluation mechanism to attest the client identity

peppelinux:
Do we suppose to resolve the issue https://github.com/openid/OpenID4VP/issues/33 too using this PR?

jogu:
> Do we suppose to resolve the issue #33 too using this PR?

I think for simplicity we should keep that in a separate PR, as this one is hopefully very close to being merged now.

selfissued (commented):
I withdraw my request for changes.

javereec (approved):
I approve this PR

jogu:
I think we're close on this one - there is one comment from Oliver that needs addressed somehow, and there's some conflicts, but I'm happy to fix the conflicts once we have resolution on Oliver's comment.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -441,7 +441,7 @@ This specification defines the following values for the `client_id_scheme` param
 
 * `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-* `redirect_uri`: This value indicates that the Verifier's redirect URI is also the value of the Client Identifier. In this case, the Authorization Request MUST NOT be signed, the Verifier MAY omit the `redirect_uri` Authorization Request parameter, and all Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request). 
+* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request).
```
Sakurann, line 444:
```suggestion
* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#vp_token_request).
```

PR #146, opened 2024-04-04
ed: change "SIOP" term to "wallet"

### LINKED ISSUE #96: Clarification needed on SIOP and client_metadata paragraph
The current text has the following section:
> Usage of client_metadata or client_metadata_uri parameters with client_id that the Wallet might be seeing for the first time is mutually exclusive with the registration mechanism where Self-Issued OP assigns client_id to the Verifier after receiving Verifier metadata.

Can somebody remind me of the purpose of this paragraph?

Sakurann:
fixes #96.


jogu (changes requested):
We should update this based on the discussion on the issue ( https://github.com/openid/OpenID4VP/issues/96 )

jogu (changes requested):
We should update this based on the discussion on https://github.com/openid/OpenID4VP/issues/96

Sakurann:
the main is outdated. closing to reopen as a new PR for clarity.

PR #165, opened 2024-05-03
clarify what is mutually exclusive (client_metadata and pre-registered clients)

Sakurann:
fixes https://github.com/openid/OpenID4VP/issues/96.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -266,6 +266,8 @@ This specification defines the following new parameters:
 `client_metadata_uri`: 
 : OPTIONAL. A string containing an HTTPS URL pointing to a resource where a JSON object with the Verifier metadata can be retrieved. The scheme used in the `client_metadata_uri` value MUST be `https`. The `client_metadata_uri` value MUST be reachable by the Wallet. It MUST NOT be present if `client_metadata` parameter is present.
 
+`client-metadata` and `client_metadata_uri` parameters MUST NOT be used when the Wallet knows Client Identifier and corresponding Client metadata prior to the transaction (i.e., Client Identifier scheme is absent or has value `pre-registered`).
+
 A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` or `client_metadata_uri` request parameter.
 
 `request_uri_method`: 
@@ -766,8 +768,6 @@ The error response follows the rules as defined in [@!RFC6749], with the followi
 
 - `client_metadata` or `client_metadata_uri` parameter defined in (#vp_token_request) is present, but the Wallet recognizes Client Identifier and knows metadata associated with it.
 - Verifier's pre-registered metadata has been found based on the Client Identifier, but `client_metadata` parameter is also present.
-
-Usage of `client_metadata` or `client_metadata_uri` parameters with `client_id` that the Wallet might be seeing for the first time is mutually exclusive with the registration mechanism where Self-Issued OP assigns `client_id` to the Verifier after receiving Verifier metadata.
 
 `access_denied`:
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -266,6 +266,8 @@ This specification defines the following new parameters:
 `client_metadata_uri`: 
 : OPTIONAL. A string containing an HTTPS URL pointing to a resource where a JSON object with the Verifier metadata can be retrieved. The scheme used in the `client_metadata_uri` value MUST be `https`. The `client_metadata_uri` value MUST be reachable by the Wallet. It MUST NOT be present if `client_metadata` parameter is present.
 
+`client-metadata` and `client_metadata_uri` parameters MUST NOT be used when the Wallet knows Client Identifier and corresponding Client metadata prior to the transaction (i.e., Client Identifier scheme is absent or has value `pre-registered`).
```
jogu, line 269: Looking at this again it looks odd to me.

1. I can't see why we'd separate out `pre-registered`, it seems to me like the same concerns would apply to `entity_id` given that is very very similar to a pre-registered client.

2. It overlaps with the discussion we had about #17 where people were saying the `client_metadata` could be used to pass ephemeral encryption keys, and I can't see why we'd want to disallow that for `pre-registered`.

jogu (changes requested):

openid-4-verifiable-presentations-1_0.md
Sakurann: on 1, I disageee. it is only pre-registered or when client id scheme is absent that requires wallet to know verifier's client metadata, entity_id might be similar, but it does not require wallet knowing a client_id and i think there are use-cases where it might be useful to use client metadata with entity_id scheme.

On 2, can we please take it step by step? i want to resolve #96 first because that sentence is confusing and this PR updates the text. once we have agreement on #17, we can update this.

jogu: > i think there are use-cases where it might be useful to use client metadata with entity_id scheme.

I think there's something that you're seeing that I'm not yet getting - I'm not sure which cases you are thinking of? Or can you explain a case where using client_metadata is very obviously wrong for pre-registered but not for any other client_id_scheme?

I think there are use cases where we need client metadata with pre_registered too? i.e. The ephemeral encryption key case, particularly once you're using https://datatracker.ietf.org/doc/draft-ietf-oauth-attestation-based-client-auth/ in combination with pre-registered client_ids.

I think I'd feel more comfortable removing the sentence at this stage. It feels like we're deciding on one part of #19 but without a full justification why it's sensible.

If we really want to keep the sentence we should fix it, as "Wallet knows Client Identifier and corresponding Client metadata prior to the transaction" is also true for entity_id the second time a wallet sees a particular verifier (federation requires the AS to register the client_id once it's seen it).

tplooker: > I think there are use cases where we need client metadata with pre_registered too? i.e. The ephemeral encryption key case, particularly once you're using https://datatracker.ietf.org/doc/draft-ietf-oauth-attestation-based-client-auth/ in combination with pre-registered client_ids.

I agree with this point, although I think we also need to also go further here e.g 

1. Establish an order of precedent too, for example if there is metadata from registration that takes priority over what is specified in this parameter 
2. We also need to either make this parameter work with only a subset of metadata when the request isn't signed as its not authoritative in that case.

tplooker (changes requested):
I'd prefer we discuss this wider there is more discussion to be had on how this parameter is handled in different contexts.

Sakurann:
ok, let's delete this paragraph for now in this PR (I agree it makes sense to use ephemeral encryption key with pre-registered clients). and I made a note in https://github.com/openid/OpenID4VP/issues/17#issuecomment-2129776329 that we should revisit if the usage of client metadata is affected by client_id_scheme or not to make sure this does not get forgotten.

openid-4-verifiable-presentations-1_0.md
Sakurann: per https://github.com/openid/OpenID4VP/pull/165#issuecomment-2129787074

openid-4-verifiable-presentations-1_0.md
```diff
@@ -266,6 +266,8 @@ This specification defines the following new parameters:
...
+
```
Sakurann, line 270:
```suggestion
```

jogu (approved):

tplooker (approved):

PR #184, opened 2024-05-29
make path_nested path relative in the presentation submission example

### LINKED ISSUE #57: PresentationSubmission example for multiple VPs seems to be wrong
The current example in https://github.com/openid/OpenID4VP/tree/main/examples/response/presentation_submission_multiple_vps.json uses an absolute path of `$[0].verifiableCredential[0]`:

```
"path_nested": {
    "format": "ldp_vc",
    "path": "$[0].verifiableCredential[0]"
}
```

But §7.1 of the [PresentationExchange specification](https://identity.foundation/presentation-exchange/#processing-of-submission-entries) says that the `path` is to be executed on the _current traversal object_ meaning it should be relative to the previous path. See example from PEX:

```
{
  "presentation_submission": {
    "id": "a30e3b91-fb77-4d22-95fa-871689c322e2",
    "definition_id": "32f54163-7166-48f1-93d8-ff217bdb0653",
    "descriptor_map": [
      {
        "id": "banking_input_2",
        "format": "jwt_vp",
        "path": "$.outerClaim[0]",
        "path_nested": {
          "id": "banking_input_2",
          "format": "ldp_vc",
          "path": "$.innerClaim[1]",
          "path_nested": {
            "id": "banking_input_2",
            "format": "jwt_vc",
            "path": "$.mostInnerClaim[2]"
          }
        }
      }
    ]
  }
}
```

nemqe:
resolves #57

### DIFF
--- a/examples/response/presentation_submission_multiple_vps.json
+++ b/examples/response/presentation_submission_multiple_vps.json
@@ -8,7 +8,7 @@
       "path": "$[0]",
       "path_nested": {
         "format": "ldp_vc",
-        "path": "$[0].verifiableCredential[0]"
+        "path": "$.verifiableCredential[0]"
       }
     },
     {


Sakurann (commented):
I think the current examples are correct because the whole point is to say whether it is [0] or [1] in the VP Token..

nemqe:
I think that the `path_nested` needs to be scoped to the upper level of path_nested, i.e. if we have a presentation at index [0] of a `vp_token` the credential we are pointing to needs to be contained in that presentation, and it cannot escape that 'sandbox'.

Maybe I am reading the spec wrong: https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-submission, and I opened this PR to get clarity on this.

Sakurann:
@sudeshrshetty to review

sudeshrshetty:
In OID4VP, `presentation_submission` and VerifiablePresentation(s) are part of different objects unlike those given in [Presentation Exchange Examples](https://identity.foundation/presentation-exchange/#current-traversal-object). 

In case of multiple presentations in `vp_token`, paths like [0], [1] are necessary. 
But `path_nested` property always provides a relative path within a given nested value, so changes in this PR makes sense to me.

sudeshrshetty (approved):

jogu:
This has unfortunately now conflicted with https://github.com/openid/OpenID4VP/pull/198/files - are you able to resolve the conflict please @nemqe ?

nemqe:
@jogu conflict resolved

jogu:
Thank you!

Is it definitely right how it is now:

```
{
  "id": "Presentation example 2",
  "definition_id": "Example with multiple VPs",
  "descriptor_map": [
    {
      "id": "ID Card with constraints",
      "format": "ldp_vp",
       "path": "$[0]",
       "path_nested": {
         "format": "ldp_vc",
         "path": "$.verifiableCredential[0]"
       }
     },
     {
      "id": "Example credential disclosing only address",
      "format": "vc+sd-jwt",
      "path": "$[1]"
    }
  ]
}
```

It's confusing me as I thought `$` basically referred to the root of the response, can that be accessed as both an array (`$[1]`) and a object (`$.verifiableCredential`) ?

nemqe:
My understanding is that when `path_nested` you are 'sandboxed' inside of the object specified by `path`, and that your path is always relative to the 'nesting'.

In this particular example it would mean that `path_nested.path=$` is not referring to the root, but to the `path=$[0]`.

There is a section describing `path_nested` and an example in https://identity.foundation/presentation-exchange/spec/v2.0.0/#presentation-submission

PR #195, opened 2024-06-14
remove path_nested description from Response Parameters section and m…

### LINKED ISSUE #63: Explanation of presentation_submission should be independent of credential format
The current description in section 6.1. is tight to W3C VCs.

Here is one example

> The path_nested object inside an Input Descriptor Mapping Object is used to describe how to find a returned Credential within a Verifiable Presentation, and the value of the path field in it will ultimately depend on the credential format. 

The text needs to be modified to be format agnostic. Additionally, we might want to move format specific text into the W3C VC profile.

paulbastian:
…ove it into W3C VC Annex

resolves #63


Sakurann (approved):

peppelinux (approved):

jogu (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -575,7 +575,7 @@ When a VP Token is returned, the respective response MUST include the following
...
+: REQUIRED. The `presentation_submission` element as defined in [@!DIF.PresentationExchange]. It contains mappings between the requested Verifiable Credentials and where to find them within the returned VP Token. This is expressed via elements in the `descriptor_map` array, known as Input Descriptor Mapping Objects. These objects contain a field called `path`, which, for this specification, MUST have the value `$` (top level root path) when only one Verifiable Presentation is contained in the VP Token, and MUST have the value `$[n]` (indexed path from root) when there are multiple Verifiable Presentations, where `n` is the index to select. Additional parameters may be defined by Credential Formats, see (#alternative_credential_formats) for details.
```
Sakurann, line 578:
```suggestion
: REQUIRED. The `presentation_submission` element as defined in [@!DIF.PresentationExchange]. It contains mappings between the requested Verifiable Credentials and where to find them within the returned VP Token. This is expressed via elements in the `descriptor_map` array, known as Input Descriptor Mapping Objects. These objects contain a field called `path`, which, for this specification, MUST have the value `$` (top level root path) when only one Verifiable Presentation is contained in the VP Token, and MUST have the value `$[n]` (indexed path from root) when there are multiple Verifiable Presentations, where `n` is the index to select. Additional parameters can be defined by Credential Formats, see (#alternative_credential_formats) for details.
```

PR #197, opened 2024-06-18
add transaction_data

### LINKED ISSUE #172: Where in the presentation request to include `transaction_data` objects
(separating a discussion started in https://github.com/openid/OpenID4VP/issues/156#issuecomment-2098575876 into a new issue)

On a May-14-2024 WG call, there seems to be a rough consensus that `transaction_data` should be outside the `input_descriptor` objects and should contain reference to the input_descriptors that are requesting credentials to which the content of a `transaction_data` can be bound to (instead of putting transaction data obejct inside an input_descriptor, which was an original idea). This updated approach would prevent duplication of the same transaction data object across multiple input descriptors when the verifier is asking to "bind `transaction_data` to credential A or credential B".

opening an issue to make sure there is a consensus on this direction, expecting to be able to close this issue relatively soon.

### LINKED ISSUE #173: How to encode transaction_data in the presentation request
(opening a new issue that emerged as part of the discussion of https://github.com/openid/OpenID4VP/issues/172)

it has been brought up that encoding of `transaction_data` needs to be designed so that it helps the verifier to check that transaction data that the wallet includes in the response has not been modified. There seems to be multiple options:

1. (current design) `transaction_data` is array of objects in the request and the response. the verifier needs to check if the `transaction_data` object(s) in the response have not been modified, which is not exactly the simplest.
2. `transaction_data` is array of strings in the request, where each string is a base64url encoded object of a distinct transaction data type. the response has two options:
    a. wallet includes the exact same string (base64url encoded object) in the response
    b. wallet includes **the hash of the base64url encoded string in the response** (need to understand what would be the value of hashing)
3. `transaction_data` is array of objects in the request, where each object contains type value, and a string that is a base64url encoded typed object. during the WG call, there have already been a concern raised that mixing up encoded and non-encoded parameters is not clean/complicates processing.
4. `transaction_data` itself is a string that is a base64url encoded array of multiple objects with different transaction data types. This probably prevents the use-case when the verifier is passing multiple transaction data types and is asking to bind each of them to different credentials (ie use credential A for qes_authorization and credential B for payment_confirmation). probably same option to hash or not as in 2a and 2b:
    a. string in the response
    b. hash in the response

### LINKED ISSUE #174: where in the response to include `transaction_data`
two options:
1. Current design is to include `transaction_data` (object or string depending on the outcome of #173) in the presentation, which is credential format specific (ie KB JWT in SD-JWT, session transcript in mdoc, etc). 
2. Alternative that has been proposed is to define a new thingy (something like a transaction_data_token) that includes `transaction_data` and is signed using the key that is used for cryptographic key binding of credential presentation to which this transaction data is bound to.

Option 1 would require each credential format to define how to include transaction data in the response. Option 2 would require defining how to express when multiple transaction data types have been requested to be bound to different credentials(ie use credential A for qes_authorization and credential B for payment_confirmation).

Sakurann:
resolves #174 
resolves #173 
resolves #172 

- [x] need to clarify how the hashing function is communicated
- [x] add changelog
- [x] clarify that transaction data type might require a credential specifically issued for that purpose
- [x] opened #259 on mdocs
- [x]  might add examples


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1749,6 +1771,8 @@ Setting `limit_disclosure` property defined in [@!DIF.PresentationExchange] to `
 
 A non-normative example of the Authorization Response would look the same as in the examples of other Credential formats in this Annex.
 
+`transaction_data` response parameter defined in (#transaction-data) MUST be included in the Key Binding JWT as a top level parameter.
```
jogu, line 1774: The link is invalid, I believe `-` should be `_`: 
```suggestion
`transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level parameter.
```

bc-pi: I dare say @jogu is correct in his analysis here.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -272,7 +272,13 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a Base64url encoded object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST refuse to process any unknown transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
+
+* `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification.
+* `input_descriptor_ids`: REQUIRED. Array of strings each pointing to an Input Descriptor that identifies a request for a Credential that the Verifier is requesting transaction data in a particular object to be bound to.
```
bc-pi, line 281: with the likely coming of the new query language (i.e., #220), do we want to further ingrain the usage of input descriptor at this point?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -746,7 +752,15 @@ The following is a non-normative example of the payload of the JWT used in the e
...
 
-## Error Response
+## Transaction Data {#transaction_data}
+
+Transaction data mechanism enables binding between the user identification/authentication and user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
+
+The Wallet that received `transaction_data` parameter in the request, MUST include in the response a `transaction_data` parameter defined below: 
+
+* `transaction_data`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format Profile, but it has to be included in the  mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
```
bc-pi, line 761: I think I might know where the words "ensures the integrity of, and maps to, the respective" came from :) 

I know one is in the request and one is somewhere in the response but using the same name `transaction_data` for the things in the request and the hashes of those things in the response feels erroneous. Maybe something like `tx_data_hashes` for this response one. Or because it has to be defined by each Credential Format Profile anyway, let the name be defined there too. Maybe.

openid-4-verifiable-presentations-1_0.md
bc-pi: also a  top level parameter in a JWT is a claim so this is a place where claim would be the right term vs. parameter 

```suggestion
`transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: depends on the timeline, I guess? if query language PR gets merged first, we probably need to add another parameter here.

openid-4-verifiable-presentations-1_0.md
Sakurann: good point! I think for interop purposes, this parameter should be defined in VP

```suggestion
* `tx_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format Profile, but it has to be included in the  mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
```

bc-pi: That suggested change https://github.com/openid/OpenID4VP/pull/197#discussion_r1709406435 works for me. But note it would also need a change to what is currently line on 759.

openid-4-verifiable-presentations-1_0.md
bc-pi: yeah, fair point. I was thinking in the other direction though. And knowing how darn hard it is to remove things once they are in, I'd suggest removing this and leaving it implicit for the initial support of the feature.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -746,7 +752,15 @@ The following is a non-normative example of the payload of the JWT used in the e
...
 
-## Error Response
+## Transaction Data {#transaction_data}
+
+Transaction data mechanism enables binding between the user identification/authentication and user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
+
+The Wallet that received `transaction_data` parameter in the request, MUST include in the response a `tx_data_hashes` parameter defined below: 
+
+* `tx_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format Profile, but it has to be included in the  mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
```
Sakurann, line 761: need to more explicitly define either in this PR or in a follow up PR how to include this parameter in KB JWT and mdoc deviceResponse

openid-4-verifiable-presentations-1_0.md
```diff
@@ -746,7 +752,15 @@ The following is a non-normative example of the payload of the JWT used in the e
 
 <{{examples/response/jarm_jwt_vc_json_body.json}}
 
-## Error Response
+## Transaction Data {#transaction_data}
+
+Transaction data mechanism enables binding between the user identification/authentication and user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
```
danielfett, line 757:
```suggestion
The transaction data mechanism enables a binding between the user's identification/authentication and the user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -746,7 +752,15 @@ The following is a non-normative example of the payload of the JWT used in the e
...
+
+The Wallet that received `transaction_data` parameter in the request, MUST include in the response a `tx_data_hashes` parameter defined below:
```
danielfett, line 759:
```suggestion
The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `tx_data_hashes` parameter defined below: 
```

openid-4-verifiable-presentations-1_0.md
danielfett, line 761: Lacks description of how the hashes should be encoded and which hash function to use.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -791,6 +805,14 @@ This document also defines the following additional error codes and error descri
...
 - The value of the `request_uri_method` request parameter is neither `get` nor `post` (case-sensitive).
 
+`invalid_transaction_data`:
+
+- any of the following are true of the objects in the transaction_data structure:
+  - contains an unknown transaction data type value,
+  - is an object of known type but containing unknown fields,
+  - contains fields of the wrong type for the transaction data type,
+  - contains fields with invalid values for the transaction data type, or
+  - is missing required fields for the transaction data type.
```
danielfett, line 815:
```suggestion
- any of the following is true for at least one object in the transaction_data structure:
  - contains an unknown transaction data type value,
  - is an object of known type but containing unknown fields,
  - contains fields of the wrong type for the transaction data type,
  - contains fields with invalid values for the transaction data type, or
  - is missing required fields for the transaction data type.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1749,6 +1771,8 @@ Setting `limit_disclosure` property defined in [@!DIF.PresentationExchange] to `
 
 A non-normative example of the Authorization Response would look the same as in the examples of other Credential formats in this Annex.
 
+`transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim.
```
danielfett, line 1774:
```suggestion
The `transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim.
```

danielfett, line 1774: Should this refer to `tx_data_hashes`?

openid-4-verifiable-presentations-1_0.md
danielfett, line 759: Why add this to the response and not just to the KB-JWT as defined below? If that is what is meant, this should be clarified here.

danielfett (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -272,7 +272,13 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request. 
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a Base64url encoded object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST refuse to process any unknown transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
+
+* `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
* `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: we will need `input_desscriptor_id` as long as presentation exchange is in the specification and we do not have an agreement/timeline to remove presentation exchange, yet until we have that, this kind of parameter is needed for interoperability. and from a priority perspective, transaction data is higher than a new query language right now since there are quite some implementers who are waiting to use transaction data with existing query language.

so 2 options:
1. keep the parameter name `input_descriptor_id` and add a new parameter for a new query language with an equivalent meaning once that query language is in
2. rename this parameter right now to be more general so it can be used with both query languages

option 2 might be better and less changes for the implementations in the future

Sakurann:
```suggestion
* `requested_credential_ids`: REQUIRED. Array of strings each pointing to an Input Descriptor that identifies a request for a Credential that the Verifier is requesting transaction data in a particular object to be bound to.
```
is this general enough?

openid-4-verifiable-presentations-1_0.md
Sakurann: the parameter is common throughout multiple credential formats and the only difference is where this credential is included. will clarify this

```suggestion
The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `tx_data_hashes` parameter defined below. Where `tx_data_hashes` parameter is included in the response is specific to each credential format and is defined in a respective Credential Format Profile.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: also Steven suggested to rename it to `transaction_data_hashes` because `tx_data` has associations different to transaction data.

Sakurann: yes.

just a draft text from sd-jwt :)

The claim `transaction_data_alg` indicates the hash algorithm used to generate the digests. If the claim is not present, a default value of sha-256 MUST be used.

The hash algorithm identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [[IANA.Hash.Algorithms](https://drafts.oauth.net/oauth-selective-disclosure-jwt/draft-ietf-oauth-selective-disclosure-jwt.html#IANA.Hash.Algorithms)] or a value defined in another specification and/or profile of this specification.

To promote interoperability, implementations MUST support the sha-256 hash algorithm.

openid-4-verifiable-presentations-1_0.md
bc-pi: works for me

bc-pi: don't hate it

Sakurann: discussed in the WG, since we do not expect both presentation_definition and credentials parameters sent together, so rough consensus to go option 2.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -272,7 +272,13 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a Base64url encoded object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST refuse to process any unknown transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
+
+* `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
+* `requested_credential_ids`: REQUIRED. Array of strings each pointing to an Input Descriptor that identifies a request for a Credential that the Verifier is requesting transaction data in a particular object to be bound to.
```
Sakurann, line 281:
```suggestion
* `credential_ids`: REQUIRED. Array of strings each pointing to an Input Descriptor that identifies a request for a Credential that the Verifier is requesting transaction data in a particular object to be bound to.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1749,6 +1771,8 @@ Setting `limit_disclosure` property defined in [@!DIF.PresentationExchange] to `
 
 A non-normative example of the Authorization Response would look the same as in the examples of other Credential formats in this Annex.
 
+The `transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim.
```
Sakurann, line 1774: need clarify that when transaction_data is used, KB JWT must be present.

openid-4-verifiable-presentations-1_0.md
bc-pi: also generalize "Input Descriptor"?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -272,7 +272,13 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
-This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request.   
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request. 
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a Base64url encoded object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST refuse to process any unknown transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
```
jogu, line 278: > The Wallet MUST refuse to process any unknown transaction data type

This makes it hard when you don't know up front what a wallet might support or if we want to define new signing mechanisms and have a transition periods where verifiers can send both.

I think we need to clarify 'MUST refuse' too (it could be interpreted as 'ignore' or as 'return an error').

Perhaps:

 "The wallet MUST return an error unless it supports at least one of the transaction data types"?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -272,7 +272,13 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
+
+* `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
```
jogu, line 280: I'd suggest indenting this and the next line (adding 4 spaces before the `*`) just to make it clearer they belong to the transaction data definition.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
jogu: I think (as per an above comment) it should be:

> if the wallet does not support transaction_data it MUST return an error.

paulbastian (changes requested):
I still believe that putting transaction data into a 
a) a dedicated response_type and/or
b) using a dedicated Credential format would be a better separation.

vp_token and existing Credential Formats assume a three party model, while transaction authorization is a 2 party model, it seems this may create problems in the future that we can't foresee, I would appreciate a better logical separation.

I don't have explicit counter-proposals, just a bad feeling. If there is no support for this direction, you may override my disapproval.

Sakurann:
- ToDo: need to clarify which credential is allowed to be used for transaction authorization and for what kind of transaction can be authorized (might not make sense for all types). Issuer would express it in a credential and that should be standardized as part of a credential format definition and not part of OpenID4VP - part of implementation consideration?

openid-4-verifiable-presentations-1_0.md
c2bo: Should we directly define the format specific parts in Appendix B?

paulbastian:
In LSP POTENTIAL there are two designs for QES with Transaction data:
- the Wallet-centric model 
- the QTSP-centric model

Are these flows applicable for both?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -272,7 +272,13 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
 
 If the Verifier set the `request_uri_method` parameter value to `post` and there is no other means to convey its capabilities to the Wallet, it SHOULD add the `client_metadata` parameter to the Authorization Request. 
-This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request.   
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request. 
+
+`transaction_data`:
```
alenhorvat, line 277: Hi,
questions:

a) Is transaction data different from a consent?
b) Transaction data needs to be displayed to the end user, except if it's a M2M exchange. Keeping it generic, can lead to implementation issues. Should at least "display" or similar parameter be defined?
c) Should the RAR or PE (presentation definition?) syntax be reused? (https://datatracker.ietf.org/doc/html/rfc9396#name-authorization-request)
d) Explicit yes/no: default mode is that the user must authorise all actions, or the transaction fails? 
e) If user declines 1 action, should it be recorded?
f) Is there a need for more elaborate transactions, such as conditional statements (e.g., you must answer all, you must answer to some of the questions), supporting multiple options (e.g., do you want to tip 5, 10 or 20%)? Or for QES: select a signature profile (e.g., advanced, qualified).

We introduced something similar about 2 years ago for our B2B/M2M VP exchange (https://hub.ebsi.eu/conformance/learn/verifiable-presentation-exchange#service-to-service-token-flow). We defined a simple key:value approach where, where key is a scope, and value is a VP definition. I believe this matches your definition of transaction_data. In our model, for a given action all conditions must be fulfilled, no conditional statements, no multiple options (like tipping). 

Thank you!

Sakurann: Hi Alen,
a) transaction data is about authorization. the end user agrees to authorize a certain transaction (payments, QES, etc.)
b), c), f) specific details of transaction data are left to the experts of each use-case. so they will not be defined in openid4vp
d) yes, there is not partial succcess. we agreed to make that clearer. I just have not reflected in the PR text yet..
e) recorded where? in the wallet? according to eidas regulation, i think so?

I will look at the pointer! thank you!

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
    * `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: @bc-pi I tried generalize - please review

openid-4-verifiable-presentations-1_0.md
Sakurann: we said that either wallet fulfills all requirements in transaction data, or none, so it would probably be more appropriate to say "The wallet MUST return an error unless it supports all of the transaction data types"

Sakurann:
```suggestion
: OPTIONAL. Array of strings, where each string is a base64url encoded JSON object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST return an error if a request contains even one unrecognized transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
```

we actually use `base64url encoded` throughout the text.

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
The `transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim. This means that transaction data mechanism cannot be used with SD-JWT VCs that do not have cryptographic key binding and, therefore, do not use KB JWT.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -746,7 +752,18 @@ The following is a non-normative example of the payload of the JWT used in the e
...
 
-## Error Response
+## Transaction Data {#transaction_data}
+
+The transaction data mechanism enables a binding between the user's identification/authentication and the user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
+
+
+The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `tx_data_hashes` parameter defined below. If the wallet does not support `transaction_data` parameter, it MUST return an error.
+
+Where in the response `tx_data_hashes` parameter is included is specific to each credential format and is defined in a respective Credential Format Profile.
```
Sakurann:
```suggestion
The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `transaction_data_hashes` parameter defined below. If the wallet does not support `transaction_data` parameter, it MUST return an error.

Where in the response `transaction_data_hashes` parameter is included is specific to each credential format and is defined in a respective Credential Format Profile.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1749,6 +1774,8 @@ Setting `limit_disclosure` property defined in [@!DIF.PresentationExchange] to `
 
 A non-normative example of the Authorization Response would look the same as in the examples of other Credential formats in this Annex.
 
+The `transaction_data` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim. This means that transaction data mechanism cannot be used with SD-JWT VCs without cryptographic key binding and, therefore, do not use KB JWT.
```
Sakurann:
```suggestion
The `transaction_data_hashes` response parameter defined in (#transaction_data) MUST be included in the Key Binding JWT as a top level claim. This means that transaction data mechanism cannot be used with SD-JWT VCs without cryptographic key binding and, therefore, do not use KB JWT.
```

paulbastian:
> > In LSP POTENTIAL there are two designs for QES with Transaction data:
> >
> >  the Wallet-centric model
> >  the QTSP-centric model
> >
> > Are these flows applicable for both?
> 
> @paulbastian I would say so, because in both flows, presenting a credential from the wallet is used as a way to authenticate the user, right? do you have a hypothesis why this would not work for the wallet-centric model?

I would say the workings of the QTSP centric model does not align with the general model of OpenID4VP and what most readers here would assume. For example, it wouldn't allow the real relying party to request a credential and QES in the same request, as the actual request is from QTSP embedded into relying party website.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -269,7 +269,17 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
-This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request.   
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. If the Verifier uses the `client_id_scheme` parameter in the Request Object, it MUST also add the same `client_id_scheme` value in the Authorization Request. 
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a base64url encoded JSON object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST return an error if a request contains even one unrecognized transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
```
jogu, line 275: I have a feeling there's an unintended consequence here between this PR (if merged after 1.0) and the extensibility PR.

I think this language says "a wallet MUST fail a request that includes a `transaction_data` if it doesn't support transaction data". But the other PR says "wallets supporting 1.0 of this spec must accept requests that contain unknown elements, including `transaction_data`.

I think we need the first behaviour - i.e. particularly when using the browser api, a wallet that doesn't support transaction_data must never say it can satisfy a request that contains transaction_data.

If we agree on that, then I can see two solutions:

1. We add `transaction_data` to 1.0 with language like "wallets MUST reject claims that contain this element".
2. We add something like '[crit](https://www.rfc-editor.org/rfc/rfc7515#section-4.1.11)', but for the body, define that in 1.0, and when 1.1 comes along with transaction_data support then verifiers that want to use it are required to include `"crit": [ "transaction_data" ]` in the request payload.

openid-4-verifiable-presentations-1_0.md
bc-pi: thank you

openid-4-verifiable-presentations-1_0.md
```diff
@@ -269,7 +269,17 @@ A public key to be used by the Wallet as an input to the key agreement to encryp
...
+
+    * `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
+    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (Input Descriptor in Presentation Exchange).
```
bc-pi, line 278: adding one little generalization to the generalization you did (thanks for that generalization btw) 
```suggestion
    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (e.g., Input Descriptor in Presentation Exchange).
```

paulbastian:
@Sakurann I would like to better understand the QTSP centric model for QES.
I imagine a scenario where I am on a website (OpenID4VP Relying Party) and it wants me to sign a contract.
How does an Authorization Request look like in this case, especially what is the client_id and request_uri?

paulbastian:
This PR adds a significant feature that should be described in a use case appendix, similar to what [OpenID4VCI](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#appendix-C) does in my opinion.

paulbastian:
I am looking at this from the perspective using this within eIDAS QES use case. Illustrating with this picture, there are two possible flows:
- QTSP centric flow
- Wallet centric flow
image

The QTSP centric flow and how transaction data is used is very clear to me.

The Wallet centric flow is less clear to me. As I imagine this, the communication between Wallet and QTSP look quite similar with the Wallet having a fixed relation with a particular QTSP. The question arises how a Verifier would send a request for a QES (either including the hash or the document itself) to the Wallet, marked with "?" in the diagrams. Questions:
- Should we use OpenID4VP for this? (I think yes)
- Should this be done be using a new Credential Format? Should this use transcaction_data or credential-specific parameter to communicate the hash/document? (I am not sure and would like to clarify if transaction_data applies here or not)

Sakurann:
@paulbastian regarding wallet-centric flow... I re-read LSP POTENTIAL UC5 QES diagrams and if I understand correctly, the difference between QTSP centric and Wallet centric is where the Signature Creation Application is located ouside or inside the EUDIW and whether the user chooses QTSP to use from the RP or from the Wallet. Which means that the part where QTSP authenticates the user using the wallet is the same between both flows, hence how transaction data is used is also the same. Am I missing something?

paulbastian:
If my question wasn't clear enough: what do we intend to use for "?" in my figure for the wallet centric flow?

alenhorvat:
> If my question wasn't clear enough: what do we intend to use for "?" in my figure for the wallet centric flow?

For ? the wallet should receive a request to share data/VC/... signed with one of the signature profiles requested or supported by the RP. So OID4VP would be a good fit.
If wallet has a web interface, RP -> QTSP and RP-> (Web) Wallet could use the same protocol/interfaces.

I'm not sure the QTSP-specific API will be symmetric as above. At least today we see the following patterns

QTSP --> Wallet (1st flow) (today this is implemented in a closed setting where QTSPs are using push notifications)
Wallet --> QTSP (2nd flow) (today oauth or other client-secret based approaches are used, again, having in mind a closed system)

If this needs to be opened up, QTSPs would need to align on the interfaces (I guess this is what CSC is trying to achieve?)

Sakurann:
> If my question wasn't clear enough: what do we intend to use for "?" in my figure for the wallet centric flow?

probably CSC specification etc. but I really don't think that's in scope for this PR and need to hold it up.

openid-4-verifiable-presentations-1_0.md
Sakurann: > I think we need the first behaviour - i.e. particularly when using the browser api, a wallet that doesn't support transaction_data must never say it can satisfy a request that contains transaction_data.

I think we already have a statement like your first option? but I can clarify extensibility sentence does not apply to transaction data

Sakurann: tried to clarify in this commit https://github.com/openid/OpenID4VP/pull/197/commits/3672f78a6d9211ba8d162e163d90e3a2deefeadd

openid-4-verifiable-presentations-1_0.md
Sakurann: ok, how RP records is probably out of scope of this PR.
I am not sure there is any specific change for the PR requested? resolving for now. please feel free to open issue(s) if anything is not addressed

paulbastian (commented):
I'm not objecting

jogu:
@paulbastian 

> I'm not objecting

Paul, did you mean to clear your 'requested changes' then? It's still showing currently.

paulbastian:
> @paulbastian 
> 
> > I'm not objecting
> 
> Paul, did you mean to clear your 'requested changes' then? It's still showing currently.
> 

Yes, it seems there is no way to clear "request changes" other than to approve. But I didn't have enough time yet to read the PR in all its details to feel confident to approve.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,18 @@ This specification defines the following new request parameters:
...
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a base64url encoded JSON object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST return an error if a request contains even one unrecognized transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
+
+    * `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
+    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (e.g., Input Descriptor in Presentation Exchange).
```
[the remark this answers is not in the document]
Sakurann: I think the intended behavior is, use one of the credentials referenced by the values in the array.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,18 @@ This specification defines the following new request parameters:
...
+    The claim transaction_data_alg indicates the hash algorithm used to generate the digests. If the claim is not present, a default value of sha-256 MUST be used.
+    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
```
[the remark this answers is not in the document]
Sakurann: > does it mean verifiers must include sha-256 in transaction_data_hashes_alg if they include that parameter?

`If this parameter is not present, a default value of `sha-256` MUST be used.` and if a verifier wants to explicitly communicate, it can include sha-256, but no such obligation.

> Or does it just mean wallets must support that value?
`To promote interoperability, implementations MUST support the sha-256 hash algorithm.` so both verifiers and wallets

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,18 @@ This specification defines the following new request parameters:
...
+
+Each document defining a transaction data type specifies whether that type requires a Credential that is bound to transaction data to be specifically issued for that purpose. How the Credential Issuer expresses that a particular Credential can be used for obtaining user's authorization using transaction data mechanism is out of scope for this specification.
```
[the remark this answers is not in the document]
Sakurann: umm i actually think this one is ok as a top level sentence, not part of the definition?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -312,14 +323,15 @@ GET /authorize?
...
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` parameter): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` and `transaction_data` parameters): 
 
 ```
 GET /authorize?
   client_id=x509_san_dns:client.example.org
   &client_metadata=...
   &request_uri=https%3A%2F%2Fclient.example.org%2Frequest%2Fvapof4ql2i7m41m68uep
   &request_uri_method=post HTTP/1.1
+  &transaction_data=...
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
```
true.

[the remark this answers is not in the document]
Sakurann:
```suggestion
```
true.
but is this example even correct? i mean, where in the spec do we define that client_metadata and client_id_scheme parameters can be sent alongside request_uri...?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -312,14 +323,15 @@ GET /authorize?
...
 
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` parameter): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` and `transaction_data` parameters): 
 
 ```
 GET /authorize?
   client_id=x509_san_dns:client.example.org
   &client_metadata=...
   &request_uri=https%3A%2F%2Fclient.example.org%2Frequest%2Fvapof4ql2i7m41m68uep
   &request_uri_method=post HTTP/1.1
```
Sakurann, line 333:
```suggestion
  &request_uri_method=post
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -312,14 +323,15 @@ GET /authorize?
   &nonce=n-0S6_WzA2Mj HTTP/1.1
 ```
 
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` parameter): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` and `transaction_data` parameters):
```
Sakurann, line 326:
```suggestion
The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata`): 
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
Sakurann:
```suggestion
    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.JOSE] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
Sakurann:
```suggestion
    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (i.e., the `id` field in the Input Descriptor in Presentation Exchange or the `id` field in the Credential Query in Verifiable Presentation Query Language defined in (#vp_query)). When there are more than one element in the array, the Wallet MUST use only one of those Credentials.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -312,14 +322,14 @@ GET /authorize?
...
 
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` parameter): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata`): 
 
 ```
 GET /authorize?
   client_id=x509_san_dns:client.example.org
   &client_metadata=...
   &request_uri=https%3A%2F%2Fclient.example.org%2Frequest%2Fvapof4ql2i7m41m68uep
-  &request_uri_method=post HTTP/1.1
```
c2bo, line 322: Why remove the `HTTP/1.1`?

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
jogu: `client_id_scheme` isn't there. I agree we should remove `client_metadata` too.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,10 +310,6 @@ The following additional considerations are given for pre-existing Authorization
 `client_id`:
 : REQUIRED. Defined in [@!RFC6749]. This specification defines additional requirements to enable the use of Client Identifier Schemes as described in (#client_metadata_management).
 
-Additional request parameters MAY be defined and used,
-as described in [@!RFC6749].
-The Wallet MUST ignore any unrecognized parameters.
-
```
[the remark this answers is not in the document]
jogu: It got moved up to around line 250 as it made more sense to have it near the explanation of how to handle `transaction_data` when you don't support it.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,17 @@ This specification defines the following new request parameters:
...
 
 If the Verifier set the `request_uri_method` parameter value to `post` and there is no other means to convey its capabilities to the Wallet, it SHOULD add the `client_metadata` parameter to the Authorization Request. 
-This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. 
+
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request.
```
c2bo, line 285: I don't think the newline here was intentional? These parts seem to belong together
```suggestion
This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request.
```

jogu (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,17 @@ This specification defines the following new request parameters:
...
+    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (i.e., the `id` field in the Input Descriptor in Presentation Exchange or the `id` field in the Credential Query in Verifiable Presentation Query Language defined in (#vp_query)). When there are more than one element in the array, the Wallet MUST use only one of those Credentials.
+    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.JOSE] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
```
selfissued, line 292:
```suggestion
    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
```

stefan-kauhaus:
At the EU Digital Wallet Consortium (EWC) we are about to start piloting a use case that includes using the wallet to confirm payment details (Dynamic Linking requirement from PSD2). Therefore we would be happy and explicitly support transaction_data to go into the next Implementer's Draft so that we can implement and test it and report our findings back to the community.

openid-4-verifiable-presentations-1_0.md
Sakurann: is it possible to add it to your PR, Mike? #274

openid-4-verifiable-presentations-1_0.md
Sakurann: i think it was intentional, but i think it works either way.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -312,14 +322,14 @@ GET /authorize?
...
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` parameter): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata`): 
 
 ```
 GET /authorize?
   client_id=x509_san_dns:client.example.org
   &client_metadata=...
   &request_uri=https%3A%2F%2Fclient.example.org%2Frequest%2Fvapof4ql2i7m41m68uep
-  &request_uri_method=post HTTP/1.1
+  &request_uri_method=post
```
Sakurann, line 332:
```suggestion
  &request_uri_method=post HTTP/1.1
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -809,7 +821,18 @@ The following is a non-normative example of the payload of the JWT used in the e
...
+
+The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `transaction_data_hashes` parameter defined below. If the wallet does not support `transaction_data` parameter, it MUST return an error.
+
+Where to include`transaction_data_hashes` parameter in the response is specific to each credential format and is defined in a respective Credential Format Profile.
+
+* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format Profile, but it has to be included in the  mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
+* `transaction_data_hashes_alg`: REQUIRED when this parameter was present in the `transaction_data` request parameter. String representing a hash algorithm identifier used to calculate hashes in `transaction_data_hashes` response parameter.
+
+## Error Response {#error_response}
```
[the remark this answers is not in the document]
Sakurann: this helps with URL to this section being better formed

openid-4-verifiable-presentations-1_0.md
```diff
@@ -251,6 +251,10 @@ Depending on the Client Identifier Scheme, the Verifier can communicate a JSON o
 
 This specification enables the Verifier to send both Presentation Definition JSON object and Client Metadata JSON object by value or by reference.
 
+Additional request parameters, other than those defined in this section, MAY be defined and used, as described in [@!RFC6749].
+The Wallet MUST ignore any unrecognized parameters, other than the `transaction_data` parameter.
+The wallets that do not support the `transaction_data` parameter MUST reject requests that contain it.
```
paulbastian, line 256: why is this in this paragraph and not in the new parameters section?

Sakurann: because this is about extensibility. the previous sentence says ignore if you don't recognize a parameter, and the next sentence says, but for one specific `transaction_data` parameter, do not ignore (ie the rule in the previous sentence does not apply)

Sakurann, line 256:
```suggestion
The Wallet MUST ignore any unrecognized parameters, other than the `transaction_data` parameter.
One exception to this rule is `transaction_data` parameter, and the wallets that do not support this parameter MUST reject requests that contain it.
```

Sakurann: i can clarify this

danielfett (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,16 @@ This specification defines the following new request parameters:
...
+
+    * `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
+    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (i.e., the `id` field in the Input Descriptor in Presentation Exchange or the `id` field in the Credential Query in Verifiable Presentation Query Language). When there are more than one element in the array, the Wallet MUST use only one of those Credentials.
```
paulbastian, line 290:
```suggestion
   * `credential_ids`: REQUIRED. Array of strings each referencing a Credential requested by the Verifier that shall be used to authorize the transaction. In Presentation Exchange the string matches the `id` field in the Input Descriptor. In Verifiable Presentation Query Language (defined in (#vp_query)) the string matches the `id` field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of those Credentials.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,16 @@ This specification defines the following new request parameters:
...
+    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
+
+Each document defining a transaction data type specifies whether that type requires a Credential that is bound to transaction data to be specifically issued for that purpose. How the Credential Issuer expresses that a particular Credential can be used for obtaining user's authorization using transaction data mechanism is out of scope for this specification.
```
paulbastian, line 293:
```suggestion
Each document specifying a transaction data type must define the Credential that is being used to authorize the transaction data. The Credential may be a general purpose Credential or issued specifically for the authorization use case. A mechanism for Credential Issuers to express that a particular Credential can be used for authorization of transaction data is out of scope for this specification.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -854,11 +874,19 @@ This document also defines the following additional error codes and error descri
 
 - The value of the `request_uri_method` request parameter is neither `get` nor `post` (case-sensitive).
 
+`invalid_transaction_data`:
+
+- any of the following is true for at least one object in the `transaction_data` structure:
+  - contains an unknown transaction data type value,
```
paulbastian, line 880:
```suggestion
  - transaction data type is not supported by the wallet
  - contains an unknown transaction data type value,
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,16 @@ This specification defines the following new request parameters:
...
+    * `credential_ids`: REQUIRED. Array of strings each pointing to an identifier that identifies a set of information describing a Credential that the Verifier is requesting transaction data in a particular object to be bound to (i.e., the `id` field in the Input Descriptor in Presentation Exchange or the `id` field in the Credential Query in Verifiable Presentation Query Language). When there are more than one element in the array, the Wallet MUST use only one of those Credentials.
+    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
+
```
paulbastian, line 292: Could we add an Authorization Request example with some dummy transcation data?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -854,11 +874,19 @@ This document also defines the following additional error codes and error descri
...
 - The value of the `request_uri_method` request parameter is neither `get` nor `post` (case-sensitive).
 
+`invalid_transaction_data`:
+
+- any of the following is true for at least one object in the `transaction_data` structure:
+  - contains an unknown transaction data type value,
+  - is an object of known type but containing unknown fields,
+  - contains fields of the wrong type for the transaction data type,
+  - contains fields with invalid values for the transaction data type, or
+  - is missing required fields for the transaction data type.
```
paulbastian, line 884:
```suggestion
  - is missing required fields for the transaction data type.
  - the credential_ids does not match
  - the referenced Credential is not available in the Wallet
```

paulbastian (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,16 @@ This specification defines the following new request parameters:
...
-This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. 
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request.
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a base64url encoded JSON object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST return an error if a request contains even one unrecognized transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
+
+    * `type`: REQUIRED. String that is the Identifier of the transaction data type and determines the allowable contents of the object that contains it. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
    * `type`: REQUIRED. String that identifies the type of transaction data . This value determines parameters that can be included in the `transaction_data` object. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for `type` values.
```

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
    * `credential_ids`: REQUIRED. Array of strings each referencing a Credential requested by the Verifier that shall be used to authorize this transaction. In Presentation Exchange, the string matches the `id` field in the Input Descriptor. In Verifiable Presentation Query Language (defined in (#vp_query)), the string matches the `id` field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
```

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
Each document specifying details of a transaction data type defines what Credential(s) can be used to authorize those transactions. Those Credential(s) can be issued specifically for the transaction authorization use case or re-use existing Credential(s) used for user identification. A mechanism for Credential Issuers to express that a particular Credential can be used for authorization of transaction data is out of scope for this specification.
```

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
  - contains an unknown or unsupported transaction data type value,
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -854,11 +874,21 @@ This document also defines the following additional error codes and error descri
...
+`invalid_transaction_data`:
+
+- any of the following is true for at least one object in the `transaction_data` structure:
+  - contains an unknown or unsupported transaction data type value,
+  - is an object of known type but containing unknown fields,
+  - contains fields of the wrong type for the transaction data type,
+  - contains fields with invalid values for the transaction data type, or
+  - is missing required fields for the transaction data type.
+  - the credential_ids does not match
+  - the referenced Credential is not available in the Wallet
```
Sakurann, line 886:
```suggestion
  - the referenced Credential(s) are not available in the Wallet
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,26 @@ This specification defines the following new request parameters:
...
+Each document specifying details of a transaction data type defines what Credential(s) can be used to authorize those transactions. Those Credential(s) can be issued specifically for the transaction authorization use case or re-use existing Credential(s) used for user identification. A mechanism for Credential Issuers to express that a particular Credential can be used for authorization of transaction data is out of scope for this specification.
+
+The following is a non-normative example of a transaction data content, after base64url decoding one of the strings in the `transaction_data` parameter:
+
+```
+{
+  "type": "example_type",
+  "credential_ids": [ "id card credential" ],
+  "transaction_data_hashes_alg": [ "sha-256" ]
+}
```
paulbastian, line 302:
```suggestion
  "transaction_data_hashes_alg": [ "sha-256" ],
  "example_type_specific_parameter": "some-values"
}
```

paulbastian (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -277,7 +281,26 @@ This specification defines the following new request parameters:
...
 
 If the Verifier set the `request_uri_method` parameter value to `post` and there is no other means to convey its capabilities to the Wallet, it SHOULD add the `client_metadata` parameter to the Authorization Request. 
-This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request. 
+This enables the Wallet to assess the Verifier's capabilities, allowing it to transmit only the relevant capabilities through the `wallet_metadata` parameter in the Request URI POST request.
+
+`transaction_data`: 
+: OPTIONAL. Array of strings, where each string is a base64url encoded JSON object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST return an error if a request contains even one unrecognized transaction data type or transaction data not conforming to the respective type definition. Each object consists of the following parameters:
```
c2bo, line 287: I am wondering if we should state a bit more explicitly here that there might be more parameters in that object depending on the transaction data type?

c2bo (approved):
Not sure about that one sentence, but I think this is good to go as a starting point for transaction data - other formats like mDoc will then follow in another PR?

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
: OPTIONAL. Array of strings, where each string is a base64url encoded JSON object that contains a typed parameter set with details about the transaction that the Verifier is requesting the End-User to authorize. See (#transaction_data) for details. The Wallet MUST return an error if a request contains even one unrecognized transaction data type or transaction data not conforming to the respective type definition. In addition to the parameters determined by the type of transaction data, each `transaction_data` object consists of the following parameters defined by this specification:
```

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
  "transaction_data_hashes_alg": [ "sha-256" ]
  // other transaction data type specific parameters
}
```

c2bo:
Looks good to me with the changes! I guess we should not merge this until the Query Language PR is merged so it doesn't break the github action (vp_query reference)?

jogu:
4 approvals, discussed in multiple WG calls, open for over 4 months, all comments addressed - merging! Thank you everyone!

jogu:
> 4 approvals, discussed in multiple WG calls, open for over 4 months, all comments addressed - merging! Thank you everyone!

Well, merging after making a trivial tweak so the spec compiles (removing the #vp_query reference) - we should probably add back the two VP query section links once query language is merged.

PR #200, opened 2024-06-25
fix: add wallet_unavailable error code; fixes #191

### LINKED ISSUE #191: Error code if the wallet is not installed
We have a use case where a wallet backend handles HTTPS Deep Links (e.g., https://wallet.example.com) if the mobile platform does not invoke the user's wallet on the mobile device.

Expected flow:
- The user visits the RP in their mobile browser.
- The user taps a button containing a HTTPS Deep Link on the RP website.
- The mobile browser checks with the mobile platform if the intent can be handled by an app.
- In this case, it cannot be handled because the wallet is not installed.
- The mobile browser redirects to the wallet backend (e.g., https://wallet.example.com). Note, this is standard behaviour on mobile platforms.
- The wallet backend generates an error and redirects the user back with a `wallet_unavailable` error.
- The RP can now respond to the `wallet_unavailable` error with an alternative flow that does not require a wallet.

To improve UX, the wallet backend could redirect the user back to the RP with an error code, but we have not defined one for this case. I couldn't find a suitable error code in OAuth either. Defining a custom error code is not an option since it would be wallet vendor specific and this information should not be disclosed to the RP.

My suggestion is to define a new error code, `wallet_unavailable`, that MAY be returned in such cases. Note that, as with many OAuth implementations, it is at the AS's discretion to return an error code at all.

awoie:
This PR adds an optional error code `wallet_unavailable`; fixes #191.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -793,7 +793,7 @@ This document also defines the following additional error codes and error descri
 
 `wallet_unavailable`:
 
-- The Wallet could not respond to the request.
+- The Wallet appears to be unavailable and therefore unable to respond to the request.
 
 
 ## VP Token Validation
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -793,7 +793,7 @@ This document also defines the following additional error codes and error descri
 
 `wallet_unavailable`:
 
-- The Wallet appears to be unavailable and therefore unable to respond to the request. This can be useful in situations where the user agent cannot invoke the Wallet and another component receives the request while the user still controls the user experience and wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
+- The Wallet appears to be unavailable and therefore unable to respond to the request. This can be useful in situations where the user agent cannot invoke the Wallet and another component receives the request while and the user wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
 
 
 ## VP Token Validation
@@ -1211,7 +1211,7 @@ Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount 
 
 ## Authorization Error Response with Wallet unavailable error code
 
-In the event that another component is invoked instead of the Wallet while the user still controls the user experience, the user MUST be informed and give consent before returning the `wallet_unavailable` Authorization Error Response to the Verifier.
+In the event that another component is invoked instead of the Wallet, the user MUST be informed and give consent before returning the `wallet_unavailable` Authorization Error Response to the Verifier.
 
 {backmatter}
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -793,7 +793,7 @@ This document also defines the following additional error codes and error descri
 
 `wallet_unavailable`:
 
-- The Wallet appears to be unavailable and therefore unable to respond to the request. This can be useful in situations where the user agent cannot invoke the Wallet and another component receives the request while and the user wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
+- The Wallet appears to be unavailable and therefore unable to respond to the request. It can be useful in situations where the User Agent cannot invoke the Wallet and another component receives the request while the End-User wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
 
 
 ## VP Token Validation
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -793,7 +793,7 @@ This document also defines the following additional error codes and error descri
 
 `wallet_unavailable`:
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -791,6 +791,10 @@ This document also defines the following additional error codes and error descri
 
 - The value of the `request_uri_method` request parameter is neither `get` nor `post` (case-sensitive).
 
+`wallet_unavailable`:
+
+- The Wallet could not respond to the request.
```
peppelinux, line 796:
```suggestion
- The Wallet appears as currently unavailable and therefore unable to process the request.
```

jogu (changes requested):
I think we probably need some more explanation in the specification text; it's not intuitive how this code is used and I think we need to mention the privacy/security concerns (as discussed on the issue) that the error code mustn't be returned without consent from the user.

openid-4-verifiable-presentations-1_0.md
jogu: The grammar in Guiseppe's suggestion is a little odd; this would be slightly better I think:

```suggestion
- The Wallet appears to be unavailable and therefore unable to respond to the request.
```

awoie:
> I think we probably need some more explanation in the specification text; it's not intuitive how this code is used and I think we need to mention the privacy/security concerns (as discussed on the issue) that the error code mustn't be returned without consent from the user.

That sounds good to me. I'll update the PR.

awoie:
@peppelinux @jogu I updated the PR to incorporate your requested changes. Please review again.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -791,6 +791,10 @@ This document also defines the following additional error codes and error descri
 
 - The value of the `request_uri_method` request parameter is neither `get` nor `post` (case-sensitive).
 
+`wallet_unavailable`:
+
+- The Wallet appears to be unavailable and therefore unable to respond to the request. This can be useful in situations where the user agent cannot invoke the Wallet and another component receives the request while the user still controls the user experience and wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
```
peppelinux, line 796: I would use the term user-agent to clarify that's a specialized entity

jogu: I'm not sure I'm following, could you expand on your suggestion please Guiseppe? Maybe suggest alternative text?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1205,6 +1209,10 @@ Mandatory user interaction before sending the request, like clicking a button, u
 
 Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount of information possible, and in particular, without any HTTP headers identifying the software used for the request (e.g., HTTP libraries or their versions). The Wallet MUST NOT send PII or any other data that could be used for fingerprinting to the Request URI in order to prevent user tracking. 
 
+## Authorization Error Response with Wallet unavailable error code
+
+In the event that another component is invoked instead of the Wallet while the user still controls the user experience, the user MUST be informed and give consent before returning the `wallet_unavailable` Authorization Error Response to the Verifier.
```
jogu, line 1214: I'm not sure I follow the 'while the user still controls the user experience' part. What is an example of where the user isn't in control?

awoie: I think I can remove that by-sentence because we are already saying the user MUST be informed. I updated the PR.

jogu (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -791,6 +791,10 @@ This document also defines the following additional error codes and error descri
 
 - The value of the `request_uri_method` request parameter is neither `get` nor `post` (case-sensitive).
 
+`wallet_unavailable`:
+
+- The Wallet appears to be unavailable and therefore unable to respond to the request. This can be useful in situations where the user agent cannot invoke the Wallet and another component receives the request while and the user wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
```
Sakurann, line 796:
```suggestion
- The Wallet appears to be unavailable and therefore unable to respond to the request. It can be useful in situations where the User Agent cannot invoke the Wallet and another component receives the request while the End-User wishes to continue the journey on the Verifier website. For example, this applies when using claimed HTTPS URIs handled by the Wallet provider in case the platform cannot or does not translate the URI into a platform intent to invoke the Wallet.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1205,6 +1209,10 @@ Mandatory user interaction before sending the request, like clicking a button, u
 
 Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount of information possible, and in particular, without any HTTP headers identifying the software used for the request (e.g., HTTP libraries or their versions). The Wallet MUST NOT send PII or any other data that could be used for fingerprinting to the Request URI in order to prevent user tracking. 
 
+## Authorization Error Response with Wallet unavailable error code
+
+In the event that another component is invoked instead of the Wallet, the user MUST be informed and give consent before returning the `wallet_unavailable` Authorization Error Response to the Verifier.
```
Sakurann, line 1214: so we are relying on this `another component that is invoked` to return this error, right?

openid-4-verifiable-presentations-1_0.md
Sakurann, line 796: If my understanding is correct, I would add something to clarify that it is not the wallet, but the wrong component invoked instead of a correct wallet that return this error?

Sakurann (commented):
please add document history entry :)

openid-4-verifiable-presentations-1_0.md
awoie: Yes, that is correct. I can add this.

openid-4-verifiable-presentations-1_0.md
awoie: Yes

openid-4-verifiable-presentations-1_0.md
awoie: @Sakurann I updated the PR.

Sakurann (approved):

ssanchocanela (approved):

jogu:
Now has 3 reviews (many thanks @ssanchocanela!), all outstanding comments resolved and has been open for over a week - merging!

PR #217, opened 2024-07-17
gh-25: clarify applicability of the redirect_uri protection technique

### LINKED ISSUE #25: direct_post response mode, response with a redirect_uri, and protection against session fixation
### Original Reporter: pedro-felix

In the context of the OpenID4VP draft 18, I’ve have some questions regarding the `direct_post` response mode, responding with a `redirect_uri`, and protection against session fixation.

1. The use of a payload containing a `redirect_uri` when responding to the POST with the Authorization Response, namely the algorithm described in [https://openid.net/specs/openid-4-verifiable-presentations-1\_0.html#section-10.5](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-10.5), is only effective when on a _same-device_ scenario, right? If so, the spec should make that information more clearer.  
  Even on a same-device scenario, the algorithm on section 10.5 only works if the wallet opens the same user-agent/browser as the one used initially by the user to interact with the verifier, right? If so, that can be an issue when the user has multiple browsers installed.
2. How can the `redirect_uri` also be usable on cross-device scenarios to protect against session fixation?

    1. Have the UI shown by the resource at the `redirect_uri` present a code that needs to be presented directly at the verifier \(on the initial session between the user and the verifier\)?
    2. Are there any more guidelines on how to protect this flow?
    

‌

pmhsfelix:
This PR clarifies the applicability of the `redirect_uri` protection technique against session fixation attacks.

closes #25

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1140,7 +1140,10 @@ Such an attack is impossible against flows implemented with the Response Mode `f
 
 However, the Response Mode `direct_post` is susceptible to such an attack as the result is sent from the Wallet out-of-band to the Verifier's Response Endpoint. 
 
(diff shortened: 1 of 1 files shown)


jogu (approved):
lgtm - this fully address the issue right? If so can you add `closes #25` to the description in the first comment please so we don't have to remember to manually close it :)

bc-pi (approved):

c2bo (approved):

Sakurann (approved):

pmhsfelix:
@Sakurann @jogu I believe this one is also good to be merged.

jogu:
Indeed - 3+ approvals, open for 7 days, no unresolved comments - merging!

Issue #211, opened 2024-07-11
co-existence of multiple query languages (to PE, or not to PE)


Sakurann:
I think we should have an idea how the co-existence of or transition between two query languages in OpenID4Vp will happen: PE v2 and a new one.

Is one or the other mandatory? for whom?
How does the wallet know which query language it is receiving? does this mean we should define a different high level parameter that is not `presentation_definition` for a new query language?

Sakurann:
discussed in a WG. 
- there seems to be an agreement to use a different top -level parameter for a new query language. @jogu proposed `vp_query`, but we should bikeshed. 
    - there was a proposal to have a top-level parameter under which query language specific parameters sit (would break existing PE implementations and we hopefully/probably won't have more than 2 query languages...) - leaning towards keeping it simple and not defining such an extension.

bc-pi:
`presentation_definition` is a... um... silly name at best so no need to keep that

selfissued:
I agree that the new query language should use its own parameter names.

David-Chadwick:
I prefer a top level parameter called something like "query-language" and then beneath this there will be the query languages that we support - currently only two, but who knows what else in future? Remember that until a few months ago we only mandated one query language, now we support two. This proposed structure is future proof so that if we want to add a third in future the structure of the current two will not change.

decentralgabe:
allowing both means that everyone needs to implement both. this increasing the change for interop gaps across implementations. 

requiring one but not another also harms interoperability. just require one.

TimoGlastra:
During the transition period it might be beneficial to allow for both the presentation definition and dcql query to be present in an authorization request. There's no easy way for the verifier to know if the wallet already supports DCQL, and our initial design was that we include both for now, but this doesn't seem allowed:
> Exactly one of the following parameters MUST be present in the Authorization Request: dcql_query, presentation_definition, presentation_definition_uri, or a scope value representing a Presentation Definition.

Then the wallet can choose, and we can avoid this problem of knowing which format to use before having interacted with a wallet. Is there a reason for the spec to be this strict? Does it hinder wallet implementations if both can be present?

The logic could be:
- Either `dcql_query` and/or `presentation_definition_uri` / `presentation_definition` needs to be present.
- Wallet chooses which format to use
- If response contains a `presentation_submission` you'll know the wallet used the presentation definition.

Sakurann:
WG discussion:
[...]

PR #220, opened 2024-07-25
Introduce new query language (Approach 1)

### LINKED ISSUE #178: Proposal for new query language
Related to #160, #161, #162 and #157.

This is a proposal for a new query language fulfilling the requirements discussed in the issues linked above. I can create a PR once we are sure that this is the direction we want to go.

Note: This proposal for now does not cover restricting/querying values of claims. See [examples 6 and 7 here](https://hackmd.io/1siVhjzOTWOE9ppdF_t57A?both#/) for a proposal, but that is out of scope for this issue.

## Example 1a - Request one Credential (SD-JWT VC)

Just one credential is requested. It must contain the claims `last_name`, `first_name`, and `address/street_address`.

```yaml
{
  "credentials": {
    "my_cred_1": {
      "format": "vc+sd-jwt",
      "vc+sd-jwt": {  # format-specific requirements
          "vct": "https://credentials.example.com/identity_credential",
          "alg_values": ["ES256", "ES384"],
          "hash_alg_values": ["SHA-256"]
      },
      "claims":[
        { "path": ["last_name"] },
        {
          "path": ["first_name"],
          "intent_to_retain": false  # might be defined as an add-on, not in this PR
        },
        {  "path": ["address", "street_address"]  }
      ]
    }
  }
}
```

Note the identifier `my_cred_1`. It will show up in the VP Token, which is restructured to be an object mapping the identifiers in the request to the presentation (string or object depending on a credential format):

```yaml
vp_token={
    "my_cred_1": "eY..."
}
```

## Example 1b - Request one Credential (mdoc)

The syntax for the claims is slightly different due to the structure of mdocs:

```yaml
{
  "credentials": {
  {
     "my_mdoc_credential": {
        "format": "mso_mdoc",
        "mso_mdoc": {  # format-specific requirements
            "doctype": "org.iso.7367.1.mVR",
            "alg_values": [ "EdDSA" ],
            "hash_algorithm_values": [ "SHA-384" ],
        },
        "claims": [
          {
            "namespace": "org.iso.7367.1",
            "claim_name": "vehicle_holder",
            "intent_to_retain": false
          },
          {
            "namespace": "org.iso.18013.5.1",
            "claim_name": "first_name",
            "intent_to_retain": false
          }      
        ]
      }
    }
  }
}
```

Response:

```yaml
vp_token={
    "my_mdoc_credential": "<mdoc string>"
}
```

## Example 2 - Request multiple Credentials (all of the requested credentials needs to be returned)

All three credentials have to be delivered:

```yaml
{
  "credentials": {
    "my_cred_1": {
      "format": "vc+sd-jwt",
      "vc+sd-jwt": {
          "vct": "https://credentials.example.com/identity_credential",
          "alg_values": ["ES256", "ES384"],
          "hash_alg_values": ["SHA-256"]
      },
      "purpose": "give me the credential",  # we might pull this to the top
      "claims":[
        { "path": ["last_name"] }
      ]
    },
    "my_cred_2": {
      "format": "vc+sd-jwt",
      "vc+sd-jwt": {
          "vct": "https://credentials.example.com/identity_credential/2",
          "alg_values": ["ES256", "ES384"],
          "hash_alg_values": ["SHA-256"]
      },
      "claims": [
          { "path": ["addresses"] }
      ]
    },
    "my_cred_3": {
      "format": "vc+sd-jwt",
      "vc+sd-jwt": {
          "vct": "https://credentials.example.com/identity_credential/3",
          "alg_values": ["ES256", "ES384"],
          "hash_alg_values": ["SHA-256"]
      },
      "claims": [
        { "path": ["addresses"] }
      ]
    }
  }
}
```

All three credentials will show up in the VP Token:

```yaml
vp_token={
    "my_cred_1": "eY...",
    "my_cred_2": "eY...",
    "my_cred_3": "eY..."
}
```

Note: If there is a requirement on sending multiple credentials fulfilling one request, this should be turned into an array.

## Example 3 - Request alternative claims (one rule)

Require `age_over_18` or `birth_date` or `age_birth_year` by grouping them in an object with the properties `require` and `from`. The order of the claims in the `from` list implies a preference by the verifier.

```yaml
{
  "credentials": {
    "my_cred_1": {
      "format": "vc+sd-jwt",
      "vc+sd-jwt": {
          "vct": "https://credentials.example.com/identity_credential",
          "alg_values": ["ES256", "ES384"],
          "hash_alg_values": ["SHA-256"]
      },
      "purpose": "give me the credential",
      "claims":[
        { "path": ["last_name"] },
        {
          "required": 1,
          "from": [
            { "path": ["age_over_18"] },
            { "path": ["birth_date"]  },
            { "path": ["age_birth_year"] }
          ]
        },
      ]
    }
  }
}

```

Note: the expectation is for the wallet to check which of the alternative claims are present and render a user screen based on that - not for the wallet to display to the user all alternative claims (`age_over_18` or `birth_date` or `age_birth_year` in this case).

## Example 4 - Request alternative claims (multiple rules)

By nesting groups, complex queries can be expressed; here, either "zip_code" or both "city" and "state" are required.

```yaml
{
  "credentials": {
    "my_cred_1": {
      "format": "vc+sd-jwt",
      "vc+sd-jwt": {
          "vct": "https://credentials.example.com/identity_credential",
          "alg_values": ["ES256", "ES384"],
          "hash_alg_values": ["SHA-256"]
      },
      "purpose": "give me the credential",
      "claims":[
        { "path": ["last_name"] }, 
        { 
          "required": 1, 
          "from": [ 
            { "path": [ "zip_code" ] }, 
            {
              "required": 2,
              "from": [
                { "path": [ "city" ] },
                { "path": [ "state" ] }
              ]
            }
          ] 
        } 
      ]
    }
  }
}
```

## Example 5 - Credentials Alternatives

The same logic to requesting alternative claims can be applied to requesting alternative credentials as well. The "rules" for credentials are separated from the credentials to make the processing easier and the syntax more clean. The following example requests "mycred_1" and either "my_cred_2" or "my_cred_3":

```yaml
{
  "rules": {
    "required": 2,
    "from": [
      { "ref": "my_cred_1" },
      {
        "required": 1,
        "from": [
          { "ref": "my_cred_2" },
          { "ref": "my_cred_3" }
        ]
      }
    ]
  },
  "credentials": { # each object follows the syntax described above
    "my_cred_1": {...}, 
    "my_cred_2": {...},
    "my_cred_3": {...}
  }
}
```

As shown in the examples further up, if `rules` is omitted, all credentials have to be delivered.

If the verifier just wants the user's claim (like street address/age), it would express it as "give me one of these credentials and outline any number of credentials whose schema includes a desired claim" - the verifier still needs to specify the format/type of the credential to indicate what it can understand.

Only the available/selected credential will show up in the VP Token:

```yaml
vp_token={
    "my_cred_1": "eY..."
    "my_cred_2": "eY..."
}
```

danielfett:
Fixes #178 

Todo:

 - [x] Determine which "extensions" we want to specify ("intent_to_retain" etc.)
 - [ ] Determine which format-specific  parameters we need
 - [ ] Define interaction with existing query lang (see #211)
 - [ ] Define new format for the VP Token

### DIFF
--- a/examples/query_lang/credentials_alternatives.json
+++ b/examples/query_lang/credentials_alternatives.json
@@ -17,8 +17,8 @@
       "format": "vc+sd-jwt",
       "vct_values": ["https://credentials.example.com/identity_credential"],
       "claims": [
-        { "path": ["last_name"] },
-        { "path": ["first_name"] },
+        { "path": ["given_name"] },
+        { "path": ["family_name"] },
         { "path": ["address", "street_address"] }
       ]
     },
@@ -26,17 +26,17 @@
       "format": "vc+sd-jwt",
       "vct_values": ["https://credentials.example.com/identity_credential"],
       "claims": [
-        { "path": ["last_name"] },
-        { "path": ["first_name"] }
+        { "path": ["family_name"] },
+        { "path": ["given_name"] }
       ]
     },
     "residence_info": {
       "format": "vc+sd-jwt",
       "vct_values": ["https://credentials.example.com/residence_credential"],
       "claims": [
-        { "path": ["zip_code"] },
-        { "path": ["city"] },
-        { "path": ["state"] }
+        { "path": ["postal_code"] },
+        { "path": ["locality"] },
+        { "path": ["region"] }
       ]
     }
   }
--- a/examples/query_lang/credentials_alternatives.json
+++ b/examples/query_lang/credentials_alternatives.json
@@ -1,0 +1,43 @@
+{
+  "rules": {
+    "required": 2,
+    "from": [
+      { "ref": "pid" },
+      {
+        "required": 1,
+        "from": [
+          { "ref": "pid_reduced" },
+          { "ref": "residence_info" }
+        ]
+      }
+    ]
+  },
+  "credentials": {
+    "pid": {
+      "format": "vc+sd-jwt",
+      "vct_values": ["https://credentials.example.com/identity_credential"],
+      "claims": [
+        { "path": ["last_name"] },
+        { "path": ["first_name"] },
+        { "path": ["address", "street_address"] }
+      ]
+    },
+    "pid_reduced": {
+      "format": "vc+sd-jwt",
+      "vct_values": ["https://credentials.example.com/identity_credential"],
+      "claims": [
+        { "path": ["last_name"] },
+        { "path": ["first_name"] }
+      ]
+    },
+    "residence_info": {
+      "format": "vc+sd-jwt",
+      "vct_values": ["https://credentials.example.com/residence_credential"],
+      "claims": [
+        { "path": ["zip_code"] },
+        { "path": ["city"] },
+        { "path": ["state"] }
+      ]
+    }
+  }
+}
--- a/examples/query_lang/multi_credentials.json
+++ b/examples/query_lang/multi_credentials.json
@@ -11,7 +11,7 @@
             }],
             "claims": [
                 { "path": [ "last_name" ] },
-                { "path": [ "first_name" ] },
+                { "path": [ "given_name" ] },
                 { "path": [ "address", "street_address" ] }
             ]
         },
@@ -21,8 +21,8 @@
                 "vct_values": [ "https://example.com/education_credential" ]
             },
             "claims": [
-                { "path": [ "last_name" ] },
-                { "path": [ "first_name" ] },
+                { "path": [ "family_name" ] },
+                { "path": [ "given_name" ] },
                 { "path": [ "degree" ] }
             ]
         },
@@ -32,10 +32,10 @@
                 "vct_values": [ "https://example.com/residence_proof" ]
             },
             "claims": [
-                { "path": [ "last_name" ] },
-                { "path": [ "first_name" ] },
-                { "path": [ "city" ] },
-                { "path": [ "state" ] }
+                { "path": [ "family_name" ] },
+                { "path": [ "given_name" ] },
+                { "path": [ "locality" ] },
+                { "path": [ "region" ] }
             ]
         }
     }
--- a/examples/query_lang/multi_credentials.json
+++ b/examples/query_lang/multi_credentials.json
@@ -1,0 +1,42 @@
+{
+    "credentials": {
+        "pid": {
+            "format": "vc+sd-jwt",
+            "vc+sd-jwt": {
+                "vct_values": [ "https://example.com/personal_identity" ]
+            },
+            "purpose": [{
+                "lang": "en-US",
+                "text": "To prove the identity of the holder"
+            }],
+            "claims": [
+                { "path": [ "last_name" ] },
+                { "path": [ "first_name" ] },
+                { "path": [ "address", "street_address" ] }
+            ]
+        },
+        "degree": {
+            "format": "vc+sd-jwt",
+            "vc+sd-jwt": {
+                "vct_values": [ "https://example.com/education_credential" ]
+            },
+            "claims": [
+                { "path": [ "last_name" ] },
+                { "path": [ "first_name" ] },
+                { "path": [ "degree" ] }
+            ]
+        },
+        "residence": {
+            "format": "vc+sd-jwt",
+            "vc+sd-jwt": {
+                "vct_values": [ "https://example.com/residence_proof" ]
+            },
+            "claims": [
+                { "path": [ "last_name" ] },
+                { "path": [ "first_name" ] },
+                { "path": [ "city" ] },
+                { "path": [ "state" ] }
+            ]
+        }
+    }
+}
--- a/examples/query_lang/multi_credentials.json
+++ b/examples/query_lang/multi_credentials.json
@@ -5,10 +5,6 @@
             "vc+sd-jwt": {
                 "vct_values": [ "https://example.com/personal_identity" ]
             },
-            "purpose": [{
-                "lang": "en-US",
-                "text": "To prove the identity of the holder"
-            }],
             "claims": [
                 { "path": [ "last_name" ] },
                 { "path": [ "given_name" ] },
--- a/examples/query_lang/nested_claim_queries.json
+++ b/examples/query_lang/nested_claim_queries.json
@@ -25,12 +25,12 @@
         {
           "required": 1,
           "from": [
-            { "path": [ "zip_code" ] },
+            { "path": [ "postal_code" ] },
             {
               "required": 2,
               "from": [
-                { "path": [ "city" ] },
-                { "path": [ "state" ] }
+                { "path": [ "locality" ] },
+                { "path": [ "region" ] }
               ]
             }
           ]
--- a/examples/query_lang/nested_claim_queries.json
+++ b/examples/query_lang/nested_claim_queries.json
@@ -7,19 +7,6 @@
         "alg_values": [ "ES256", "ES384" ],
         "sd_alg_values": [ "SHA-256" ]
       },
-      "purpose": [{
-          "lang": "en-US",
-          "text": "To prove the identity of the holder"
-        },
-        {
-          "lang": "no-NO",
-          "text": "For å bevise innehaverens identitet"
-        },
-        {
-          "lang": "jp-JP",
-          "text": "保持者の身元を証明するため"
-        }
-      ],
       "claims": [
         { "path": [ "last_name" ] },
         {
--- a/examples/query_lang/nested_claim_queries.json
+++ b/examples/query_lang/nested_claim_queries.json
@@ -1,0 +1,41 @@
+{
+  "credentials": {
+    "my_cred_1": {
+      "format": "vc+sd-jwt",
+      "vc+sd-jwt": {
+        "vct_values": [ "https://credentials.example.com/identity_credential" ],
+        "alg_values": [ "ES256", "ES384" ],
+        "sd_alg_values": [ "SHA-256" ]
+      },
+      "purpose": [{
+          "lang": "en-US",
+          "text": "To prove the identity of the holder"
+        },
+        {
+          "lang": "no-NO",
+          "text": "For å bevise innehaverens identitet"
+        },
+        {
+          "lang": "jp-JP",
+          "text": "保持者の身元を証明するため"
+        }
+      ],
+      "claims": [
+        { "path": [ "last_name" ] },
+        {
+          "required": 1,
+          "from": [
+            { "path": [ "zip_code" ] },
+            {
+              "required": 2,
+              "from": [
+                { "path": [ "city" ] },
+                { "path": [ "state" ] }
+              ]
+            }
+          ]
+        }
+      ]
+    }
+  }
+}
--- a/examples/query_lang/simple.json
+++ b/examples/query_lang/simple.json
@@ -8,8 +8,8 @@
         "sd_alg_values": ["SHA-256"]
       },
       "claims":[
-        { "path": ["last_name"] },
-        { "path": ["first_name"] },
+        { "path": ["family_name"] },
+        { "path": ["given_name"] },
         { "path": ["address", "street_address"] }
       ]
     }
--- a/examples/query_lang/simple.json
+++ b/examples/query_lang/simple.json
@@ -1,0 +1,17 @@
+{
+  "credentials": {
+    "my_cred_1": {
+      "format": "vc+sd-jwt",
+      "vc+sd-jwt": {
+        "vct_values": [ "https://credentials.example.com/identity_credential" ],
+        "alg_values": ["ES256", "ES384"],
+        "sd_alg_values": ["SHA-256"]
+      },
+      "claims":[
+        { "path": ["last_name"] },
+        { "path": ["first_name"] },
+        { "path": ["address", "street_address"] }
+      ]
+    }
+  }
+}
--- a/examples/query_lang/simple_mdoc.json
+++ b/examples/query_lang/simple_mdoc.json
@@ -1,0 +1,22 @@
+{
+  "credentials": {
+    "my_mdoc_credential": {
+      "format": "mso_mdoc",
+      "mso_mdoc": {
+        "doctype_values": "org.iso.7367.1.mVR",
+        "alg_values": [ "EdDSA" ],
+        "hash_algorithm_values": [ "SHA-384"]
+      },
+      "claims": [
+        {
+          "namespace": "org.iso.7367.1",
+          "claim_name": "vehicle_holder"
+        },
+        {
+          "namespace": "org.iso.18013.5.1",
+          "claim_name": "first_name"
+        }
+      ]
+    }
+  }
+}
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -46,7 +46,7 @@ organization="Mattr"
 
 .# Abstract
 
-This specification defines a protocol for requesting and presenting Verifiable Credentials. 
+This specification defines a protocol for requesting and presenting Verifiable Credentials.
 
 {mainmatter}
 
@@ -54,7 +54,7 @@ This specification defines a protocol for requesting and presenting Verifiable C
 
 This specification defines a mechanism on top of OAuth 2.0 [@!RFC6749] that enables presentation of Verifiable Credentials as Verifiable Presentations. Verifiable Credentials and Verifiable Presentations can be of any format, including, but not limited to W3C Verifiable Credentials Data Model [@VC_DATA], ISO mdoc [@ISO.18013-5], IETF SD-JWT VC [@!I-D.ietf-oauth-sd-jwt-vc], and AnonCreds [@Hyperledger.Indy].
 
-OAuth 2.0 [@!RFC6749] is used as a base protocol as it provides the required rails to build a simple, secure, and developer-friendly Credential presentation layer on top of it. Moreover, implementers can, in a single interface, support Credential presentation and the issuance of Access Tokens for access to APIs based on Verifiable Credentials in the Wallet. OpenID Connect [@!OpenID.Core] deployments can also extend their implementations using this specification with the ability to transport Verifiable Presentations. 
+OAuth 2.0 [@!RFC6749] is used as a base protocol as it provides the required rails to build a simple, secure, and developer-friendly Credential presentation layer on top of it. Moreover, implementers can, in a single interface, support Credential presentation and the issuance of Access Tokens for access to APIs based on Verifiable Credentials in the Wallet. OpenID Connect [@!OpenID.Core] deployments can also extend their implementations using this specification with the ability to transport Verifiable Presentations.
 
 This specification can also be combined with [@!SIOPv2], if implementers require OpenID Connect features, such as the issuance of Self-Issued ID Tokens [@!SIOPv2].
 
@@ -98,11 +98,11 @@ Verifier:
 Issuer-Holder-Verifier Model:
 :  A model for exchanging claims, where claims are issued in the form of Verifiable Credentials independent of the process of presenting them as Verifiable Presentation to the Verifiers. An issued Verifiable Credential can (but must not necessarily) be used multiple times.
 
-Holder Binding: 
-: Ability of the Holder to prove legitimate possession of a Verifiable Credential. 
+Holder Binding:
+: Ability of the Holder to prove legitimate possession of a Verifiable Credential.
 
 Cryptographic Holder Binding:
-:  Ability of the Holder to prove legitimate possession of a Verifiable Credential by proving control over the same private key during the issuance and presentation. Mechanism might depend on the Credential Format. For example, in `jwt_vc_json` Credential Format, a Verifiable Credential with Cryptographic Holder Binding contains a public key or a reference to a public key that matches to the private key controlled by the Holder. 
+:  Ability of the Holder to prove legitimate possession of a Verifiable Credential by proving control over the same private key during the issuance and presentation. Mechanism might depend on the Credential Format. For example, in `jwt_vc_json` Credential Format, a Verifiable Credential with Cryptographic Holder Binding contains a public key or a reference to a public key that matches to the private key controlled by the Holder.
 
 Claims-based Holder Binding:
 :  Ability of the Holder to prove legitimate possession of a Verifiable Credential by proofing certain claims, e.g., name and date of birth, for example by presenting another Verifiable Credential. Claims-based Holder Binding allows long term, cross device use of a Credential as it does not depend on cryptographic key material stored on a certain device. One example of such a Verifiable Credential could be a Diploma.
@@ -116,13 +116,13 @@ VP Token:
 Wallet:
 :  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
 
-# Overview 
+# Overview
 
 This specification defines a mechanism on top of OAuth 2.0 to request and present Verifiable Credentials as Verifiable Presentations.
 
 As the primary extension, OpenID for Verifiable Presentations introduces the VP Token as a container to enable End-Users to present Verifiable Presentations to Verifiers using the Wallet. A VP Token contains one or more Verifiable Presentations in the same or different Credential formats.
 
-This specification supports any Credential format used in the Issuer-Holder-Verifier Model, including, but not limited to those defined in [@VC_DATA] (VCDM), [@ISO.18013-5] (mdoc), [@!I-D.ietf-oauth-sd-jwt-vc] (SD-JWT VC), and [@Hyperledger.Indy] (AnonCreds). Credentials of multiple formats can be presented in the same transaction. The examples given in the main part of this specification use W3C Verifiable Credentials, while examples in other Credential formats are given in (#alternative_credential_formats). 
+This specification supports any Credential format used in the Issuer-Holder-Verifier Model, including, but not limited to those defined in [@VC_DATA] (VCDM), [@ISO.18013-5] (mdoc), [@!I-D.ietf-oauth-sd-jwt-vc] (SD-JWT VC), and [@Hyperledger.Indy] (AnonCreds). Credentials of multiple formats can be presented in the same transaction. The examples given in the main part of this specification use W3C Verifiable Credentials, while examples in other Credential formats are given in (#alternative_credential_formats).
 
 Implementations can use any pre-existing OAuth 2.0 Grant Type and Response Type in conjunction with this specification to support different deployment architectures.
 
@@ -138,7 +138,7 @@ Any of the OAuth 2.0 related specifications, such as [@RFC9126] and [@RFC9101], 
 
 Below is a diagram of a flow where the End-User presents a Credential to a Verifier interacting with the End-User on the same device that the device the Wallet resides on.
 
-The flow utilizes simple redirects to pass Authorization Request and Response between the Verifier and the Wallet. The Verifiable Presentations are returned to the Verifier in the fragment part of the redirect URI, when Response Mode is `fragment`. 
+The flow utilizes simple redirects to pass Authorization Request and Response between the Verifier and the Wallet. The Verifiable Presentations are returned to the Verifier in the fragment part of the redirect URI, when Response Mode is `fragment`.
 
 Note: The diagram does not illustrate all the optional features of this specification.
 
@@ -146,7 +146,7 @@ Note: The diagram does not illustrate all the optional features of this specific
 ~~~ ascii-art
 +--------------+   +--------------+                                    +--------------+
 |     User     |   |   Verifier   |                                    |    Wallet    |
-+--------------+   +--------------+                                    +--------------+  
++--------------+   +--------------+                                    +--------------+
         |                 |                                                   |
         |    Interacts    |                                                   |
         |---------------->|                                                   |
@@ -164,7 +164,7 @@ Note: The diagram does not illustrate all the optional features of this specific
 !---
 Figure: Same Device Flow
 
-(1) The Verifier sends an Authorization Request to the Wallet. It contains a Presentation Definition as defined in [@!DIF.PresentationExchange] that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Authorization Request and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers consent to present the requested Credentials. 
+(1) The Verifier sends an Authorization Request to the Wallet. It contains a Presentation Definition as defined in [@!DIF.PresentationExchange] that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Authorization Request and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers consent to present the requested Credentials.
 
 (2) The Wallet prepares the Verifiable Presentation(s) of the Verifiable Credential(s) that the End-User has consented to. It then sends to the Verifier an Authorization Response where the Verifiable Presentation(s) are contained in the `vp_token` parameter.
 
@@ -207,11 +207,11 @@ Note: The usage of the Request URI as defined in [@!RFC9101] does not depend on 
 !---
 Figure: Cross Device Flow
 
-(1) The Verifier sends to the Wallet an Authorization Request that contains a Request URI from where to obtain the Request Object containing Authorization Request parameters. 
+(1) The Verifier sends to the Wallet an Authorization Request that contains a Request URI from where to obtain the Request Object containing Authorization Request parameters.
 
 (2) The Wallet sends an HTTP GET request to the Request URI to retrieve the Request Object.
 
-(2.5) The HTTP GET response returns the Request Object containing Authorization Request parameters. It especially contains a Presentation Definition as defined in [@!DIF.PresentationExchange] that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Request Object and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers her consent to present the requested Credentials. 
+(2.5) The HTTP GET response returns the Request Object containing Authorization Request parameters. It especially contains a Presentation Definition as defined in [@!DIF.PresentationExchange] that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Request Object and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers her consent to present the requested Credentials.
 
 (3) The Wallet prepares the Verifiable Presentation(s) of the Verifiable Credential(s) that the End-User has consented to. It then sends to the Verifier an Authorization Response where the Verifiable Presentation(s) are contained in the `vp_token` parameter.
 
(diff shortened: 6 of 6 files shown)


examples/query_lang/credentials_alternatives.json
```diff
@@ -0,0 +1,43 @@
...
+      {
+        "required": 1,
+        "from": [
+          { "ref": "pid_reduced" },
+          { "ref": "residence_info" }
+        ]
+      }
+    ]
+  },
+  "credentials": {
```
decentralgabe, line 15: how would someone specify claims from different formats?
if that is allowed, we should have an example.
if not allowed, then it seems the `format` property can be extracted to a higher level.

examples/query_lang/credentials_alternatives.json
```diff
@@ -0,0 +1,43 @@
...
+    "pid": {
```
decentralgabe, line 16: how would someone specify claims from different formats?
if that is allowed, we should have an example.
if not allowed, then it seems the `format` property can be extracted to a higher level.

examples/query_lang/multi_credentials.json
```diff
@@ -0,0 +1,42 @@
...
+            "format": "vc+sd-jwt",
+            "vc+sd-jwt": {
+                "vct_values": [ "https://example.com/personal_identity" ]
+            },
+            "purpose": [{
+                "lang": "en-US",
+                "text": "To prove the identity of the holder"
+            }],
+            "claims": [
+                { "path": [ "last_name" ] },
```
decentralgabe, line 13: are nested claims allowed here?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+# Verifiable Presentation Query Language {#vp_query}
```
decentralgabe, line 548: I would like to see this as a separate specification. It can be fully tested outside of OID4VP flows.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+- Claim query: An object describing a single claim that is to be returned, as
+  specified in (#claim_query).
+- An optionality rule: An object containing the following properties:
+  - `required` (REQUIRED): defines a number of claims that must be returned from
+    the `from` set of claims, as specified next.
+  - `from` (REQUIRED): An array; each entry is either a claim query or another
+    optionality rule. The entries are listed in order of preference, with the
+    first entry being the most preferred.
+
+More than two nested levels of optionality rules MUST NOT be used by a Verifier
```
decentralgabe, line 607: if credentials have >2 layers of nesting what's the solution?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+or allowed by a Wallet.
+
+## Format-specific Properties {#format_specific_properties}
+
+The following format-specific properties are defined:
+
+### Format `vc+sd-jwt` {#format_vc+sd-jwt}
```
decentralgabe, line 614: are we identifying formats by media type?
this is an unregistered media type

it may be better to have unique format identifiers that point to versions of specifications, defined by this specification.

decentralgabe (changes requested):
it's not clear to me that this is an improvement over presentation exchange.

are presentation submissions from PE still being used? just a different presentation definition format defined here?

I would recommend not merging this as is. My recommendation is to create a new dependent spec that fully defines a query language.

examples/query_lang/credentials_alternatives.json
danielfett: Good point, an example should be added. The other formats would just be separate credential request objects.

examples/query_lang/multi_credentials.json
danielfett: yes

openid-4-verifiable-presentations-1_0.md
danielfett: This works for any depth of nesting; the optionality rules are just capped to enable easier processing, but this doesn't reduce expressiveness of the language.

martijnharing: Does "two nested levels of optionality rules" mean that there is a total of 2 or 3 instances of "required" in a single "claim" element?

openid-4-verifiable-presentations-1_0.md
Sakurann: i don't think the intention of this new query language has been to be protocol independent, so not sure what would be the benefit of "fully testing outside of OID4VP flows" unless I am misinterpreting it..

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+A claim query is an object that describes one or more claims that are to be
+returned. Depending on the format of the requested Verifiable Credential, the
+claim query can contain different properties.
+
+
+
+### JSON-based Claims Structures {#json_based_claims}
+
+For credentials containing JSON-based claims structures, the following
```
awoie, line 654: We should clearly define what credential formats use this section rather than saying credentials using JSON-based claims structures.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+For ISO 18013-5 Credentials, the following properties are defined:
+
+`namespace`:
+: REQUIRED. A string that specifies the namespace of the claim
+within the Verifiable Credential, e.g., `org.iso.18013.5.1`.
+
+`claim_name`:
+: REQUIRED. A string that specifies the name of the claim within
+the Verifiable Credential, e.g., `first_name`.
```
awoie, line 671:
```suggestion
: REQUIRED. A string that specifies the name of the claim within the provided namespace
in the Verifiable Credential, e.g., `first_name`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,24 +252,27 @@ This specification enables the Verifier to send both Presentation Definition JSO
...
+: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. Exactly one of `vp_query` or `presentation_definition` or `presentation_definition_uri` MUST be present in the Authorization Request.
 
 `presentation_definition_uri`:
-: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is not present. See (#request_presentation_definition_uri) for more details.
+: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. See (#request_presentation_definition_uri) for more details. Exactly one of `vp_query` or `presentation_definition` or `presentation_definition_uri` MUST be present in the Authorization Request.
+
+`vp_query`:
+: A string containing a JSON-encoded VP Query as defined in (#vp_query). This parameter is OPTIONAL. Exactly one of `vp_query` or `presentation_definition` or `presentation_definition_uri` MUST be present in the Authorization Request.
```
awoie, line 261:
```suggestion
: A string containing a JSON-encoded VP Query as defined in (#vp_query). Exactly one of `vp_query` or `presentation_definition` or `presentation_definition_uri` MUST be present in the Authorization Request.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+contains the following properties:
+
+`lang:`
+: REQUIRED. A string that specifies the language of the purpose
+description. The value of this property MUST be a valid language tag as defined
+in [@!BCP47].
+
+`text`:
+: REQUIRED. A string that contains the purpose description in the
+language specified by the `lang` property.
```
awoie, line 772: IMO, free text is not a good idea for a lot of reasons, e.g., disambiguation between different translations of the the same reason and actually providing different reasons -> prone to errors; privacy; i18n/l18n which is hard to scale with this approach. A more appropriate approach is to define reason codes. There was a lot of work in other standards orgs on that, e.g., Kantara.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,24 +252,27 @@ This specification enables the Verifier to send both Presentation Definition JSO
 This specification defines the following new parameters:
 
 `presentation_definition`:
-: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. This parameter MUST be present when `presentation_definition_uri` parameter, or a `scope` value representing a Presentation Definition is not present.
+: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. Exactly one of `vp_query` or `presentation_definition` or `presentation_definition_uri` MUST be present in the Authorization Request.
```
TallTed, line 255:
```suggestion
: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. Exactly one of `vp_query`, `presentation_definition`, or `presentation_definition_uri` MUST be present in the Authorization Request.
```

openid-4-verifiable-presentations-1_0.md
TallTed: Including @awoie's change
```suggestion
: A string containing a JSON-encoded VP Query as defined in (#vp_query). Exactly one of `vp_query`, `presentation_definition`, `presentation_definition_uri` MUST be present in the Authorization Request.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,24 +252,27 @@ This specification enables the Verifier to send both Presentation Definition JSO
...
 
 `presentation_definition_uri`:
-: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is not present. See (#request_presentation_definition_uri) for more details.
+: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. See (#request_presentation_definition_uri) for more details. Exactly one of `vp_query` or `presentation_definition` or `presentation_definition_uri` MUST be present in the Authorization Request.
```
TallTed, line 258:
```suggestion
: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. See (#request_presentation_definition_uri) for more details. Exactly one of `vp_query`, `presentation_definition`, or `presentation_definition_uri` MUST be present in the Authorization Request.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+### Claims Path Pointer {#claims_path_pointer}
+
+A claims path pointer is a pointer into the JSON structure of the Verifiable
+Credential, identifying one or more claims. A claims path pointer MUST be a
+non-empty array of strings and non-negative integers. A string indicates that
+the respective key is to be selected, a `*` value indicates that all elements of
+the currently selected array(s) are to be selected, and a non-negative integer
+indicates that the respective index in an array is to be selected. The path is
+formed as follows:
```
TallTed, line 681:
```suggestion
non-empty array of strings and non-negative integers. A `*` string value indicates
that all elements of the currently selected array(s) are to be selected; any other
string indicates that the respective key is to be selected; and a non-negative
integer indicates that the respective index in an array is to be selected. The path
is formed as follows:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+: OPTIONAL.
+Additional properties requested by the Verifier that are specific to the
+Credential format, as defined in (#format_specific_properties). If omitted,
+no format-specific restrictions are placed on the requested Credential.
+
+`purpose`:
+: OPTIONAL. A purpose description array as defined in
+(#purpose_description).
+
+`claims`:
```
awoie, line 591: I'm wondering whether it is a good idea to have `claims` being a polymorphic object with a format-dependent object type, or if it would be better to have everything nested inside a format identifier object, e.g., `sd-jwt-vc: { claims ... }`?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +545,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+In detail, the array is processed by the Wallet from left to right as follows:
+
+ 1. Select the root element of the credential, i.e., the top-level JSON object.
+ 2. Process the query components from left to right:
+    1. If the query component is a string, select the element in the respective
+       key in the currently selected element(s). If any of the currently
+       selected element(s) is not an object, abort processing and return an
+       error. If the key does not exist in any element currently selected,
+       remove that element from the selection.
+    2. If the query component is `*`, select all elements of the currently
```
TallTed, line 743: It seems that processing the wildcard should precede any specific key string or integer index, none of which should be processed following the wildcard, as they would then be pointed to twice in the same query.

TallTed (changes requested):

openid-4-verifiable-presentations-1_0.md
danielfett: I would not put it into the format-specific part, as `claims` applies to either all JSON-based structures (which can be multiple formats) or all ISO-based structures (mdoc). We could rename to `claims` and `claims_iso` maybe.

openid-4-verifiable-presentations-1_0.md
decentralgabe: My thought is that a software implementation of a query language is distinct from an implementation for OID4VP. It could have its own conformance suite and open source tooling, and potentially, utility outside of this specification.

If those broader ambitions are shared amongst the group then breaking it out could be the right path.

examples/query_lang/credentials_alternatives.json
```diff
@@ -0,0 +1,43 @@
...
+      "format": "vc+sd-jwt",
+      "vct_values": ["https://credentials.example.com/identity_credential"],
+      "claims": [
+        { "path": ["last_name"] },
+        { "path": ["first_name"] },
```
selfissued, line 21:
```suggestion
        { "path": ["given_name"] },
        { "path": ["family_name"] },
```

examples/query_lang/credentials_alternatives.json
```diff
@@ -0,0 +1,43 @@
...
+        { "path": ["address", "street_address"] }
+      ]
+    },
+    "pid_reduced": {
+      "format": "vc+sd-jwt",
+      "vct_values": ["https://credentials.example.com/identity_credential"],
+      "claims": [
+        { "path": ["last_name"] },
+        { "path": ["first_name"] }
```
selfissued, line 30:
```suggestion
        { "path": ["family_name"] },
        { "path": ["given_name"] }
```

examples/query_lang/credentials_alternatives.json
```diff
@@ -0,0 +1,43 @@
...
+      ]
+    },
+    "residence_info": {
+      "format": "vc+sd-jwt",
+      "vct_values": ["https://credentials.example.com/residence_credential"],
+      "claims": [
+        { "path": ["zip_code"] },
+        { "path": ["city"] },
+        { "path": ["state"] }
```
selfissued, line 39:
```suggestion
        { "path": ["postal_code"] },
        { "path": ["locality"] },
        { "path": ["region"] }
```

examples/query_lang/multi_credentials.json
```diff
@@ -0,0 +1,42 @@
...
+                { "path": [ "first_name" ] },
```
selfissued, line 14:
```suggestion
                { "path": [ "given_name" ] },
```

examples/query_lang/multi_credentials.json
```diff
@@ -0,0 +1,42 @@
...
+            ]
+        },
+        "degree": {
+            "format": "vc+sd-jwt",
+            "vc+sd-jwt": {
+                "vct_values": [ "https://example.com/education_credential" ]
+            },
+            "claims": [
+                { "path": [ "last_name" ] },
+                { "path": [ "first_name" ] },
```
selfissued, line 25:
```suggestion
                { "path": [ "family_name" ] },
                { "path": [ "given_name" ] },
```

examples/query_lang/multi_credentials.json
```diff
@@ -0,0 +1,42 @@
...
+        "residence": {
+            "format": "vc+sd-jwt",
+            "vc+sd-jwt": {
+                "vct_values": [ "https://example.com/residence_proof" ]
+            },
+            "claims": [
+                { "path": [ "last_name" ] },
+                { "path": [ "first_name" ] },
+                { "path": [ "city" ] },
+                { "path": [ "state" ] }
```
selfissued, line 38:
```suggestion
                { "path": [ "family_name" ] },
                { "path": [ "given_name" ] },
                { "path": [ "locality" ] },
                { "path": [ "region" ] }
```

examples/query_lang/nested_claim_queries.json
```diff
@@ -0,0 +1,41 @@
...
+        { "path": [ "last_name" ] },
+        {
+          "required": 1,
+          "from": [
+            { "path": [ "zip_code" ] },
+            {
+              "required": 2,
+              "from": [
+                { "path": [ "city" ] },
+                { "path": [ "state" ] }
```
selfissued, line 33:
```suggestion
            { "path": [ "postal_code" ] },
            {
              "required": 2,
              "from": [
                { "path": [ "locality" ] },
                { "path": [ "region" ] }
```

examples/query_lang/simple.json
```diff
@@ -0,0 +1,17 @@
...
+    "my_cred_1": {
+      "format": "vc+sd-jwt",
+      "vc+sd-jwt": {
+        "vct_values": [ "https://credentials.example.com/identity_credential" ],
+        "alg_values": ["ES256", "ES384"],
+        "sd_alg_values": ["SHA-256"]
+      },
+      "claims":[
+        { "path": ["last_name"] },
+        { "path": ["first_name"] },
```
selfissued, line 12:
```suggestion
        { "path": ["family_name"] },
        { "path": ["given_name"] },
```

examples/query_lang/simple_mdoc.json
```diff
@@ -0,0 +1,22 @@
...
+        "hash_algorithm_values": [ "SHA-384"]
+      },
+      "claims": [
+        {
+          "namespace": "org.iso.7367.1",
+          "claim_name": "vehicle_holder"
+        },
+        {
+          "namespace": "org.iso.18013.5.1",
+          "claim_name": "first_name"
```
selfissued, line 17:
```suggestion
          "claim_name": "given_name"
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+#### Example
+
+The following shows a non-normative, simplified example of a Credential:
+
+```json
+{
+  "name": "Arthur Dent",
+  "address": {
+    "street_address": "42 Market Street",
+    "city": "Milliways",
```
selfissued, line 699:
```suggestion
    "locality": "Milliways",
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1973,6 +2244,33 @@ The following is a non-normative example of the payload of a Self-Issued ID Toke
...
+Credential in the format `mso_mdoc` with the claims `vehicle_holder` and
+`first_name`:
+
+<{{examples/query_lang/simple_mdoc.json}}
+
+The following is a non-normative example of a VP Query that shows the usage of
+the `purpose` property, and requests the claim `last_name` and
+
+- either the claim `zip_code`, or
+- the claims `city` and `state`:
```
selfissued, line 2259:
```suggestion
- either the claim `postal_code`, or
- the claims `locality` and `region`:
```

openid-4-verifiable-presentations-1_0.md
selfissued: Many of the JWT claim names used in the examples are wrong.  For instance, `first_name` is used, whereas the correct claim name is `given_name`.  I suggested changes.

examples/query_lang/credentials_alternatives.json
```diff
@@ -0,0 +1,43 @@
+{
+  "rules": {
+    "required": 2,
```
danielfett, line 3: Discussed on editors' call: add mapping to "and" and "or"

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+description. The value of this property MUST be a valid language tag as defined
+in [@!BCP47].
+
+`text`:
+: REQUIRED. A string that contains the purpose description in the
+language specified by the `lang` property.
+
+## Credential Rules {#credential_rules}
+
+Just as for claims, the Verifier can define rules that specify which
```
sloops77, line 773: We are using the PE submission_requirements feature. I see the intent here is to replace `submission_requirements` with `rules`. The features that we in HR use cases that are missing from rules are

1. ability to require `all` credentials of a particular credential type (eg. all transcripts) vs one of a particular type (eg mDL). From my understanding both would expressed as `{ "required": 1, "from": [{"ref": "transcript"}]}` and `{ "required": 1, "from": [{"ref": "mDL"}]}`
2. in HR use cases the user has more flexibility in what to share. therefore the ability to also require at least one is useful `{"min": 1, "from": [{"ref": "email"}]}`.
3. explaining the purpose of  a `rule` (e.g. the RP requires a mDL or national id card to get the applicants name)
4. is there a maximum depth to rules?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+: REQUIRED. An object. Each key in the object represents one
+Credential Query as defined in (#credential_query). The key is a string that
+uniquely identifies the Credential Query within the VP Query.
+
+`rules`:
+: OPTIONAL. An object defining rules that specify which combinations of
+credentials MAY be returned by the Wallet, as defined in (#credential_rules). If
+omitted, all credentials listed in `credentials` are requested for presentation.
+
+## Credential Query {#credential_query}
```
martijnharing, line 567: The credential query structure is defined as a polymorphic structure that contains both elements that always have the same content (purpose), elements that are partly format specific (claims) and elements that are fully format specific (everything else).
Since there is only a single element that is not currently format specific (purpose), we should consider making the request fully profile specific to simplify implementations.

openid-4-verifiable-presentations-1_0.md
sloops77: > ability to require all credentials of a particular credential type (eg. all transcripts) vs one of a particular type (eg mDL). From my understanding both would expressed as { "required": 1, "from": [{"ref": "transcript"}]} and { "required": 1, "from": [{"ref": "mDL"}]}
in HR use cases the user has more flexibility in what to share. therefore the ability to also require at least one is useful {"min": 1, "from": [{"ref": "email"}]}.

As per a simiilar suggestion above `required` should be `all`, `optional`, `one`, `at_least_one` with the default either being `all` or `optional`. An alternative if a number is required for certain use cases would be to replace `one` with `count` which would then have an additional property with the expected `count`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+least one claim that MUST be returned by the Wallet to fulfill the credential
+request (unless specified otherwise by extensions to this specification). Each
+entry MUST be one of the following objects:
+
+- Claim query: An object describing a single claim that is to be returned, as
+  specified in (#claim_query).
+- An optionality rule: An object containing the following properties:
+  - `required` (REQUIRED): defines a number of claims that must be returned from
+    the `from` set of claims, as specified next.
+  - `from` (REQUIRED): An array; each entry is either a claim query or another
```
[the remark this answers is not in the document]
danielfett: I'm not sure I fully understand the proposal - could you please give an example?

openid-4-verifiable-presentations-1_0.md
danielfett: Could you please propose a change introducing reason codes so that we can discuss on the calls?

openid-4-verifiable-presentations-1_0.md
danielfett: The `at_least_one` case is not currently captured, but we should consider introducing it and therefore moving back to keywords instead of numbers. Do you have a concrete use case in mind for `at_least_one`?

danielfett: How would a purpose for the rule be exposed to the user?

sloops77: @danielfett the current use case we have is in the application for a job. In the recruitment use case the holder can share whatever information they feel is relevant. For example, a person may have had studied and achieved a number of diplomas, but one or more maybe irrelevant and/or perhaps detrimental to them being successful in the application. The recruiter therefore will request at least one degree but leave it up to the holder to select the relevant one or two. 

Again the recruitment use case means that certain types of credentials are needed for a purpose. We have a UI in the reference wallet in the HR use case that takes the user through a wizard-like process to meet each rule requirement. A purpose for that rule is required for this UX to make sense

openid-4-verifiable-presentations-1_0.md
martijnharing: In order to provide reason codes that can help users guide in how to respond in a meaningful way, we need to find discuss the proper places where to define such reason codes. These reason codes are likely to be use case and/or document type specific. Therefore one approach could be to define reason codes in such a way that they can be scoped/namespaced according to the entity defining them. 
Separate to that we should also decide whether we can define or reference a set of such codes that could apply more generically.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
martijnharing: Hereby some examples on what it could look like. Option 1 is the current proposal, option 2 and 3 are alternatives. 

```
; option (1) What we have in the current proposal

claims = [+claim]   ; you can ask for multiple "claim", each "claim" MUST be satisfied

claim = claim_query / optionality_rule  ; claim_query OR optionality_rule
optionality_rule = {
    "required": uint,   ; number of claims to be returned from the 'from' set
    "from": claim       ; recursively contains the claim structure
}

claim_query = claim_query_json / claim_query_mdoc   ; claim_query_json OR claim_query_mdoc

claim_query_json = {
    "path": tstr
}

claim_query_mdoc = {
    "namespace": tstr,
    "claim_name": tstr

}


; option (2) For each claim, define the main element, then define alternates for that claim

claims = [+claim]   ; you can ask for multiple "claim", each "claim" MUST be satisfied

claim = {
    "claim_query" = claim_query,            ; claim_query as defined in (1)
    "alternate_claims" = [+ claim_query]    ; Alternate claims, in order of preference
}


; option (3) For each claim, define an array 

claims = [+claim]       ; you can ask for multiple "claim", each "claim" MUST be satisfied
claim = [+claim_query]  ; to satisfy the "claim" array, one element from the array (e.g. one claim_query) must be returned

```

Note, what is the best solution also depends on the outcome of the other discussions. Specifically the ones regarding intent_to_retain, if_present and whether to have the claim_query be polymorphic or to have different versions of the "claims" element, e.g. claims_json and claims_mdoc.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+`https://credentials.example.com/identity_credential` and the claims `last_name`,
+`first_name`, and `address.street_address`:
+
+<{{examples/query_lang/simple.json}}
+
+Additional examples can be found in (#vp_query_examples).
+
+## Additional Request Properties {#additional_request_properties}
+
+`intend_to_retain`: TBD
```
c2bo, line 801:
```suggestion
`intent_to_retain`: TBD
```

openid-4-verifiable-presentations-1_0.md
danielfett: I must admit that I lack the imagination how a set of codes could ever be complete. "Complete" here meaning in particular that I can use them to refer to all possible reasons for obtaining data through a wallet. Just as a few examples:

- "Required for account opening according to [law XYZ]" — we cannot enumerate all possible laws; also many laws and their applicability will be dependent on the region.
- "Age verification according to [regulatory body XYZ]"
- "Required only for obtaining [product  XYZ]"
- "Alternative when not in possession of [credential XYZ]"

examples/query_lang/simple_mdoc.json
danielfett: This was taken from an ISO example, where the claim was called 'first_name'.

openid-4-verifiable-presentations-1_0.md
martijnharing: Agreed that it's difficult / unlikely to have reasons codes that are both very specific and can address every possible situation. However, allowing free text is not a solution to achieve that as it introduces privacy and security problems that now need to be addressed. it's  a lot more difficult for the wallet to meaningfully use a free form text field to help guide the user compared to pre-defined reason codes.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
martijnharing: As discussed yesterday, hereby some examples for these structures. Note that other solutions besides option 2 or option 3 may exist that are more appropriate, especially when we consider the other comments on this PR.

```
Example 3: given_name, age_over_18 OR birth_date OR age_birth_year

option(1)

claims = [
    { "path": ["given_name"] },
    {
        "required": 1,
        "from": [
            { "path": ["age_over_18"] },
            { "path": ["birth_date"] },
            { "path": ["age_birth_year"] }
        ] 
    }
]


option(2)

claims = [ 
    {
        "claim_query": { "path": ["given_name"] }   
    },
    {
        "claim_query": { "path": ["age_over_18"] },
        "alternate_claims": [
            { "path": ["birth_date"] },
            { "path": ["age_birth_year"] }
        ]  
    },  
]

option(3)

claims = [ 
    [ { "path": ["given_name"] } ]
    [             
        { "path": ["age_over_18"] },
        { "path": ["birth_date"] },
        { "path": ["age_birth_year"] }
    ]
]



Example 4: given_name, zip_code OR (city AND state)

option(1)

claims = [
    { "path": ["given_name"] },
    {
        "required": 2,
        "from": [
            { "path": ["zip_code"] },
            {
                "required": 1,
                "from": [
                    { "path": ["city"] },
                    { "path": ["state"] },
                ]
            }
        ] 
    }
]

option(2)

claims = [ 
    {
        "claim_query": { "path": ["given_name"] }   
    },
    {
        "claim_query": { "path": ["zip_code"] },
        "alternate_claims": [
            { "path": ["city"] },
        ]  
    },  
    {
        "claim_query": { "path": ["zip_code"] },
        "alternate_claims": [
            { "path": ["state"] },
        ]  
    }
]

option(3)

claims = [ 
    [ { "path": ["given_name"] } ]
    [             
        { "path": ["zip_code"] },
        { "path": ["city"] },
    ],
    [             
        { "path": ["zip_code"] },
        { "path": ["state"] },
    ]
]
```

openid-4-verifiable-presentations-1_0.md
tlodderstedt: In my experience, providing a meaningful purpose to a user requires customization of the respective text to the particular RP and process, even the specific transaction in some cases. Yes, free text introduces security risks, but those are not unsolvable, developers cope with them every day when building applications with dynamic input. The multi language aspect is a true challenge, it would most likely require the RP sending the purpose string in multiple languages.  
Pre-defined reasons code on the other hand posse their own challenges: 
1) There needs to be registries for those codes. Which parties operate these registries?
2) What does a RP need to do in order to be able to send the first presentation request if this RP wants to implement a use case no one has implemented/standardized before? I think requiring pre-defined reason codes would create a huge administrative burden to RPs and delay deployment of new integrations in an inacceptable manner. 
3) How are those code distributed along with the respective text? What is the update policy? 
4) In the simplest case, there is no link possible to an actual transaction performed by a RP.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+language that allows the Verifier to request the Wallet to present Verifiable
+Credentials that match the query. The query consists of a set of constraints
+that the requested Verifiable Credentials must satisfy. The Wallet evaluates the
+query against the Verifiable Credentials it holds and returns the Verifiable
+Presentations that match the query.
+
+A valid VP Query is defined as a JSON-encoded object with the following
+properties:
+
+`credentials`:
```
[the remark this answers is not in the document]
Sakurann: i think it is pretty common to use strings even when they are not meant for human consumption. it has also been a convention throughout openid4vc specs: credential_configuration_id and credential_identifier in VCI just being some of the examples. I do not see why deviate from that here and would prefer to keep it a string.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+`sd_alg_values`:
+: OPTIONAL. An array of strings that specifies the allowed hash algorithms for
+digests in the SD-JWT and KB-JWT.
+
+### Format `mso_mdoc` {#format_mso_mdoc}
+
+TBD
+
+### Format `jwt_vp*`
```
Sakurann, line 637:
```suggestion
### Format `jwt_vp`
```

openid-4-verifiable-presentations-1_0.md
Sakurann: assuming by profile specific, you meant format specific? I do not understand how making everything format specific would simplify things... I think the feedback we have been getting is actually in the opposite direction of removing "partly format specific" part and use `path` for mdocs too, instead of `namespaces` and `data_elements` claims. the rest of the structure of defining the framework in a format agnostic way, while allowing format specific content for certain claims (like for the algorithms supported) makes sense to me. path vs namespace/data_elements feels pretty unnatural.

openid-4-verifiable-presentations-1_0.md
Sakurann: i would prefer using path for mdocs too to solve this weird deviation between the formats for this top-level parameter layer

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+A purpose description array is an object that describes the purpose for which
+the Verifier is requesting the Verifiable Credential. It can contain multiple
+objects, each describing the purpose in a specific language. Each object
+contains the following properties:
+
+`lang:`
+: REQUIRED. A string that specifies the language of the purpose
+description. The value of this property MUST be a valid language tag as defined
+in [@!BCP47].
```
selfissued, line 765: I suggest using the identifier `locale` as is done in OpenID Connect.  The description there is
> End-User's locale, represented as a BCP47 [RFC5646] language tag. This is typically an ISO 639 Alpha-2 [ISO639] language code in lowercase and an ISO 3166-1 Alpha-2 [ISO3166‑1] country code in uppercase, separated by a dash. For example, en-US or fr-CA. As a compatibility note, some implementations have used an underscore as the separator rather than a dash, for example, en_US

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
deshmukhrajvardhan: Does `,` in the above mean AND/required? I think the description text should be more specific and informative.
Also, does the `Example 4` mean this `given_name AND (zip_code OR (city AND state))`
seems like in option 3 `zip_code OR city` and `zip_code OR state` could result in zip_code and zip_code. Resulting in not returning city or state.

Can we redefine option 3 by swapping the meaning of things in the 1st and 2nd level `[ ]`?
E.g. given_name AND (zip_code OR (city AND state))
```
redefining option 3
claims = [ 
    [ 
        { "path": ["given_name"] },
       { "path": ["zip_code"] },
    ]
    [             
       { "path": ["given_name"] },
       { "path": ["state"] },
       { "path": ["city"] },
    ],
]
```

This would make it simpler as entries in the root level `[ ]` would mean OR and the 2nd level will mean MUST.
In the above it translates to `(given_name and zipcode) OR (given_name and zip_code and city)`

Similarly for `Example 3: given_name and (age_over_18 OR birth_date OR age_birth_year)`
```
redefining option 3
claims = [
  [ 
        { "path": ["given_name"] },
       { "path": ["age_over_18"] },
  ],
 [ 
        { "path": ["given_name"] },
       { "path": ["birth_date"] },
  ],
  [ 
        { "path": ["given_name"] },
       { "path": ["age_birth_year"] },
  ],
```
This also corresponds to all the potential credentials User can consent to send. So it won't result in cases where `zip_code and zip_code` are sent.
I recognize that this could cause the query to be longer than other methods in few cases.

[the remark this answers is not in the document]
deshmukhrajvardhan: I think it would be more clear to use a notation (req) instead of `and` 
as in boolean "zip_code" and "zip_code" translates to just 1 "zip_code"

https://grace.bluegrass.kctcs.edu/~kdunn0001/files/Simplification/4_Simplification_print.html

Where as in (req)zip_code (req)zip_code it is sent twice?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -542,6 +542,266 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+  - Start with an empty array.
+  - To address a particular claim within an object, append the key (claim name)
+    to the array.
+  - To address an element within an array, append the index to the array (as a
+    non-negative, 0-based integer).
+  - To address all elements within an array, append the string `*` to the array.
+
+Verifiers MUST NOT point to the same claim more than once in a single query.
+Wallets SHOULD ignore such duplicate claim queries.
```
TallTed, line 688: I think it very likely that verifiers will accidentally point to the same claim more than once, likely when mixing claim indexes with claim names, and certainly when mixing the wildcard with either or both claim indexes and/or claim names.

It might be viable to forbid use of the wildcard in combination with integers and/or strings, but that is not sufficient to cover all scenarios.

```suggestion
Verifiers SHOULD NOT point to the same claim more than once in a single query,
e.g., by combining the wildcard `*` with one or more claim names or indexes,
or by using claim names *and* indexes.
Wallets SHOULD return each claim once, regardless of duplicate claim queries.
```

jogu: Hi @TallTed I've hidden this comment because as I replied to a previous comment a week or two ago we don't appear to have a contribution agreement in place for you so can't use your contributions. If you'd like you contribute you can complete the contribution agreement here: https://openid.net/intellectual-property/openid-foundation-contribution-agreements/

If you are already covered by a contribution agreement, please let me know the details so we can locate it. Thank you!

danielfett:
Please review this new approach: https://github.com/openid/OpenID4VP/pull/260

danielfett:
I'll close this PR as it seems that there is consensus that #266 is the better approach.

PR #233, opened 2024-08-25
Clarify what can go in client_metadata authz request parameter

### LINKED ISSUE #17: what metadata goes into client_metadata parameter?

currently we just say that any client metadata can go into a `client_metadata` parameters, but I think parameters like `redirect_uris`, `response_types`, `grant_types` should be prohibited to be included.

jogu:
As per working group consensus agreed on 21st May call:

https://github.com/openid/OpenID4VP/issues/17#issuecomment-2123350290

closes #17

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -261,9 +261,13 @@ This specification defines the following new parameters:
 : OPTIONAL. A string identifying the scheme of the value in the `client_id` Authorization Request parameter (Client Identifier scheme). The `client_id_scheme` parameter namespaces the respective Client Identifier. If an Authorization Request uses the `client_id_scheme` parameter, the Wallet MUST interpret the Client Identifier of the Verifier in the context of the Client Identifier scheme. If the parameter is not present, the Wallet MUST behave as specified in [@!RFC6749]. See (#client_metadata_management) for the values defined by this specification. If the same Client Identifier is used with different Client Identifier schemes, those occurrences MUST be treated as different Verifiers. Note that the Verifier needs to determine which Client Identifier schemes the Wallet supports prior to sending the Authorization Request in order to choose a supported scheme.
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. It may contain the following:
 
-A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` request parameter.
+    * `jwks`: OPTIONAL. A JWKS that can contain a public key to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This does not replace any JWKS for the Verifier that the Wallet has obtained from other sources (e.g. keys for verifying signed requests may have been obtained from [@!OpenID.Federation]).
+    * `vp_formats`: REQUIRED when not available to the wallet via another mechanism. As defined in (#client_metadata_parameters).
+    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
 
 `request_uri_method`: 
 : OPTIONAL. A string determining the HTTP method to be used when the `request_uri` parameter is included in the same request. Two case-sensitive valid values are defined in this specification: `get` and `post`. If `request_uri_method` value is `get`, the Wallet MUST send the request to retrieve the Request Object using the HTTP GET method, i.e., as defined in [@RFC9101]. If `request_uri_method` value is `post`, a supporting Wallet MUST send the request using the HTTP POST method as detailed in (#request_uri_method_post). If the `request_uri_method` parameter is not present, the Wallet MUST process the `request_uri` parameter as defined in [@RFC9101]. Wallets not supporting the `post` method will send a GET request to the Request URI (default behavior as defined in [@RFC9101]). `request_uri_method` parameter MUST NOT be present if a `request_uri` parameter is not present.
@@ -2047,6 +2051,10 @@ The technology described in this specification was made available from contribut
 # Document History
 
    [[ To be removed from the final specification ]]
+
+   -22
+
+   * clarified what can go in the `client_metadata` parameter
 
    -21
 


openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,13 @@ This specification defines the following new parameters:
 : OPTIONAL. A string identifying the scheme of the value in the `client_id` Authorization Request parameter (Client Identifier scheme). The `client_id_scheme` parameter namespaces the respective Client Identifier. If an Authorization Request uses the `client_id_scheme` parameter, the Wallet MUST interpret the Client Identifier of the Verifier in the context of the Client Identifier scheme. If the parameter is not present, the Wallet MUST behave as specified in [@!RFC6749]. See (#client_metadata_management) for the values defined by this specification. If the same Client Identifier is used with different Client Identifier schemes, those occurrences MUST be treated as different Verifiers. Note that the Verifier needs to determine which Client Identifier schemes the Wallet supports prior to sending the Authorization Request in order to choose a supported scheme.
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. It may contain the following:
```
martijnharing, line 264: "It MAY contain the following:" is somewhat ambiguous.

Is it the intent to say
"It MAY contain the following, other parameters MUST NOT be used."
or
"It MAY contain the following, other parameters MAY be used."
Or something else?

jogu: Thanks for the feedback Martijn!

It's actually the 'something else' option, other values MUST NOT be used (because, for example, a self-asserted `client_name` passed in the authorization request just opens up phishing attacks without really achieving any useful) - but we do want to leave it open for other specifications to define new values that can be used.

I've pushed new text, please let me know what you think of it.

martijnharing: Is the following an accurate interpretation of that?

OpenID4VP defines that the client_metadata parameter contains metadata with values defined in:
- Section 4.3 of the OpenID4VP specification
- Section 2.1 of the OpenID Connect dynamic Client Registration specification 1.0 (should this be section 2.0?)
- RFC 7591 (should we explicitly call out section 2 here?)

The intent is to say that out of the values that are defined in these sections, only the 5 parameters defined in the PR are allowed to be used and others are explicitly not allowed. But any value that's not on the list of the referenced section may also be used, but their use is out of scope of the OpenID4VP specification?

And the following is somewhat pedantic, but if there is a PR it's good to clear it up.

- The OpenID4VP specification references section mentions "OpenID Connect Dynamic Client Registration 1.0 incorporating errata set 1", 8 November 2014,", but the link links to "https://openid.net/specs/openid-connect-registration-1_0.html" which incorporates errata set 2.
- Both errata 1 and errata 2 have a line saying that additional client metadata values are allowed. Specific specs are referenced between errata 1 and errata 2. Are the parameters in these specifications ([OpenID Connect Session Management 1.0 and [OpenID Connect RP-Initiated Logout 1.0]) also not allowed because they are referenced in the client registration specification?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new parameters:
...
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
-A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` request parameter.
+    * `jwks`: OPTIONAL. A JWKS that can contain a public key with the `"use": "enc"` claim to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This does not replace any JWKS for the Verifier that the Wallet has obtained from other sources (e.g. keys for verifying signed requests may have been obtained from [@!OpenID.Federation]). This allows the verifier to pass an empheral encryption key that is only used for this authorization request.
```
paulbastian, line 266:
```suggestion
    * `jwks`: OPTIONAL. A JWKS that can contain a public key with the `"use": "enc"` claim to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This does not replace any JWKS for the Verifier that the Wallet has obtained from other sources (e.g. keys for verifying signed requests may have been obtained from [@!OpenID.Federation]). This allows the Verifier to pass an ephemeral encryption key that is only used for this Authorization Request. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new parameters:
...
+    * `vp_formats`: REQUIRED when not available to the wallet via another mechanism. As defined in (#client_metadata_parameters).
+    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+
+    Values passed in `client_metadata` MUST NOT override authorative data the wallet is able to obtain about the client from other sources, for example those from an OpenID Federation Entity Statement.
+
+    Other metadata parameters defined in, for example, Section 4.3 and Section 2.1 of the OpenID Connect Dynamic Client Registration 1.0 [@!OpenID.Registration] specification or [@!RFC7591] MUST NOT be used unless a profile of this specification explicitly defines them as usable in the `client_metadata` parameter.
```
Sakurann, line 274:
```suggestion
    Other metadata parameters MUST NOT be used unless a profile of this specification explicitly defines them as usable in the `client_metadata` parameter.
```

openid-4-verifiable-presentations-1_0.md
bc-pi, line 266: https://datatracker.ietf.org/doc/html/rfc7517#section-4.2 not all JSON things are claims

also spellcheck caught an ephemeral typo 

```suggestion
    * `jwks`: OPTIONAL. A JWKS that can contain a public key with the `"use": "enc"` parameter to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This does not replace any JWKS for the Verifier that the Wallet has obtained from other sources (e.g. keys for verifying signed requests may have been obtained from [@!OpenID.Federation]). This allows the verifier to pass an ephemeral encryption key that is only used for this authorization request. 
```

Sakurann, line 266:
```suggestion
    * `jwks`: OPTIONAL. A JWKS that can contain a public key with the `"use": "enc"` parameter to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This does not replace any JWKS for the Verifier that the Wallet has obtained from other sources (e.g. keys for verifying signed requests may have been obtained from [@!OpenID.Federation]). This allows the verifier to pass an ephemeral encryption key that is only used for this authorization request. 
```
point to RFC7591 and clarify multiple encryption keys are ok. and maybe clarify `this does not replace any JWKS...` part per @awoie comment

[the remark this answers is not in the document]
peppelinux: I prefer "allows". @awoie it would be better if your commit would take also paul's and brian's ones

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new parameters:
...
+    * `jwks`: OPTIONAL. A JWKS that can contain a public key with the `"use": "enc"` claim to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This does not replace any JWKS for the Verifier that the Wallet has obtained from other sources (e.g. keys for verifying signed requests may have been obtained from [@!OpenID.Federation]). This allows the verifier to pass an empheral encryption key that is only used for this authorization request. 
+    * `vp_formats`: REQUIRED when not available to the wallet via another mechanism. As defined in (#client_metadata_parameters).
+    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+
+    Values passed in `client_metadata` MUST NOT override authorative data the wallet is able to obtain about the client from other sources, for example those from an OpenID Federation Entity Statement.
```
peppelinux, line 272:
```suggestion
    Values passed in `client_metadata` MUST NOT override authoritative data the Wallet is able to obtain about the Client from other sources, for example those from an OpenID Federation Entity Statement.
```

peppelinux: @jogu here I would add an additional guidance for the implementers about how to deal with overlapping metadata, one obtained within the request and another one from an authoritative sources such as Federation or any other trusted metadata registry.

We may apply a rule based on the json object update

we have two objects in an hierarchy of priority: json_authoritative and json_ephemeral, the first from Federation and the latter from the request.

````python

In [1]: json_authoritative = {"a": 1, "b": 2, "c": 3}

In [2]: json_ephemeral = {"a": 2, "b":1, "d":5}

In [3]: json_ephemeral.update(json_authoritative)

In [4]: json_ephemeral
Out[4]: {'a': 1, 'b': 2, 'd': 5, 'c': 3}
````
therefore we may say that authoritative one must overwrite the ephemeral one.
As we may notice, the parameters that are not present in the authoritative but in the ephemeral, will be added to the authoritative one. The weakness of this approach is that the authoritative is therefore "updated"  due to the merge with the ephemeral one, and this eluded the federation policies.

For this reason we should say something about this, and therefore distinguish two different moments:

1. merge the two metadata having the federation one as priority
2. apply again the trust chain policies

This might find space in the [openid federation wallet architecure draft](https://github.com/peppelinux/federation-wallet), for its specific purpose on this topic

File: openid-4-verifiable-presentations-1_0.md (hunk 3 above)
Comment (peppelinux) [inline openid-4-verifiable-presentations-1_0.md:274]:
```suggestion
    Other metadata parameters defined in, for example, Section 4.3 and Section 2.1 of the OpenID Connect Dynamic Client Registration 1.0 [@!OpenID.Registration] specification or [@!RFC7591] SHOULD NOT be used unless a profile of this specification explicitly defines them as usable in the `client_metadata` parameter.
```

I have changed my mind on this, according to OAuth philosophy things that you don't understand you ignore. It would be better to not break this rule to facilitate extensions based on openid4vp

openid-4-verifiable-presentations-1_0.md
```diff
@@ -727,7 +735,7 @@ The JWT response document MUST include `vp_token` and `presentation_submission`
 
 The key material used for encryption and signing SHOULD be determined using existing metadata mechanisms. 
 
-To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` or `jwks_uri` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms.
+To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms.
```
peppelinux, line 738:
```suggestion
To securely encrypt the Authorization Response, the Wallet uses the 'jwks` parameter within the `client_metadata` request parameter. This is REQUIRED if no other methods are available to retrieve the Verifier's metadata, such as through the trust chain described in [@!OpenID.Federation].
```
Generally metadata in Federation are only final metadata derived from the trust chain.

peppelinux (changes requested):
I approve this PR, I ask changes from an editorial perspective and for some further pedantic clarifications

openid-4-verifiable-presentations-1_0.md
Sakurann: I think what you are trying to get at, Giuseppe, is totally possible with a current text, you just need an extension or a profile says why it is ok to use a specific parameter. i would have a strong preference to keep it must not.

openid-4-verifiable-presentations-1_0.md
Sakurann: > therefore we may say that authoritative one must overwrite the ephemeral one.

I think the text already says that?

Sakurann:
```suggestion
    Authoritative data the wallet is able to obtain about the client from other sources, for example those from an OpenID Federation Entity Statement, overrides the values passed in `client_metadata`.
```

we could paraphrase like this, but I am hesitant to add anything more than that

jogu: I think Giuseppe is asking: what happens if an encryption key is specified both in the request and in OpenID Federation?

I think this is a bigger topic that probably needs a specific issue and separate PR.

There are two conflicting requirements:

1. Some people want to use per-session encryption keys
2. Some people want to ensure that only encryption keys from federation are used

I am pretty sure it is not possible to put an ephemeral key into federation metadata (it's just technically not feasible to have the federation metadata for a client potentially changing multiple times in a second?), so I'm not sure I see an obvious way to satisfy both sets of people.

jogu: I'll apply something along the lines of Kristina's suggestions.

openid-4-verifiable-presentations-1_0.md
jogu: I've changed to 'MUST be ignored" which I think better reflects the intent and extensibility. As Kristina says, it is very deliberate to make clear people must not just use other already defined client metadata values.

openid-4-verifiable-presentations-1_0.md
jogu: This reads like encryption is mandatory, but we've not agreed that as a WG. Also I don't think 'securely encrypt' should be used, it's just encryption. As this is existing text (I only removed 'jwks_uri' from this sentence) I think we should handle changing this in a separate issue.

jogu:
Thanks everyone for this comments - this is ready for re-review! I've applied most suggestions, other than some I've commented that I think we should handle separately if they're needed.

openid-4-verifiable-presentations-1_0.md
Sakurann: how about paraphrasing like this?

```suggestion
To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, if no other mechanisms such as the metadata defined in the Entity Configuration in [@!OpenID.Federation] are used.
```

Sakurann (approved):
LGTM - thanks a lot, Joseph!

c2bo:
Linking #251 here: We still have a dedicated section (Section 9) for `client_metadata` which is currently linked to only in section 5.1 and has less detail than the introduction done here. Should we just remove section 9 then?

openid-4-verifiable-presentations-1_0.md
jogu: @Sakurann I'm not sure how to read this. Does it mean "if the verifier has a jwks in its federation metadata it can't use ephemeral keys"? If so I think that would be a normative change that I'm not sure I'm comfortable making in this PR.

Sakurann:
I don't think there is a need to remove section 9. that one defines a new parameter and talks about discovery.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new parameters:
...
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
-A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` request parameter.
+    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591] that can contain one or more public keys with the `"use": "enc"` parameter to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This allows the verifier to pass an empheral encryption key that is only used for this authorization request. Public keys included in the `jwks` parameter MUST NOT be used to verify the signature of signed Authorization Requests.
+    * `vp_formats`: REQUIRED when not available to the wallet via another mechanism. As defined in (#client_metadata_parameters).
```
peppelinux, line 267:
```suggestion
    * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
```

peppelinux (approved):
Approved in advance with some pending suggestions I feel familiar to get included

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new parameters:
...
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
-A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` request parameter.
+    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591] that can contain one or more public keys with the `"use": "enc"` parameter to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This allows the verifier to pass an empheral encryption key that is only used for this authorization request. Public keys included in the `jwks` parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```
awoie, line 266:
```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591] that can contain one or more public keys with the `"use": "enc"` parameter and `"alg":"ECDH-ES" parameter, to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This allows the verifier to pass an ephemeral encryption key that is only used for this authorization request. Public keys included in the `jwks` parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```

awoie, line 266: With ISO mdocs, this key is also used with ECDH + HMAC; not only for JARM. Does this language prevent the usage of this? I assume some similar usage is done also with SD-JWT VC @paulbastian ?
```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591] that can contain one or more public keys with the `"use": "enc"` parameter to be used by the Wallet as an input to the key agreement to encrypt the Authorization Response (see (#jarm)). This allows the verifier to pass an empheral encryption key that is only used for this authorization request. Public keys included in the `jwks` parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```

awoie (commented):
Generally looks good. I just had some questions on the limitations of the usage of the keys in the jwks parameter.

openid-4-verifiable-presentations-1_0.md
paulbastian, line 266:
```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591] that can contain one or more public keys with the `"use": "enc"` parameter to be used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm) or for signature algorithms that require a public key of the Verifier). This allows the verifier to pass an empheral encryption key that is only used for this authorization request. Public keys included in the `jwks` parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```

paulbastian (approved):
I approve given that the key usage for `jwks` for signature algorithms is accepted

openid-4-verifiable-presentations-1_0.md
jogu: I've applied this but generalised this a bit so that keys without `"use": "enc"` are permitted too, as I think that's what this/some algorithms require - please check the new text, thanks!

openid-4-verifiable-presentations-1_0.md
jogu: I'm somewhat confused as to what the latest suggestion here is, but I think if people think updates are needed to the text around how keys from Federation overlap with keys from `client_metadata` parameter then  we should handle that in a separate issue please as that's outside of the scope of this PR and hasn't been discussed in the working group yet. This PR does not make anything worse in that area.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new parameters:
...
+    * `vp_formats`: REQUIRED when not available to the wallet via another mechanism. As defined in (#client_metadata_parameters).
+    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+
+    Authoritative data the wallet is able to obtain about the client from other sources, for example those from an OpenID Federation Entity Statement, take precedence over the values passed in `client_metadata`.
```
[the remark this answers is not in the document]
jogu: change applied, thanks!

openid-4-verifiable-presentations-1_0.md
jogu: Change applied, thanks!

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
jogu: I've combined this with Paul/Oliver's suggestions and applied, thanks!

jogu:
I've applied all comments now (other than two about the overlap with federation that I believe aren't in scope nor agreed with WG yet and have responded saying so).

@awoie could you check the new text please and approve if you're happy? It should allow for both the mdoc & jwt mac based signature you/Paul mentioned.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new request parameters:
...
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
-A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` request parameter.
+    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as keys with the `"use": "enc"` parameter used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm)), or keys for signature algorithms that require a public key of the Verifier. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```
paulbastian, line 266:
```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as keys  used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm)), or keys for signature algorithms that require a public key of the Verifier. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```
I think the discussion about the use parameter is misleading and unnecessary detail here, as it is defined in JARM anyway

jogu: Updated version of Paul's suggestion just to remove a double space:

```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as keys used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm)), or keys for signature algorithms that require a public key of the Verifier. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,9 +261,17 @@ This specification defines the following new request parameters:
...
 
 `client_metadata`:
-: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
+: OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
-A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` request parameter.
+    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as keys used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm)), or keys for signature algorithms that require a public key of the Verifier. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```
awoie, line 266:
```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as keys used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm)), or for proofs generated by the Wallet, which are incorporated into the Verifiable Presentations and require the public key of the Verifier. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```

awoie (approved):
lgtm. it would be great if we could incorporate my suggestion.

openid-4-verifiable-presentations-1_0.md
jogu: Thanks Oliver! I think this is a bit difficult to follow. I presume it's the "signature algorithms" part that you weren't happy with? If so how about this:

```suggestion
    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as those used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#jarm)), or where the Wallet will require the public key of the Verifier to generate the Verifiable Presentation. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
```

paulbastian (approved):

PR #234, opened 2024-08-25
Better define response uri

### LINKED ISSUE #111: Response Endpoint may not be suffciently described
as discussed on DCP call

jogu:
As per WG consensus on 4th April 2024 call, replace Response Endpoint with Response URI (to be consistent with response_uri parameter name), and update the definition of response_uri to say it is a URL.

closes #111


nemqe (approved):

Sakurann (approved):
I don't think i was on a call that agreed on this, but this totally makes sense and makes things so much easier to read.

c2bo (approved):

Issue #221, opened 2024-07-25
"Examples with Credentials in Various Formats" section should be renamed as it contains normative text


jogu:
The "Examples with Credentials in Various Formats" section should be renamed as it contains normative text, e.g. the names of the  Credential format identifiers.

deshmukhrajvardhan:
A name is an identifier, so can the text be "Credential format Identifiers"? or "Creating and using Credential format Identifiers"

PR #238, opened 2024-08-27
Rename Appendix B

deshmukhrajvardhan:
Closes https://github.com/openid/OpenID4VP/issues/221

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1646,7 +1646,7 @@ if (credential.protocol == "openid4vp") {
 }
 ```
 
-# Credential Format Identifiers {#alternative_credential_formats}
+# Credential Format Profiles {#alternative_credential_formats}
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1646,7 +1646,7 @@ if (credential.protocol == "openid4vp") {
 }
 ```
 
-# Examples with Credentials in Various Formats {#alternative_credential_formats}
+# Credential Format Identifiers {#alternative_credential_formats}
```
nemqe, line 1649: is this a good name for a section that explains and defines more than just client identifiers (every credential format has sections **credential identifier**, **credential metadata**, **authorization request**, etc..)

c2bo:
How about something like "Credential Format specific Parameters"?

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
# Credential Format Profiles {#alternative_credential_formats}
```

good point. we call these credential format profiles ;)

Sakurann (changes requested):
Please use "Credential Format Profiles"

bc-pi (approved):

c2bo (approved):

nemqe (approved):

Sakurann (approved):

PR #242, opened 2024-09-01
Add normative reference to JWK spec

### LINKED ISSUE #228: Add normative reference to JWK - RFC 7517
JWK keys and key sets are referenced in the specification but there is no formal reference to RFC 7517 (unlike OIDC4VCI and SIOPv2)

selfissued:
Fixes #228


David-Chadwick (commented):
Has 7517 also been added to the normative references section?

c2bo:
> Has 7517 also been added to the normative references section?

That should happen automatically with the syntax (@! should be normative and @ should be informative with mmark)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -881,7 +881,7 @@ A Verifier Attestation JWT MUST contain the following claims:
...
-* `cnf`: REQUIRED. This claim contains the confirmation method as defined in [@!RFC7800]. It MUST contain a JWK as defined in Section 3.2 of [RFC7800]. This claim determines the public key for which's corresponding private key the Verifier MUST proof possession of when presenting the Verifier Attestation JWT. This additional security measure allows the Verifier to obtain a Verifier Attestion JWT from a trusted issuer and use it for a long time independent of that issuer without the risk of an advisary impersonating the Verifier by replaying a captured attestation. 
+* `cnf`: REQUIRED. This claim contains the confirmation method as defined in [@!RFC7800]. It MUST contain a JWK [@!RFC7517] as defined in Section 3.2 of [RFC7800]. This claim determines the public key for which's corresponding private key the Verifier MUST proof possession of when presenting the Verifier Attestation JWT. This additional security measure allows the Verifier to obtain a Verifier Attestion JWT from a trusted issuer and use it for a long time independent of that issuer without the risk of an advisary impersonating the Verifier by replaying a captured attestation.
```
c2bo, line 884: Not sure about this, but felt more natural to me to use the long form when formally mentioning JWK for the first time.
```suggestion
* `cnf`: REQUIRED. This claim contains the confirmation method as defined in [@!RFC7800]. It MUST contain a JSON Web Key [@!RFC7517] as defined in Section 3.2 of [RFC7800]. This claim determines the public key for which's corresponding private key the Verifier MUST proof possession of when presenting the Verifier Attestation JWT. This additional security measure allows the Verifier to obtain a Verifier Attestion JWT from a trusted issuer and use it for a long time independent of that issuer without the risk of an advisary impersonating the Verifier by replaying a captured attestation.
```

c2bo (approved):
When reading the text I started wondering if our references for `client_metadata` need a bit of work though: `jwks` and  `jwks_uri` is used throughout the text but does not reference section 9 where client_metadata is formally introduced with the reference to DCR. I think we should reference section 9 at least once (which is only referenced for `vp_token` right now). I guess that should be a separate issue/PR?

Sakurann (changes requested):
let's fix other typos too if we are doing a PR on this section

Sakurann:
@c2bo 
> When reading the text I started wondering if our references for client_metadata need a bit of work though: jwks and jwks_uri is used throughout the text but does not reference section 9 where client_metadata is formally introduced with the reference to DCR. I think we should reference section 9 at least once (which is only referenced for vp_token right now). I guess that should be a separate issue/PR?

agree with proposed clarifications. a separate issue/PR would be good.

jogu (approved):

Sakurann (approved):

PR #243, opened 2024-09-01
Clarify signing and encryption of JARM responses

### LINKED ISSUE #216: [JARM] Additional clarifications about signed JWT, Nested JWT and encrypted JWT
Regarding the section [Signed and/or Encrypted Responses](https://github.com/openid/OpenID4VP/blob/main/openid-4-verifiable-presentations-1_0.md#signed-andor-encrypted-responses-jarm).

To provide concrete guidance for implementations, I suggest the following change:

`To sign, or sign and encrypt the Authorization Response, implementations MAY use JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM].`

to 

`To sign, encrypt, or both sign and encrypt the Authorization Response using a Nested JWT [RFC7519], implementations must utilize the JWT Secured Authorization Response Mode for OAuth 2.0. (JARM) [@!JARM].`

In addition to this, I would open a conversation about how a wallet is supposed to provide its public keys to the verifier for the signature validation, when the signed JWT or the Nested JWT is used. I suppose using wallet_metadata and or wallet instance attestation. we need to better clarify this if we agree

selfissued:
Fixes #216

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -706,9 +706,9 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
 
 ## Signed and/or Encrypted Responses {#jarm}
 
-This section defines how Authorization Response containing a VP Token can be signed and/or encrypted at the application level when the Response Type value is `vp_token` or `vp_token id_token`. Encrypting the Authorization Response can prevent personal data in the Authorization Response from leaking, when the Authorization Response is returned through the front channel (e.g., the browser).
+This section defines how an Authorization Response containing a VP Token can be signed and/or encrypted at the application level when the Response Type value is `vp_token` or `vp_token id_token`. Encrypting the Authorization Response can prevent personal data in the Authorization Response from leaking, when the Authorization Response is returned through the front channel (e.g., the browser).
 
-To sign, or both sign and encrypt the Authorization Response, implementations MUST use the JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM].
+To sign, encrypt, or both sign and encrypt the Authorization Response, implementations MUST use the JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM], and when only encrypting, the JARM extension described below.
 
 This specification also defines how to encrypt an unsigned Authorization Response by extending the mechanisms defined in [@!JARM]. The JSON containing the Authorization Response parameters can be encrypted as the payload of the JWE.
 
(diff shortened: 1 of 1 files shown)


peppelinux (approved):

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -708,7 +708,7 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
 
 This section defines how Authorization Response containing a VP Token can be signed and/or encrypted at the application level when the Response Type value is `vp_token` or `vp_token id_token`. Encrypting the Authorization Response can prevent personal data in the Authorization Response from leaking, when the Authorization Response is returned through the front channel (e.g., the browser).
 
-To sign, or sign and encrypt the Authorization Response, implementations MAY use JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM].
+To sign, or both sign and encrypt the Authorization Response as a Nested JWT [@!RFC7519], implementations MUST use the JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM].
```
awoie, line 711: If the authorization response might not be nested in case it is only encrypted, or only signed.
```suggestion
To sign, or both sign and encrypt the Authorization Response, implementations MUST use the JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM].
```

awoie (changes requested):
Looks good to me. I'll approve after my suggestion was accepted.

awoie (approved):
After reviewing again, I think the text is correct.

openid-4-verifiable-presentations-1_0.md
Sakurann: I agree with this suggested change.

Sakurann, line 711:
```suggestion
To sign, encrypt, or both sign and encrypt the Authorization Response, implementations MUST use the JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM].
```

without this, this PR does not solve the original issue. and nested jwt here feels redundant and confusing.

Sakurann (changes requested):
the PR does not solve the issue unless it explicitly adds an option only to encrypt

openid-4-verifiable-presentations-1_0.md
jogu: I think this sentence is actually getting more confusing. Kristina's suggestion now says you must use the extension in this section when you're just signing a response, which would not be correct.

I would leave the scope of this sentence at dealing with the signed or signed+encrypted cases only.

jogu (approved):

PR #246, opened 2024-09-04
fix: user agent term and client_id value in credential api request

peppelinux:
this PR:

- aligns the term "user agent"
- within the Credential API non-normative example about the Verifier's request, it fixes the `client_id` parameter value when the `client_id_scheme` parameter value is set with `entity_id` value.

AFAIK we don't have "User Agent" as a defined term, otherwise I would align it accordingly from `user agent` to `User Agent`.

ps = there is stil la place where we use the term "web browser". It is easy to read and understand but, formally, I am wondering if we should align it as well using "user agent".

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1526,7 +1526,7 @@ This OpenID4VP profile utilizes the mechanisms of the W3C Digital Credentials AP
 
 The Digital Credentials API offers several advantages for implementers of both Verifiers and Wallets. 
 
-Firstly, the API serves as a privacy-preserving alternative to invoking Wallets via URLs, particularly custom URL schemes. The underlying app platform will only invoke a Wallet if the End-User confirms the request based on contextual information about the Credential request and the requestor (Verifier). 
+Firstly, the API serves as a privacy-preserving alternative to invoking Wallets via URLs, particularly custom URL schemes. The underlying app platform will only invoke a Wallet if the End-User confirms the request based on contextual information about the Credential Request and the requestor (Verifier). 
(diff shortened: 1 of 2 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1621,7 +1621,7 @@ The signed request allows the Wallet to authenticate the Verifier using a trust
 
 ## Response
 
-Every OpenID4VP Authorization Request results in a response being provided through the W3C Digital Credentials API. The response is an instance of the `DigitalCredential` credential interface, as defined in [@!w3c.digital_credentials_api], and the OpenID4VP Authorization Response parameters as defined for the Response Type are represented as an object within the `data` attribute.
+Every OpenID4VP Authorization Request results in a response being provided through the W3C Digital Credentials API. The response is an instance of the `DigitalCredential` Credential interface, as defined in [@!w3c.digital_credentials_api], and the OpenID4VP Authorization Response parameters as defined for the Response Type are represented as an object within the `data` attribute.
```
peppelinux, line 1624:
```suggestion
Every OpenID4VP Authorization Request results in a response being provided through the W3C Digital Credentials API. The response is an instance of the `DigitalCredential` interface, as defined in [@!w3c.digital_credentials_api], and the OpenID4VP Authorization Response parameters as defined for the Response Type are represented as an object within the `data` attribute.
```

Sakurann (approved):

jogu (approved):

bc-pi (approved):

c2bo (approved):

selfissued (approved):

PR #250, opened 2024-09-06
response_mode was missing in browser API example unsigned request

### LINKED ISSUE #244: Browser API example unsigned request appears not to be conformant with the spec text
At https://openid.net/specs/openid-4-verifiable-presentations-1_0-21.html#appendix-A.2 the example does not include `response_mode` but the text a few paragraphs below this example appears to make omissions of `response_mode` illegal:

> The value of the response_mode parameter MUST be w3c_dc_api when the response is neither signed nor encrypted and w3c_dc_api.jwt when the response is signed and/or encrypted as defined in [Section 6.3](https://openid.net/specs/openid-4-verifiable-presentations-1_0-21.html#jarm).

I have a slight feeling that the intention was that `response_mode` could be omitted and would effectively default to `w3c_dc_api` for requests received via the API but I can't fully remember.

bc-pi:
Add response_mode: "is_w3c_dc_api" to browser API example unsigned request to fix #244

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1552,6 +1552,7 @@ try {
         protocol: "openid4vp",
         request:  {
           response_type: "vp_token",
+          response_mode: "is_w3c_dc_api",
           nonce: "n-0S6_WzA2Mj",
           client_metadata: {...},
           presentation_definition: {...}


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1552,6 +1552,7 @@ try {
         protocol: "openid4vp",
         request:  {
           response_type: "vp_token",
+          response_mode: "is_w3c_dc_api",
```
c2bo, line 1555: Perhaps I am missing something but where did the `is_` come from? 
```suggestion
          response_mode: "w3c_dc_api",
```

bc-pi: The only thing you're missing is a careless mistake on my part when copying content for the several different places one has to write something for a PR... Thanks for catching it!

jogu (approved):

c2bo (approved):

peppelinux (approved):

Issue #206, opened 2024-07-04
Incorporate browser API sequence diagrams into the specification


jogu:
As per https://github.com/openid/OpenID4VP/pull/155/files#r1653319017 there are some plantuml diagrams added in the browser API PR that are not incorporated into the specification.

We would like those diagrams to be incorporated into the specification.

To avoid synchronisation issues that should happen automatically when the specification is build (e.g. in the GitHub workflow: https://github.com/openid/OpenID4VP/blob/main/.github/workflows/Build-OID4VP.yml ).

It anyone is able to help us get that working it would be greatly appreciated.

jogu:
Someone on today's WG call (sorry, I have forgotten who!) suggested this tool might be part of a solution:

https://pypi.org/project/pandoc-imagine/

PR #254, opened 2024-09-10
Remove W3C DC API examples

marcoscaceres:
As the W3C spec is changing a lot, it's best yet to not include any examples. That way, this spec can progress without locking in changes on the W3C side.


jogu:
Are we expecting breaking changes to the Browser API, I thought it was fairly settled for now?

I'm not sure that removing the examples helps. The examples are non-normative and don't really show anything that's not already described in the normative text.

(We can actually update the examples very easily, both during any public review period for the 1.0 spec revision, and even after 1.0 is published by making an errata update, but if the normative text is likely to require changes as well that's a much bigger problem.)

jogu:
Ah, I guess https://github.com/WICG/digital-credentials/pull/165 may be the driver?

Although the 'providers' -> 'requests' change itself looks like it shouldn't break normative text in VP,  It is slightly concerning that there are suggestions on that PR that would affect the normative text, like changing 'request' to 'query'.

samuelgoto:
I'm concerned about this here: https://github.com/WICG/digital-credentials/pull/164

RByers:
Maybe leaving a link for where to see current examples would be better than just removing them? Eg. maybe they belong in this GitHub repo but in their own non-normative file for easier updating?

marcoscaceres:
I don't think we should lock in this spec with the other at all. We might want to move things around and we haven't yet added the examples to the Digital Credentials Spec. Rest assured that we will add examples, but it might not happen in the accelerated timeframe OpenID4VP is currently on. We can add examples to a future version of OpenID4VP once the W3C Spec is a) in a working group, and b) we are confident on the design (e.g. the spec reaches CR on the W3C side). We don't have enough implementation experience yet to be confident we won't change more things - specially as the W3C spec hasn't been road tested with more request formats, which may change the underlying request model/structure.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1537,34 +1537,11 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-The value of the `protocol` parameter of the W3C Digital Credentials API MUST be set to `openid4vp` for this profile.
+When used with the W3C Digital Credentials API [@!w3c.digital_credentials_api], the exchange protocol is the string `openid4vp` for this profile.
```
Sakurann, line 1540:
```suggestion
For the profile defined in this section, the value of the exchange protocol used with the W3C Digital Credentials API [@!w3c.digital_credentials_api], is the `openid4vp`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1537,34 +1537,11 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
-          presentation_definition: {...}
-        }
-      }]
-    }
-  });
-} catch (err) {
-  // Handle errors and/or fallback to other invocation mechanisms
-}
-```
+The W3C Digital Credentials API [@!w3c.digital_credentials_api] contains an OpenID4VP Authorization Request, where every OpenID4VP Authorization Request parameter is represented as a top-level JavaScript object member.
```
Sakurann, line 1544: > where every OpenID4VP Authorization Request parameter is represented as a top-level JavaScript object member

Is this a new development in Digital Credentials API? How would this work for signed requests?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1591,7 +1568,7 @@ Any OpenID4VP request compliant to this section of this specification can be use
 
 ### Unsigned Request {#unsigned_request}
 
-The Verifier MAY send all the OpenID4VP request parameters as members in the `request` member passed to the API. In this case, the Wallet will use the Verifier's origin as asserted by the user agent as the Verifer's Client Identifier.
+The Verifier MAY send all the OpenID4VP request parameters as members to the W3C Digital Credentials API [@!w3c.digital_credentials_api]. In this case, the Wallet will use the Verifier's origin as asserted by the user agent as the Verifer's Client Identifier.
```
Sakurann, line 1571: Again, how does this work for signed request where all the parameters are within a signed JWT which contains all of the parameters?

Sakurann (changes requested):

Sakurann:
Text in this PR does not sufficiently decouple openid4vp text and still makes it susceptible to any future breaking changes in the credentials API. 

"Put openid4vp request parameters in an appropriate place as defined in digital credentials API; the openid4vp parameter is X for signed requests and Y for unsigned requests" is what I would expect. 

Also very unsure about removing all of the examples. Will digital credentials API provide examples how to include openid4vp request there once it is stable? If not, we should keep some examples as of today in the spec. 

If the text is generic enough, changing the examples later as digital credentials API evolves is possible.

openid-4-verifiable-presentations-1_0.md
jogu: The change here is removing "`request` member", I think that's okay, it's removing the coupling between the two specs but it doesn't change how signed requests are handled.

openid-4-verifiable-presentations-1_0.md
jogu: This is in the "Unsigned request" section, this text won't affect signed requests.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1537,34 +1537,11 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
-          client_metadata: {...},
-          presentation_definition: {...}
-        }
-      }]
-    }
-  });
-} catch (err) {
-  // Handle errors and/or fallback to other invocation mechanisms
-}
-```
```
jogu, line 1567: I can't figure out how to actually make a suggestion that undoes the delete, but I believe we should NOT remove the actual OID4VP part of the examples. It was clear from last night's DCP WG call that many people don't understand what is passed by OID4VP to the browser API, and removing the examples will definitely make that even harder to figure out.

I hence suggest we retain this part of the example:

```
{
           response_type: "vp_token",
           response_mode: "w3c_dc_api",
           nonce: "n-0S6_WzA2Mj",
           client_metadata: {...},
           presentation_definition: {...}
         }
```

(and apply a similar change to other removed examples)

jogu (changes requested):

openid-4-verifiable-presentations-1_0.md
Sakurann: Strongly agree

openid-4-verifiable-presentations-1_0.md
Sakurann: so 
```suggestion
The Verifier MAY send all the OpenID4VP request parameters to the W3C Digital Credentials API as defined in [@!w3c.digital_credentials_api]. In this case, the Wallet will use the Verifier's origin as asserted by the user agent as the Verifer's Client Identifier.
```

openid-4-verifiable-presentations-1_0.md
jogu: Discussed on today's WG call, consensus to keeping the payload parts but removing the rest of the example.

openid-4-verifiable-presentations-1_0.md
Sakurann: is there any risk of this changing in the future? wouldn't this prevent passing multiple requests in an array? Would really prefer a text like `The Verifier MAY send all the OpenID4VP request parameters to the W3C Digital Credentials API as defined in [@!w3c.digital_credentials_api]`

marcoscaceres: It's best to say less. We don't know what the final API will look like.

openid-4-verifiable-presentations-1_0.md
marcoscaceres: Should be good now. @jogu, @Sakurann, can you confirm.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1593,36 +1593,28 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-The value of the `protocol` parameter of the W3C Digital Credentials API MUST be set to `openid4vp` for this profile.
+For the profile defined in this section, the value of the exchange protocol used with the W3C Digital Credentials API [@!w3c.digital_credentials_api], is the `openid4vp`.
```
jogu, line 1596: Not sure if there's a missing or extra word here but it doesn't read quite right:

```suggestion
For the profile defined in this section, the value of the exchange protocol used with the W3C Digital Credentials API [@!w3c.digital_credentials_api], is `openid4vp`.
```

jogu (commented):
Thanks Marcos! Some small comments, could you update the PR title/first comment to reflect what is changing now too please?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1593,36 +1593,28 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
-  // Handle errors and/or fallback to other invocation mechanisms
+{
+  protocol: "openid4vp",
+  request:  {
+    response_type: "vp_token",
+    response_mode: "w3c_dc_api",
+    nonce: "n-0S6_WzA2Mj",
+    client_metadata: {...},
+    presentation_definition: {...}
+  }
```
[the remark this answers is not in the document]
Sakurann: i think request is still there? https://wicg.github.io/digital-credentials/#the-digitalcredentialsrequest-dictionary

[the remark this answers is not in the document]
jogu: Are we looking at the same thing - it looks like it was renamed to `data` when I look at the url you shared:

Screenshot 2024-10-10 at 15 46 21

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1661,7 +1651,7 @@ The signed Request Object MAY contain all the parameters listed in (#browser_api
 Below is a non-normative example of such a request:
 
 ```js
-const credential = await navigator.identity.get({
+{
   digital: {
     providers: [{
```
samuelgoto:
```suggestion
    requests: [{
```

samuelgoto (approved):
LGTM++

This would make it easier to decouple the OpenID4VP spec and the Digital Credentials API spec.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1593,36 +1593,25 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-The value of the `protocol` parameter of the W3C Digital Credentials API MUST be set to `openid4vp` for this profile.
+For the profile defined in this section, the value of the exchange protocol used with the W3C Digital Credentials API [@!w3c.digital_credentials_api], is `openid4vp`.
 
 ## Request {#browser_api_request}
 
-The `request` member of the W3C Digital Credentials API [@!w3c.digital_credentials_api] contains an OpenID4VP Authorization Request, where every OpenID4VP Authorization Request parameter is represented as a top-level JavaScript object member.
+The W3C Digital Credentials API [@!w3c.digital_credentials_api] contains an OpenID4VP Authorization Request, where every OpenID4VP Authorization Request parameter is represented as a top-level JavaScript object member.
```
marcoscaceres, line 1600:
```suggestion
The Verifier MAY send all the OpenID4VP request parameters to the W3C Digital Credentials API as defined in [@!w3c.digital_credentials_api].
```

marcoscaceres:
Hopefully better now... did this in a hurry at the ARF event last week.

jogu (approved):

OR13 (approved):

diagrams/request_uri_mode_post_through_browser_api.md
```diff
@@ -16,8 +16,6 @@ participant "Wallet" as w
 u --> r : use
 activate r
```
peppelinux, line 17:
```suggestion
```

since the r -> wp is removed.

diagrams/request_uri_mode_post_through_browser_api.md
```diff
@@ -16,8 +16,6 @@ participant "Wallet" as w
...
 
-r -> wp: navigator.identity.get(\nprotocol="openid4vp",\nrequest="client_id, \nrequest_uri, request_uri_method=post, \npresentation_definition")
-
 deactivate r
```
peppelinux, line 19:
```suggestion
```

diagrams/request_uri_mode_post_through_browser_api.plantuml
```diff
@@ -16,8 +16,6 @@ participant "Wallet" as w
 u --> r : use
 activate r
```
peppelinux, line 17:
```suggestion
```

diagrams/request_uri_mode_post_through_browser_api.plantuml
```diff
@@ -16,8 +16,6 @@ participant "Wallet" as w
...
 
-r -> wp: navigator.identity.get(\nprotocol="openid4vp",\nrequest="client_id, \nrequest_uri, request_uri_method=post, \npresentation_definition")
-
 deactivate r
```
peppelinux, line 19:
```suggestion
```

diagrams/request_uri_mode_post_through_browser_api.md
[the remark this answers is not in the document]
jogu: See https://github.com/openid/OpenID4VP/issues/206

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1593,36 +1593,25 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
-The value of the `protocol` parameter of the W3C Digital Credentials API MUST be set to `openid4vp` for this profile.
+For the profile defined in this section, the value of the exchange protocol used with the W3C Digital Credentials API [@!w3c.digital_credentials_api], is `openid4vp`.
 
 ## Request {#browser_api_request}
 
-The `request` member of the W3C Digital Credentials API [@!w3c.digital_credentials_api] contains an OpenID4VP Authorization Request, where every OpenID4VP Authorization Request parameter is represented as a top-level JavaScript object member.
+The Verifier MAY send all the OpenID4VP request parameters to the W3C Digital Credentials API as defined in [@!w3c.digital_credentials_api].
 
-The following is a non-normative example of how the W3C Digital Credentials API can be used with an unsigned OpenID4VP request when advanced security features of OpenID4VP are not used:
+The following is a non-normative example of an unsigned OpenID4VP request when advanced security features of OpenID4VP are not used:
```
Sakurann, line 1602:
```suggestion
The following is a non-normative example of an unsigned OpenID4VP request sent over  the W3C Digital Credentials API when advanced security features of OpenID4VP are not used:
```

Sakurann, line 1602:
```suggestion
The following is a non-normative example of an unsigned OpenID4VP request (when advanced security features of OpenID4VP are not used) that can be sent over the W3C Digital Credentials API :
```

diagrams/request_uri_mode_post_through_browser_api.md
```diff
@@ -14,11 +14,7 @@ participant "App Platform" as ap
 participant "Wallet" as w
 
 u --> r : use
-activate r
-
-r -> wp: navigator.identity.get(\nprotocol="openid4vp",\nrequest="client_id, \nrequest_uri, request_uri_method=post, \npresentation_definition")
```
Sakurann, line 19: so isn't this better be something like

> r -> wp: send OpenID4VP request over the W3C Digital Credentials API

diagrams/request_uri_mode_post_through_browser_api.plantuml
```diff
@@ -14,11 +14,7 @@ participant "App Platform" as ap
 participant "Wallet" as w
 
 u --> r : use
-activate r
-
-r -> wp: navigator.identity.get(\nprotocol="openid4vp",\nrequest="client_id, \nrequest_uri, request_uri_method=post, \npresentation_definition")
```
Sakurann, line 19: same as above. shouldn't this better be something like

> r -> wp: send OpenID4VP request over the W3C Digital Credentials API

diagrams/signed_request_uri_through_browser_api.plantuml
```diff
@@ -19,7 +19,6 @@ activate r
 note over r,wp
     Note that the signed request object contains the Verifier's origin.
 end note
-r -> wp: navigator.identity.get(\nprotocol="openid4vp",\nrequest="client_id, request")
```
Sakurann, line 22: same as above. shouldn't this better be something like

> r -> wp: send OpenID4VP request over the W3C Digital Credentials API

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1661,16 +1650,7 @@ The signed Request Object MAY contain all the parameters listed in (#browser_api
 Below is a non-normative example of such a request:
```
Sakurann, line 1650:
```suggestion
Below is a non-normative example of such a request sent over the W3C Digital Credentials API:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1593,36 +1593,25 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
-    }
-  });
-} catch (err) {
-  // Handle errors and/or fallback to other invocation mechanisms
+{
+  response_type: "vp_token",
+  response_mode: "w3c_dc_api",
+  nonce: "n-0S6_WzA2Mj",
+  client_metadata: {...},
+  presentation_definition: {...}
```
bc-pi, line 1610:
```suggestion
  vc_query: {...}
```

Sakurann (approved):
approved per WG discussion yesterday

PR #240, opened 2024-08-31
Enable non-breaking extensibility

selfissued:
Fixes #227

PR #258, opened 2024-09-11
Permit the use of the new query language instead of presentation exchange.

### LINKED ISSUE #255: Permit the use of the new query language instead of presentation exchange.
Section 5 and section 6 currently require the use of the presentation exchange specification in both the request and the response. To support implementations to use the new query language instead of the presentation exchange these requirements should be changed to optional.

Specifically:

Section 5 requires that `presentation_definition` or `presentation_definition_uri` MUST be present. This should be changed to MAY.
Section 6 requires that `presentation_submission` MUST be present. This should be changed to MAY.

martijnharing:
Remove requirements to use the presentation exchange specification. Resolves #255


openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,10 +252,10 @@ This specification enables the Verifier to send both Presentation Definition JSO
 This specification defines the following new parameters:
 
 `presentation_definition`:
-: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. This parameter MUST be present when `presentation_definition_uri` parameter, or a `scope` value representing a Presentation Definition is not present.
+: OPTIONAL. A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. 
 
 `presentation_definition_uri`:
-: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is not present. See (#request_presentation_definition_uri) for more details.
+: OPTIONAL. A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. See (#request_presentation_definition_uri) for more details.
```
Sakurann, line 258:
```suggestion
: OPTIONAL. A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST NOT be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is present. See (#request_presentation_definition_uri) for more details.
```

Sakurann (changes requested):
looks good. will approve once my suggestions are incorporated, since we do not want presentation_defintion and presentation_definition_uri (or scope) both being present in the same request

peppelinux (approved):

David-Chadwick (changes requested):
I do not think this is the best way to proceed in terms of extensibility. Rather, there should be two parameters, type of query followed by the query itself. In this way we can have
 type=presentation_definition_url  and the query can be the URL, 
type= presentation_definition and the query can be the PD,
type=oid_query_language and the query can be the new query language we are defining
This also allows other query languages to be added in future should that be necessary (and history has proved it will be, given the current evolution of PEv1, PEv1.1, PEv2, PEv2.1, OIDquery language etc.

c2bo:
> @Sakurann Adding the type parameter later on will be a breaking change. Adding it now will ensure extensibility in a non-breaking way for 1.1, 2, 2.1, 3 etc ad infinitum

In principle I agree that adding a query language type parameter would be beneficial IMHO.

Do you think it makes sense to introduce the query type parameter with the new query language PR? We can also add it as an optional parameter and defaulting to presentation exchange if not present - that way it would be a simple extension without any breaking changes.

c2bo:
We should probably also adjust these lines:

244
> The Verifier articulates requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. Wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].

 547
> A VP Token is only returned if the corresponding Authorization Request contained a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a `scope` parameter representing a Presentation Definition (#vp_token_request).

Apart from that looking good imho.

jogu (approved):

jogu:
I think a type parameter is a separate discussion, and this PR also leaves the door open to adding a type parameter in the future (i.e. with language like 'if type isn't present look for presentation_definition parameter for backwards compatibility').

I think I am missing what the key advantage of a type parameter is over (if we really need to do it and I'm really hoping we don't) having a `oid_query_language` parameter and a `oid_query_language_v2` parameter. And the latter actually seems more flexible to me as it leaves open the option that the verifier can use two query languages at the same time if there's a reason to (perhaps during a transition period where it doesn't know which one a wallet will support).

openid-4-verifiable-presentations-1_0.md
```diff
@@ -572,11 +572,11 @@ When a VP Token is returned, the respective response MUST include the following
...
+: OPTIONAL. The `presentation_submission` element as defined in [@!DIF.PresentationExchange]. It contains mappings between the requested Verifiable Credentials and where to find them within the returned VP Token. This is expressed via elements in the `descriptor_map` array, known as Input Descriptor Mapping Objects. These objects contain a field called `path`, which, for this specification, MUST have the value `$` (top level root path) when only one Verifiable Presentation is contained in the VP Token, and MUST have the value `$[n]` (indexed path from root) when there are multiple Verifiable Presentations, where `n` is the index to select. Additional parameters can be defined by Credential Formats, see (#alternative_credential_formats) for details.
```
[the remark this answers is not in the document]
Sakurann: great point - agreed

Sakurann:
also agree that `type` parameter discussion requires a separate issue

David-Chadwick:
> I think I am missing what the key advantage of a type parameter is 

It provides extensibility and backwards compatibility for ever. As requirements change, it automatically meets them. It is a well known and well recognised way of adding extensibility.

David-Chadwick:
> having a `oid_query_language` parameter and a `oid_query_language_v2` parameter. 

How does that cater for `oid_query_language_v3` which there is sure to be, given the history of the last two years?
And if the query parameter is made a set of type(s) and value(s), then you can have any number of query languages in the request

jogu:
> > having a `oid_query_language` parameter and a `oid_query_language_v2` parameter.
> 
> How does that cater for `oid_query_language_v3` which there is sure to be, given the history of the last two years? And if the query parameter is made a set of type(s) and value(s), then you can have any number of query languages in the request

You can add a `oid_query_language_v3` when it's needed. For clarity, this also works for v4, v5, v6, v7 and every version beyond that. https://github.com/openid/OpenID4VP/pull/240 clarifies that adding new parameters can be done at anytime (although this particular case was already possible as ignoring unknown parameters is the standard OAuth behaviour).

I'm not supportive of changing the query language parameter to be an array. It's adding unnecessary complexity.

But I think we're off topic for this PR, anyway.

jogu:
I agree with @c2bo that we need to do this:

> We should probably also adjust these lines:
> 
> 244
> 
> > The Verifier articulates requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. Wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
> 
> 547
> 
> > A VP Token is only returned if the corresponding Authorization Request contained a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a `scope` parameter representing a Presentation Definition (#vp_token_request).
> 
> Apart from that looking good imho.

Sakurann (approved):
assuming @c2bo 's suggestions will also be incorporated

awoie (approved):

David-Chadwick (changes requested):
This change does not say what should be present if presentation_definition and presentation_definition_uri and scope are missing. The current text allows all to be missing. A must better solution is to have a type, followed by one of these three, and then in future a new type can be added

jogu:
> This change does not say what should be present if presentation_definition and presentation_definition_uri and scope are missing. The current text allows all to be missing. A must better solution is to have a type, followed by one of these three, and then in future a new type can be added

I don't think this is a good idea. Particularly not once it was suggested that means having both type and the query as arrays and defining some new rules around that.

We can instead change the language to 'If neither presentation_definition nor presentation_definition_uri nor a scope indicating a credential to be used nor any other parameter indicating what credential is being requested'. That's all that would be in scope for this PR. This language would allow 'type' to be added in the future if the working group decided to. This PR is just about giving us options later.

Sakurann:
I am ok with `'If neither `presentation_definition`, `presentation_definition_uri`, a scope indicating a credential to be used or any other parameter indicating what credential is being requested **defined by this specification/WG**`.
I think it is important that we are not opening a rabbithole but leaving an option for this WG/specification to define such 4th parameter

openid-4-verifiable-presentations-1_0.md
```diff
@@ -572,11 +572,11 @@ When a VP Token is returned, the respective response MUST include the following
...
+: OPTIONAL. The `presentation_submission` element as defined in [@!DIF.PresentationExchange]. It MUST be present if the `presentation_definition` or `presentation_definition_uri` parameters were present in the request, or if a `scope` value representing a presentation_definition was used. It contains mappings between the requested Verifiable Credentials and where to find them within the returned VP Token. This is expressed via elements in the `descriptor_map` array, known as Input Descriptor Mapping Objects. These objects contain a field called `path`, which, for this specification, MUST have the value `$` (top level root path) when only one Verifiable Presentation is contained in the VP Token, and MUST have the value `$[n]` (indexed path from root) when there are multiple Verifiable Presentations, where `n` is the index to select. Additional parameters can be defined by Credential Formats, see (#alternative_credential_formats) for details.
```
Sakurann, line 575:
```suggestion
: OPTIONAL. The `presentation_submission` element as defined in [@!DIF.PresentationExchange]. It MUST be present if the `presentation_definition` parameter, `presentation_definition_uri` parameter, a `scope` parameter with a value representing a Presentation Definition, or any other parameter indicating what credential is being requested defined by this specification was used. It contains mappings between the requested Verifiable Credentials and where to find them within the returned VP Token. This is expressed via elements in the `descriptor_map` array, known as Input Descriptor Mapping Objects. These objects contain a field called `path`, which, for this specification, MUST have the value `$` (top level root path) when only one Verifiable Presentation is contained in the VP Token, and MUST have the value `$[n]` (indexed path from root) when there are multiple Verifiable Presentations, where `n` is the index to select. Additional parameters can be defined by Credential Formats, see (#alternative_credential_formats) for details.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,10 +252,10 @@ This specification enables the Verifier to send both Presentation Definition JSO
...
+: OPTIONAL. A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details. This parameter MUST NOT be present when `presentation_definition_uri` parameter, or a `scope` value representing a Presentation Definition is present.
 
 `presentation_definition_uri`:
-: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is not present. See (#request_presentation_definition_uri) for more details.
+: OPTIONAL. A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST NOT be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is present. See (#request_presentation_definition_uri) for more details.
```
Sakurann, line 258:
```suggestion
: OPTIONAL. A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST NOT be present if any of the following parameters are present: a `presentation_definition` parameter, a `scope` parameter with a value representing a Presentation Definition, or any other parameter defined by this specification that indicates what credential is being requested. See (#request_presentation_definition_uri) for more details.
```

Sakurann:
@David-Chadwick could you please take a look at my suggestions?

tplooker (approved):

David-Chadwick:
@Sakurann This new wording is a definite improvement. But as I understand it, any optional parameters can always be omitted and so any protocol messages without them are conformant to the spec. So a message without any of these optional parameters is conformant to the protocol, but such a message is useless as it does not tell the wallet which credentials are needed. Thus in my opinion the specification is broken if an implementation can create a conformant message that is useless.
So what still needs to be added to the text is a sentence along the lines of "the request message must contain at least one indication of which credentials need to be provided by the wallet"

openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,10 +252,10 @@ This specification enables the Verifier to send both Presentation Definition JSO
...
```
```suggestion

One of the following parameters MUST be present to indicate what credential is being requested: a `presentation_definition` parameter, `presentation_definition_uri` parameter, a `scope` parameter with a value representing a Presentation Definition, or any other parameter defined by this specification for that purpose.

```

Sakurann:
@David-Chadwick I tried bring clarity along the lines you suggest here: https://github.com/openid/OpenID4VP/pull/258/files#r1766818286

David-Chadwick (approved):
Thanks, this has fixed it.

c2bo (changes requested):
I'll repeat with a change request so we don't forget this - IMHO we need to adjust these lines:

244

> The Verifier articulates requirements of the Credential(s) that are requested using presentation_definition and presentation_definition_uri parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. Wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].

547

> A VP Token is only returned if the corresponding Authorization Request contained a presentation_definition parameter, a presentation_definition_uri parameter, or a scope parameter representing a Presentation Definition (#vp_token_request).

martijnharing:
> I'll repeat with a change request so we don't forget this - IMHO we need to adjust these lines:
> 
> 244
> 
> > The Verifier articulates requirements of the Credential(s) that are requested using presentation_definition and presentation_definition_uri parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. Wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
> 
> 547
> 
> > A VP Token is only returned if the corresponding Authorization Request contained a presentation_definition parameter, a presentation_definition_uri parameter, or a scope parameter representing a Presentation Definition (#vp_token_request).

I changed those two lines to be in compliance with the others changes. Do these changes work for you?

openid-4-verifiable-presentations-1_0.md
martijnharing: I think the exact text proposed would prohibit us from introducing a new query language in a future specification without being backwards incompatible. 
The current specification draft does not contain "any other parameter by this specification for that purpose". If this text would be the 1.0 version it means that in order to be compliant with 1.0 you have to use one of the three mechanisms described. 
So even if a future version of the mechanism does describe a new parameter, like vp_query, actually using that would mean you would not be compliant with this proposed language. And therefore using vp_query would not be backwards compatible with 1.0.

It is important that there is always a request parameter present, so we could consider adding the following:
"One of the following parameters MUST be present to indicate what credential is being requested: a `presentation_definition` parameter, `presentation_definition_uri` parameter, a `scope` parameter with a value representing a Presentation Definition, or any other parameter for that purpose."

openid-4-verifiable-presentations-1_0.md
```diff
@@ -241,7 +241,7 @@ that signals to the Wallet that it can make an HTTP POST request to the Verifier
...
+The Verifier MAY articulate requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. When processing this Presention Definiotn object, wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
```
c2bo, line 244:
```suggestion
The Verifier MAY articulate requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. When processing this Presentation Definition object, wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
```

c2bo (approved):
Wording for 244/255/257 feels a bit weird to me, but I couldn't come up with a version that felt better. I do believe the current text should work well to allow extensibility for the query language.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -572,11 +572,11 @@ When a VP Token is returned, the respective response MUST include the following
 : REQUIRED. JSON String or JSON object that MUST contain a single Verifiable Presentation or an array of JSON Strings and JSON objects each of them containing a Verifiable Presentations. Each Verifiable Presentation MUST be represented as a JSON string (that is a Base64url encoded value) or a JSON object depending on a format as defined in Appendix A of [@!OpenID.VCI].  When a single Verifiable Presentation is returned, the array syntax MUST NOT be used.  If Appendix A of [@!OpenID.VCI] defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the `vp_token` response parameter. Otherwise, this specification does not require any additional encoding when a Credential format is already represented as a JSON object or a JSON string.
```
danielfett, line 572: Here, vp_token is still required, above it is optional.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -544,7 +544,7 @@ If the Verifier responds with any HTTP error response, the Wallet MUST terminate
 
 # Response {#response}
```
danielfett, line 545: It feels like this whole section is not consistent (i.e., would need more changes) if we want to make VP Token optional.

openid-4-verifiable-presentations-1_0.md
Sakurann: `any other parameter defined by this specification for that purpose.` is what intends to allow us to introduce a new query language in a future specification without breaking changes.

Sakurann (changes requested):
i think an incorrect statement got introduced. i am very unsure about vp token text and waiting for my other suggestion to be resolved

openid-4-verifiable-presentations-1_0.md
```diff
@@ -572,11 +572,11 @@ When a VP Token is returned, the respective response MUST include the following
...
 
 Other parameters, such as `state` or `code` (from [@!RFC6749]), or `id_token` (from [@!OpenID.Core]), and `iss` (from [@RFC9207]) can be included in the response as defined in the respective specifications. `state` values MUST only contain ASCII URL safe characters (uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde). For the implementation considerations of a `state` parameter, see (#state_management).
 
-The `presentation_submission` element MUST be included as a separate response parameter alongside the VP token. Clients MUST ignore any `presentation_submission` element included inside a Verifiable Presentation.
+If present, the `presentation_submission` element MUST be included as a separate response parameter alongside the VP token. Clients MUST ignore any `presentation_submission` element included inside a Verifiable Presentation.
```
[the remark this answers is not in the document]
Sakurann: actually, @javereec your suggestion changes the meaning and is not correct because when `presentation_submission` is used, it has to be included in the VP Token. @martijnharing's original text is correct.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -572,11 +572,11 @@ When a VP Token is returned, the respective response MUST include the following
...
 
 Other parameters, such as `state` or `code` (from [@!RFC6749]), or `id_token` (from [@!OpenID.Core]), and `iss` (from [@RFC9207]) can be included in the response as defined in the respective specifications. `state` values MUST only contain ASCII URL safe characters (uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde). For the implementation considerations of a `state` parameter, see (#state_management).
 
-The `presentation_submission` element MUST be included as a separate response parameter alongside the VP token. Clients MUST ignore any `presentation_submission` element included inside a Verifiable Presentation.
+The `presentation_submission` element MAY be included as a separate response parameter alongside the VP token. Clients MUST ignore any `presentation_submission` element included inside a Verifiable Presentation.
```
Sakurann, line 579:
```suggestion
If present, the `presentation_submission` element MUST be included as a separate response parameter alongside the VP token.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: umm.. in the new query language, vp_token would still be present, just it's structure would change, no? PR #266 seems to be silent on the response?

openid-4-verifiable-presentations-1_0.md
Sakurann: where can i find how the proposal on how the response with a new query language will look like? without that it's very hard for me to make sense of the changes proposed in the vp token section.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -241,7 +241,7 @@ that signals to the Wallet that it can make an HTTP POST request to the Verifier
...
+The Verifier MAY articulate requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters that contain a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. When processing this Presentation Definition object, wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
```
Sakurann, line 244:
```suggestion
The requirements of the Credential(s) that are requested using `presentation_definition` and `presentation_definition_uri` parameters are articulated by the Verified using a Presentation Definition JSON object as defined in Section 5 of [@!DIF.PresentationExchange]. When processing this Presentation Definition object, wallet implementations MUST process Presentation Definition JSON object and select candidate Verifiable Credential(s) using the evaluation process described in Section 8 of [@!DIF.PresentationExchange] unless implementing only a credential profile that provides rules on how to evaluate and process [@!DIF.PresentationExchange].
```

I am highly uncomfortable with a `MAY` here.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -544,7 +544,7 @@ If the Verifier responds with any HTTP error response, the Wallet MUST terminate
 
 # Response {#response}
 
-A VP Token is only returned if the corresponding Authorization Request contained a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a `scope` parameter representing a Presentation Definition (#vp_token_request).
+A VP Token MAY be returned if the corresponding Authorization Request contained a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a `scope` parameter representing a Presentation Definition (#vp_token_request).
```
Sakurann, line 547:
```suggestion
A VP Token MUST be returned when the corresponding Authorization Request contained a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a `scope` parameter representing a Presentation Definition (#vp_token_request).
```
this is still a MUST. and honestly, re-reading it, the original wording seems right to me - it says return VP Token when the things asking for it are used. @c2bo - what was wrong in your opinion in the original text?

openid-4-verifiable-presentations-1_0.md
jogu: See https://github.com/openid/OpenID4VP/issues/178 and the vp token examples, e.g.:

Screenshot 2024-09-25 at 09 22 34

openid-4-verifiable-presentations-1_0.md
David-Chadwick: @Sakurann The problem that @martijnharing has identified, and that I agree with, is the wording "by this specification". This means v1.0 of the specification, and only v1.0 when it is the only published specification. Thus implementors will code for 3 alternative parameters only, since v1.0 only defines 3 parameters. When a later specification comes along that specifies a 4th parameter, v1.0 implementations will be incompatible with this and will say "protocol error" when the 4th parameter is encountered (or may crash!), since from its perspective there are only 3 standardised values. What @martijnharing's proposed edit does is ensures that v1.0 implementations will also code up for any unknown parameter (in v1.0) that might be defined in the future and so should gracefully handle this by replying "<future value> query not supported".

martijnharing: David clarified the issue. I'll try to be more precise:
The issue with the proposed text comes only from it not allowing us to add vp_query in 1.1. If that's not the case then there is no issue, but the reasoning is:

Assumption 1: A change is considered a breaking change if it allows an implementation to be compliant with 1.1 while that implementation is not compliant with 1.0
Assumption 2: It's not allowed to make changes to v1.1 that are considered a breaking change.

Since "this specification" means "OpenID4VP v1.0" in the context of v1.0, the sentence "or any other parameter defined by this specification for that purpose" doesn't add any options because the 3 options listed in that paragraph are the only options defined in OpenID4VP v1.0. Therefore an implementation of a RP according to v1.0 must use one of these 3 options.

If we add the vp_query element to v1.1, then for an RP implementation, using vp_query instead of one of the other 3 options is allowed, since "or any other parameter defined by this specification for that purpose" translates to "or vp_query" in the context of v1.1. 

In this example, this means that: 
- v1.0 effectively says that an RP must use one of 3 options and using anything else is not allowed.
- v1.1 effectively says that an RP must use one of 4 options and using anything else in not allowed.

Since an implementation of an RP under v1.1 using the fourth option (vp_query) is compliant with v1.1 but not compliant with v1.0, actually adding vp_query to v1.1 would be considered a breaking change under assumption 1. 
Therefore under assumption 2 adding vp_query to v1.1 would not be allowed.

If assumption 1 or assumption 2 is not correct (and I happily stand corrected) the concern could very well not apply.

openid-4-verifiable-presentations-1_0.md
c2bo: > A VP Token is only returned if the corresponding Authorization Request contained a `presentation_definition` parameter, a `presentation_definition_uri` parameter, or a `scope` parameter representing a Presentation Definition (#vp_token_request).

We will return a VP Token whenever a request containing some kind of query was provided. In terms of extensibility of query language, I would understand the old text as we will not return a VP Token unless there was a Presentation Definition - which is not what we want with the new query language.

jogu: The new query language proposal in https://github.com/openid/OpenID4VP/pull/266/files still returns a VP token, so I think the change actually needed here is the same one as being refined above, i.e. adding something about "or any other parameter defined by this specification that indicates what credential is being requested".

Sakurann: yeah, then `is only` is totally fine.

openid-4-verifiable-presentations-1_0.md
jogu: Discussed on today's call, change to "defined by this specification (or any future versions of this specification)"

openid-4-verifiable-presentations-1_0.md
danielfett: When present where?

openid-4-verifiable-presentations-1_0.md
danielfett: And this definition MUST be relaxed as well when we want to allow a different structure with a new query language.

David-Chadwick:
and all its future versions

i.e. plural

openid-4-verifiable-presentations-1_0.md
Sakurann: @martijnharing could you please accept?

openid-4-verifiable-presentations-1_0.md
Sakurann: where it needs to be according to the specification

openid-4-verifiable-presentations-1_0.md
martijnharing: Added the language as was discussed in the call to the PR. I do think it is awkward that compliance to this specification now becomes dependent on the publication schedule of future documents. e.g. test tooling for 1.0 can never be finalized since it is always possible that it must be updated when a new version of this specification becomes available.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -252,10 +252,12 @@ This specification enables the Verifier to send both Presentation Definition JSO
...
-: A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST be present when `presentation_definition` parameter, or a `scope` value representing a Presentation Definition is not present. See (#request_presentation_definition_uri) for more details.
+: OPTIONAL. A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. This parameter MUST NOT be present if any of the following parameters are present: a `presentation_definition` parameter, a `scope` parameter with a value representing a Presentation Definition, or any other parameter defined by this specification that indicates what credential is being requested. See (#request_presentation_definition_uri) for more details.
+
+One of the following parameters MUST be present to indicate what credential is being requested: a `presentation_definition` parameter, `presentation_definition_uri` parameter, a `scope` parameter with a value representing a Presentation Definition, or any other parameter defined for that purpose by this specification and all its future versions.
```
danielfett, line 260: What would be a "parameter defined for that purpose defined by this specification"?

danielfett, line 260: "all its future versions." means that the parameter has to be defined by all future versions. That's certainly not intended.

danielfett:
To elaborate what I said on the call, the following should be our target:

From the view of a verifier, 
* when I implement 1.0, the spec forces me to use PE; I cannot use any other query language, as it is not allowed by the spec.
* when I implement 1.1, I have more options (new query language); (whether I can use one or both is up to discussion)

From the view of a wallet,
* when I implement 1.0, I only know PE; there is no other option I would understand; I somebody sends a parameter with a query in a new query language, I'll ignore it, as I don't understand it.
* when I implement 1.1, I know more options: I can allow requests in PE and the new query language.

To achieve this, the current spec without this PR is fine.
For 1.1, we can decide if PE needs to be supported by every wallet and what the options for verifiers are. We don't need to decide that now.

David-Chadwick:
> From the view of a wallet,
> 
> when I implement 1.0, ... if somebody sends a parameter with a query in a new query language, I'll ignore it, as I don't understand it.

And then I have a request with no query in it. So what reply do I send to the RP? Protocol error?

Ignoring it is the wrong behaviour. Rather the behaviour we want is:  if somebody sends a parameter with a query in a new query language, I will know this has been standardised in a future version of the standard that I have not yet implemented, so I will reply "query language not implemented" or similar

tplooker:
> To elaborate what I said on the call, the following should be our target:

Appreciate the write up as I think its important to describe exactly what will be normatively required here in v1.0 vs v1.1. I think the most important question to answer is could a v1.1 RP (verifier) implementation and wallet use only the new query syntax if they wanted to OR would an RP effectively have to have the credential query repeated twice in the request one in the new format and one in the old (P.E)?

IMO if we take the position in 1.0 that the query syntax supported by both implementations is negotiate-able rather then fixed then we could allow v1.1 implementations to only use P.E if they so wished, without it being a breaking change. Otherwise IMO we are stuck in a situation where a P.E based query has to be in the request of every v1.x to remain backward compatible.

tlodderstedt:
Negotiation requires the POST request based approach, right? How should that work for the DC API? Is the RP supposed to send two request objects at once?

javereec:
I was wondering if we also should relax the link of `scope` to Presentation Definition. The current text reads

> ## Using `scope` Parameter to Request Verifiable Credential(s) {#request_scope}
> 
> Wallets MAY support requesting presentation of Verifiable Credentials using OAuth 2.0 scope values.
> 
> Such a scope value MUST be an alias for a well-defined Presentation Definition that will be referred to in the `presentation_submission` response parameter. 

In the future it could reference the new query language

Sakurann:
The way the discussion has gone, 
- if PE is the only query language in 1.0 (ie mandatory in 1.0), there is no consensus that PE is mandatory in 1.1, too.
- If the above is true, this PR kind of intentionally introduces a backdoor/hole in the spec to introduce new query language in 1.1 . in other words 1.1 is still a breaking change to 1.0. and some are not comfortable with it.
- so the new direction seems to be to include both PE and new query language (NQL) in the implementers draft published before the end of 2024 (PE mandatory, NQL experimental), and later discuss what is mandatory and optional in 1.0.
- if the above is true, this PR should be closed and #266 is merged before ID-3 is cut.

David-Chadwick:
> if PE is the only query language in 1.0 (ie mandatory in 1.0)....
introduce new query language in 1.1 .in other words 1.1 is still a breaking change to 1.0.,

I do not think that these two statements logically follow. PE is the only query language in 1.0, this is true. But in other versions there are other languages. So if a 1.0 implementation knows this, which the 1.0 specification can make clear, then if it receives another query language it knows it is not talking to a 1.0 implementation. But this is not a breaking change, because a 1.0 implementation can code for this now.

Sakurann:
superseded by #266

Sakurann:
> I do not think that these two statements logically follow. PE is the only query language in 1.0, this is true. But in other versions there are other languages. So if a 1.0 implementation knows this, which the 1.0 specification can make clear, then if it receives another query language it knows it is not talking to a 1.0 implementation. But this is not a breaking change, because a 1.0 implementation can code for this now.

If PE is not mandatory in versions after 1.1, an implementation that built following 1.1 will not be able to interoperate with implementation following 1.0...

Sakurann:
superseded by https://github.com/openid/OpenID4VP/pull/266

David-Chadwick:
`If PE is not mandatory in versions after 1.1, an implementation that built following 1.1 will not be able to interoperate with implementation following 1.0...`

It depends what you mean by interoperate. If you mean, talk to each other and know how to respond with a conforming response, then it is not a breaking change. If you mean provide a set of credentials in response, then it is a breaking change. But there are many cases of the latter in the existing specification such as: support for different DIDs, different cryptographic functions, different credential schemas/formats. So support for different query formats is just one of many "breaking changes" that already exist.

Issue #248, opened 2024-09-06
How can verifiers that support multiple trust models/ecosystems know how to authenticate to the wallet?


jogu:
This is split out from https://github.com/openid/oid4vc-haip-sd-jwt-vc/issues/133 :

> For item 3 "what if verifier wants to pass multiple trust models, hoping one is supported by the wallet?"
> 
> This is implicitly mentioned in the WG10 ask since the 18013-5 request structure supports multiple RP authentication signatures.
> 
> Not having the ability to support multiple RP authentication signatures would mean that it's not possible to request documents that use different trust models, or request documents from issuers (e.g. regions) that have different trust models. Not having those abilities would be a significant limitation of the protocol.
> 
> Most of the consequences and possible solutions are similar to the ones mentioned in item 1, since having multiple RP authentication signatures seems to not be supported by the JAR specification.
> 
> The proposal addresses this by defining an array of verifier_authentication elements, each of which can contain a signature for a trust model. As the elements that are signed by these structures is dynamic (see item 1), this mechanism supports both the scenario of signing the encryption information, as well as also signing the vp_query.

This doesn't seem to be an mdl specific thing, there is a general problem that a verifier must submit a request to the wallet without knowing anything about the wallet, and, for example, obtaining an mdl regardless of where the mdl was issued/what wallet it is in means:

1. The verifier may support more than one `client_id_scheme` and cannot know which one(s) the wallet(s) installed on the user's device support (e.g. for mdl USA states likely require x509_san_dns, some parts of the EU may well use OID Federation instead, and for non-mdl cases there may be even more deviation - this may also mean that the client_ids differ as, for example, there isn't a client_id that is valid for both x509_san_dns and openid federation)
2. For each client_id_scheme, the verifier may have more than one way to authenticate, e.g. for x509_san_dns they may have a x509 certificate suitable for use in California and a second one suitable for use in the EU.

jogu:
And to copy across [Torsten's comment](https://github.com/openid/OpenID4VP/issues/219#issuecomment-2332252832) with one idea for a solution:

This proposal is based on a discussion with @jogu @martijnharing @awoie @danielfett @bc-pi:

We would like to omit canonicalization of the request, that's why the basis of the proposal is to use the multi signature feature provided by the JWS JSON Serialization. 

The OID4VP request

```
{
  "expected_origins": [
    "https://origin1.example.com",
    "https://origin2.example.com"
  ],
  "response_type": "vp_token",
[...]

Issue #247, opened 2024-09-05
Allow the wallet to choose how to get the public key to authenticate the request and the client


awoie:
The current specification imposes limitations on how the wallet must authenticate the client and where to retrieve the public key to verify the signature of a signed request. In scenarios where both the verifier and the wallet support multiple authentication methods, the wallet may not always have access to all the required options or trust points. This can lead to failed transactions and a suboptimal user experience.

To maintain security while improving user experience, we propose removing the `client_id_scheme` parameter and allowing the verifier to provide multiple authentication methods. The wallet would then be able to select from the supported and accepted options.

Proposal:
- Remove the `client_id_scheme` parameter.
- Allow the verifier to offer multiple authentication options for the wallet, specifically for retrieving the public verification key for signed request objects. It is up to the verifier to decide whether to provide multiple options, and the wallet can choose which options to accept.

PR #237 may address this need.

jogu:
I'm not sure I understand the details of this proposal yet, but I can see two things that would need to happen in addition to removing client_id_scheme and also multiple authentication options:

1. The verifier needs a way to be able to supply multiple x509 certificates, e.g. if it wants to authenticate using x509_san_dns and certificates from multiple countries/states (i.e. you can't specify multiple x5c headers, and although you can put multiple certs in a single x5c these must form a single chain)
2. The verifier needs a way to specify multiple client_ids as it may be trying to, for example, authenticate with both openid federation (requiring a client_id in the form `https://example.com`) and x509_san_dns (requiring a client_id in the form `example.com`) - or we need to redefine some of the client_id_schemes so a single client_id does work across all of them (but I'm not sure that's possible in practice)

jogu:
https://github.com/openid/OpenID4VP/issues/248 proposes what I think may be a more general solution.

Sakurann:
resolved by https://github.com/openid/OpenID4VP/pull/263 and https://github.com/openid/OpenID4VP/pull/308

Issue #213, opened 2024-07-11
What is effective client_id in unsigned browser requests?


jogu:
The browser API appendix says:

> The `client_id` and `client_id_scheme` MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the Client Identifier from the origin as asserted by the Web Platform and/or app platform.

I'm not clear if "determines" here means "client_id is the origin". If it's not we need to say what we do mean as the the verifier needs to know what value it needs to check for in `aud` in the response.

jogu:
In the unsigned request section it says:

> In this case, the Wallet will use the Verifier's origin as asserted by the Browser as the Verifer's Client Identifier

So I think we just need to update the language in the "determines" clause to make it more direct.

bc-pi:
https://github.com/openid/OpenID4VP/pull/263#discussion_r1775606935 if accepted and then the PR merged, I think, might address this issue

Sakurann:
addressed in PR #263

PR #263, opened 2024-09-13
Change `client_id_scheme` to a prefix

### LINKED ISSUE #29: Unclear what happens if client_id_scheme appears twice
### Original Reporter: oliver-terbu

Clarify why `client_id_scheme` is needed in the request and in the Verifier Metadata. If this is by design, we need to clarify what happens in case both are present and potentially different.

### LINKED ISSUE #124: `client_id_scheme` security considerations

danielfett:
Solves #124 by removing client_id_scheme as a separate parameter and making it a prefix instead. `entity_id` was renamed to `https` in order to avoid breaking OpenID Federation.

Fixes #124 
Fixes #29


Sakurann (approved):
I think this is the cleanest approach that solves main security problem while not breaking other specifications like OpenID Federation and DIDs, while following OAuth 2.0 principle of client_id/aud uniquely identifying the client throughout the transaction

openid-4-verifiable-presentations-1_0.md
```diff
@@ -471,15 +485,15 @@ Body
...
+* `x509_san_uri`: When the Client Identifier Scheme is `x509_san_uri`, the Client Identifier MUST be a URI and match a `uniformResourceIdentifier` Subject Alternative Name (SAN) [@!RFC5280] entry in the leaf certificate passed with the request. The request MUST be signed with the private key corresponding to the public key in the leaf X.509 certificate of the certificate chain added to the request in the `x5c` JOSE header [@!RFC7515] of the signed request object. The Wallet MUST validate the signature and the trust chain of the X.509 certificate. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter. If the Wallet can establish trust in the Client Identifier authenticated through the certificate, e.g. because the Client Identifier is contained in a list of trusted Client Identifiers, it may allow the client to freely choose the `redirect_uri` value. If not, the `redirect_uri` value MUST match the Client Identifier. Example Client Identifier: `x509_san_uri:https://client.example.org/cb`.
```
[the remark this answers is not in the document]
danielfett: Fixed, thanks!

jogu:
Thanks Daniel!

One final thought: I am dubious about adding the prefix to pre-registered client ids. I don't think it actually really matters for VP, but I think it creates a clear gap as there's no way it could be accepted in normal OAuth, and if might be better if we chose something that could potentially eventually make it into an OAuth standard. Would removing the prefix for pre-registered clients and saying this work:

> pre-registered client ids MUST NOT contain `:`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -457,9 +471,9 @@ Location: https://client.example.org/universal-link?
...
+* `https`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Since the Entity Identifier already is defined to start with `https:`, the `https` does not need to be prefixed additionally. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `https://federation-verifier.example.com`.
```
selfissued, line 474:
```suggestion
* `https`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Since the Entity Identifier already is defined to start with `https:`, the `https` is not prefixed additionally. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `https://federation-verifier.example.com`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -457,9 +471,9 @@ Location: https://client.example.org/universal-link?
...
+* `did`: This value indicates that the Client Identifier is a DID defined in [@!DID-Core]. The request MUST be signed with a private key associated with the DID. A public key to verify the signature MUST be obtained from the `verificationMethod` property of a DID Document. Since DID Document may include multiple public keys, a particular public key used to sign the request in question MUST be identified by the `kid` in the JOSE Header. To obtain the DID Document, the Wallet MUST use DID Resolution defined by the DID method used by the Verifier. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter as defined in (#vp_token_request). Example Client Identifier: `did:example:123#1`.
```
selfissued, line 476:
```suggestion
* `did`: This value indicates that the Client Identifier is a DID defined in [@!DID-Core]. Since the DID URL already is defined to start with `did:`, the `did` is not prefixed additionally. The request MUST be signed with a private key associated with the DID. A public key to verify the signature MUST be obtained from the `verificationMethod` property of a DID Document. Since DID Document may include multiple public keys, a particular public key used to sign the request in question MUST be identified by the `kid` in the JOSE Header. To obtain the DID Document, the Wallet MUST use DID Resolution defined by the DID method used by the Verifier. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter as defined in (#vp_token_request). Example Client Identifier: `did:example:123#1`.
```

selfissued (approved):
Approved with requests for clarifying changes.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -432,22 +447,21 @@ Note: "https://self-issued.me/v2" is a symbolic string and can be used as an `au
...
 
-* `redirect_uri`: This value indicates that the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used) is also the value of the Client Identifier. The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#vp_token_request).
+* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#vp_token_request). Example Client Identifier: `redirect_uri:https%3A%2F%2Fclient.example.org%2Fcb`.
```
tplooker, line 456: I dont think we should be continuing to support this client ID scheme because its security properties are quite un-usual. Namely because the authorisation request can't be signed and we can't have pre-registered metadata nor a trusted way to resolve the metadata (e.g via a URL) the client is entirely open to impersonation with regard to what it supports from the protocol.

openid-4-verifiable-presentations-1_0.md
tplooker: I don't think we want to prescribe OpenID Federation as the only way metadata could be resolved from the client ID when it is an HTTPS url. There are multiple well-known metadata locations proposed for client metadata and a similar precedent exists within OAuth/OpenID for IDP / Authorisation Server metadata.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -432,22 +447,21 @@ Note: "https://self-issued.me/v2" is a symbolic string and can be used as an `au
...
+The Client Identifier Scheme enables deployments of this specification to use different mechanisms to obtain and validate metadata of the Verifier beyond the scope of [@!RFC6749]. The term Client Identifier Scheme is used since the Verifier is acting as an OAuth 2.0 Client.
 
-This specification defines the following values for the `client_id_scheme` parameter, followed by the examples where applicable: 
+This specification defines the following Client Identifier Schemes, followed by the examples where applicable: 
 
-* `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
+* `pre-registered`: This value represents the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms. Example Client Identifier: `pre-registered:e1fada73-8c43-4745-9756-303c2dc938a6`.
```
tplooker, line 454: What happens in the event this suffix is missing? E.g ordinary OAuth2.

openid-4-verifiable-presentations-1_0.md
jogu: The intention is that federation is the only think that uses the https client id scheme.

This doesn't stop a well-known based client id scheme being created, it would just use a client_id like `wellknown:https://rp.example.com/`.

If you think the text isn't conveying this could you suggest alternative wording please?

openid-4-verifiable-presentations-1_0.md
jogu: I think we could recommend against using it in production (if we don't already), but given it gets quite wide usage in non-production implementations I'm not sure it would be helpful to remove it unless we're replacing it with something that has a similar level of ease of use. (I definitely don't think we should remove it in this PR, it would be out of scope here.)

c2bo (commented):
There are quite a few files in the diagrams folder that use client_id_scheme & client_id - should those be changed as well?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -243,11 +243,26 @@ when it receives `request_uri_method` parameter with the value `post` but does n
...
+
+For example, a request might contain `client_id=verifier_attestation:example-client` to indicate that the `verifier_attestation` Client Identifier Scheme is to be used and that within this scheme, the Verifier can be identified by the string `example-client`. The presentation would contain the string `verifier_attestation:example-client` as the audience (intended receiver). In all other places within the OAuth flow where the Client Identifier is used, the full string `verifier_attestation:example-client` would be used.
+
+Important: Without the prefix, the `orig_client_id` is not sufficient to identify the Verifier, as different Verifiers (under control by different entities and using different cryptographic material) might have the same `orig_client_id` within different `client_id_scheme` namespaces. Therefore, `orig_client_id` MUST NOT be used independently within the context of the Wallet or its responses to identify the client. Only the full string `<client_id_scheme>:<orig_client_id>` is the Client Identifier of the Verifier, in particular in places where the Client Identifier is used in [@!RFC6749] and in the presentation returned to the Verifier.
```
tplooker, line 259: How does this preserve the ordinary behaviour of OAuth2 and OpenID e.g where the client ID absent a scheme is to be treated on face value as a pre-registered client

openid-4-verifiable-presentations-1_0.md
```diff
@@ -243,11 +243,26 @@ when it receives `request_uri_method` parameter with the value `post` but does n
...
+
+Note that the Verifier needs to determine which Client Identifier schemes the Wallet supports prior to sending the Authorization Request in order to choose a supported scheme.
```
tplooker, line 261: As I've highlighted before this is a huge limitation of this approach and is the basis of discussion in #248 and #247. It essentially means that there needs be some form of out of band negotiation model in order for a verifier to pick the right scheme. I don't see how this is practically possible.

openid-4-verifiable-presentations-1_0.md
tplooker: Given the original context in which this issue was raised was due to security concerns with client_id_scheme I do think while we are revisiting this functionality in the specification we should also consider removing it. Personally I've not seen wide usage of this particular client id scheme.

openid-4-verifiable-presentations-1_0.md
jogu: The changes in https://github.com/openid/OpenID4VP/issues/248 allow this to work and can be easily updated to work with the changes in this PR (i.e. putting everything in `client_id` instead of having a separate client_id_scheme in the unprotected header). I can't see any real practical issues with that approach.

we only have two proposals on how to fix this on the table and (as I said on the issue) I don't think the suggestion in #247 fully solves the problem.

openid-4-verifiable-presentations-1_0.md
jogu: It's supported in the conformance suite and about half the tests run have used the redirect_uri client_id_scheme. 

The lower security associated with redirect_uri now only occur when the verifier deliberately uses redirect_uri (rather than their being the potential for redirect_uri to be undetectably used when the verifier hadn't intended it), so we've made progress. Happy to see if the WG would rather remove it, but we should keep that as a separate issue & PR.

openid-4-verifiable-presentations-1_0.md
jogu: I agree, see https://github.com/openid/OpenID4VP/pull/263#issuecomment-2349061834 for my suggestion.

openid-4-verifiable-presentations-1_0.md
tplooker: Thanks, I've left a comment on #248.

examples/client_metadata/client_code_format.json
```diff
@@ -1,5 +1,5 @@
 {
-  "client_id": "s6BhdRkqt3",
+  "client_id": "pre-registered:s6BhdRkqt3",
```
Sakurann, line 2:
```suggestion
  "client_id": "s6BhdRkqt3",
```

openid-4-verifiable-presentations-1_0.md
tplooker:
```suggestion
* `https`: This value indicates that the Clients metadata is available at a well-known location such as defined in [@!OpenID.Federation]. Processing rules for retrieving and validating the metadata for the well-known location being used, such as that given in [@!OpenID.Federation] MUST be followed. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `https://federation-verifier.example.com`.
```

jogu: That doesn't reflect the intention though. If people want to use https urls with a different resolution method, they use a client id scheme other than https, e.g. `wellknown:https://rp.example.com/.` The "https" **client-id scheme** is reserved for federation only.

Avoiding the vagueness of "I'm going to somehow authenticate with you using this url, you go figure out how you think I'm trying to authenticate with you" is very much the point. If you think of the browser API, using the https client id scheme to reflect multiple ways the wallet can authenticate, can result in lots of deadends as the sandboxed wallet matcher is now unable to know if it supports the method the wallet is using to authenticate, because it literally doesn't know how the wallet is trying to authenticate and has no way to find out because it can't access the network.

danielfett: @selfissued I'll change this to a "MUST NOT be prefixed" to be very clear.

@tplooker @jogu I agree with Joseph here. This is an opportunity to have a clearly defined mechanism both from a security point of view and from an implementation perspective. Unless there is a good reason to not prefix other https-based client id schemes, we should avoid it.

openid-4-verifiable-presentations-1_0.md
danielfett: This has been resolved in the latest revision of the draft using the rules proposed by @jogu

openid-4-verifiable-presentations-1_0.md
danielfett: @tplooker Please file a separate issue for this.

openid-4-verifiable-presentations-1_0.md
danielfett: This has been resolved in the latest revision of the draft.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -282,23 +277,27 @@ The following additional considerations are given for pre-existing Authorization
...
 ```
 
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_id_scheme` and `client_metadata` parameters): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional`client_metadata` parameters): 
 
 ```
 GET /authorize?
-  client_id=client.example.org
-  &client_id_scheme=x509_san_dns
+  client_id=x509_san_dns:client.example.org
```
peppelinux, line 300: I believe that this doesn't really scale

My name is "Giuseppe De Marco", not "trust-evaluation-mechanism:Giuseppe De Marco"

who verifies the trust with me uses the mechanisms it supports and that we might have in common, for this reason it is important to have a way to hint these mechanisms.

Since a usbject might be part of multiple networks and using more than a single trust framework, it would be better to hints a list of supported trust evaluation mechanisms, instead of embedding only one of them within the entity identifier, such as the client_id

I believe that this aspect is crucial for the grow of the ecosystem beyond the bounduaries if a single trust framework or network

openid-4-verifiable-presentations-1_0.md
```diff
@@ -430,24 +429,52 @@ When the Verifier is sending a Request Object as defined in [@!RFC9101], the `au
...
 
-## Verifier Metadata Management {#client_metadata_management}
+## Client Identifier Scheme and Verifier Metadata Management {#client_metadata_management}
+
+This specification defines the concept of a Client Identifier Scheme that indicates how the Wallet is supposed to interpret the Client Identifier and associated data in the process of Client identification, authentication, and authorization. The Client Identifier Scheme enables deployments of this specification to use different mechanisms to obtain and validate metadata of the Verifier beyond the scope of [@!RFC6749]. The term Client Identifier Scheme is used since the Verifier is acting as an OAuth 2.0 Client.
+
+The Client Identifier Scheme is a string that MAY be communicated by the Verifier in a prefix within the `client_id` parameter in the Authorization Request. A fallback to pre-registered clients as in [@!RFC6749] remains in place as a default mechanism in case no Client Identifier Scheme was provided. A certain Client Identifier Scheme may require the Verifier to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
```
peppelinux, line 436:
```suggestion
The Client Identifier Scheme is a string that MAY be communicated by the Verifier in a prefix within the `client_id` parameter in the Authorization Request. A fallback to pre-registered Clients as in [@!RFC6749] remains in place as a default mechanism in case no Client Identifier Scheme was provided. A certain Client Identifier Scheme MAY require the Verifier to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
```

openid-4-verifiable-presentations-1_0.md
javereec: I agree with @peppelinux. IMO, adding a prefix seems to introduce mandatory extra semantics to a field which are limiting.

I've also had this idea of something similar to `login_hint` found in other specifications that hint how client can be authenticated.

jogu (approved):

openid-4-verifiable-presentations-1_0.md
Sakurann: very strong suggestion to discuss `a single deployment enabling multiple trust frameworks and evaluation mechanisms` topic into a separate issue #248.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -243,11 +243,26 @@ when it receives `request_uri_method` parameter with the value `post` but does n
...
+
+The value of the `client_id` Authorization Request parameter, as defined in [@!RFC6749], therefore MUST be structured as follows: 
+```
+<client_id_scheme>:<orig_client_id>
+```
+
+Here, `<client_id_scheme>` is the Client Identifier Scheme and `<orig_client_id>` is the identifier for the Client within the namespace of that scheme. See (#client_metadata_management) for Client Identifier Schemes defined by this specification. 
+
+For example, a request might contain `client_id=verifier_attestation:example-client` to indicate that the `verifier_attestation` Client Identifier Scheme is to be used and that within this scheme, the Verifier can be identified by the string `example-client`. The presentation would contain the string `verifier_attestation:example-client` as the audience (intended receiver). In all other places within the OAuth flow where the Client Identifier is used, the full string `verifier_attestation:example-client` would be used.
```
lj-raidiam: We could reword it a little bit for a better clarity. 

```suggestion
For example, a presentation request might contain `client_id=verifier_attestation:example-client` to indicate that the `verifier_attestation` Client Identifier Scheme is to be used and that within this scheme, the Verifier can be identified by the string `example-client`. The presentation would contain the full `verifier_attestation:example-client` string as the audience (intended receiver) and the same full string would be used as the Client Identifier anywhere in the OAuth flow.```

File: openid-4-verifiable-presentations-1_0.md (hunk 9 above)
Comment (danielfett) [reply openid-4-verifiable-presentations-1_0.md:300]: @peppelinux If one of the trust frameworks uses federation and one uses x509_san_dns, then you cannot have the same name in both, regardless of whether client_id_scheme is a separate parameter, a prefix, or doesn't exist at all. Therefore what you're describing is a separate problem that we should discuss in #248.

Comment (jogu) [reply openid-4-verifiable-presentations-1_0.md:300]: > I agree with @peppelinux. IMO, adding a prefix seems to introduce mandatory extra semantics to a field which are limiting.
>
> I've also had this idea of something similar to login_hint found in other specifications that hint how client can be authenticated.

We don't have any other workable suggestions, and we've been at this for 6+ months now. I not sure there's any significant difference between a login_hint parameter like you mention and the existing client_id_scheme parameter.  Besides as Daniel says, we have multiple client_id_schemes, and there is no consistency in the values used (some use hostnames, some use https urls, some use dids). Having a single clientid that works across multiple ecosystems in the general case is not something anyone has proposed a workable solution for, most likely I'm guessing because it's now an unsolvable problem. (Hence why as Daniel/Kristina mentioned we have #248 to make the underlying problem work without assuming a client id is the same across all these ecosystems.)

The root problem we're getting at is that we've introduced a whole load of new ways for clients to assign their own client id, and the door is also open for other specifications to define more ways. Not namespacing the client id to indicate the mechanism via which it was assigned would be dangerous (because it opens up the very real possibility that we've seen actual exploits for in the real world that an attacker controlled piece of software is able to create a client id it was believed belonged uniquely to someone else), and there's no other solutions on the table.

The comparison with people's names isn't really valid as people's names aren't unique identifiers. Perhaps company names are a better analogy, not sure? But in that case, you are definitely namespace by the allocation mechanism, and in that case you always have to state "I'm Acme Inc registered in Delaware" when you are communicating with someone else, and if the other party wants to respond in a way that it's clear they're addressing you, they have to say "This message is for Acme Inc registered in Delaware".

File: openid-4-verifiable-presentations-1_0.md (hunk 11 above)
Comment (danielfett) [reply openid-4-verifiable-presentations-1_0.md]: Thanks, I'll commit this manually.

File: openid-4-verifiable-presentations-1_0.md (hunk 2 above)
Comment (tplooker) [reply openid-4-verifiable-presentations-1_0.md]: Federation is an overly complicated mechanism for a client to simply convey its metadata at a well-known location, I don't believe we should be promoting it as the preferred mechanism for clients that are choosing to identify via an HTTPS url which is what this proposal does. It relegates other HTTPS based url based client metadata discovery mechanism to requiring a double prefix style client id which I don't think is a clean solution. 

If we would prefer to have less ambiguity then the text I suggested, I propose we have a seperate parameter to convey the metadata location for the client such as `client_metadata_location=federation`.

File: openid-4-verifiable-presentations-1_0.md (hunk 4 above)
Comment (tplooker) [reply openid-4-verifiable-presentations-1_0.md]: Agreed have opened as #269

File: openid-4-verifiable-presentations-1_0.md (hunk 2 above)
Comment (lj-raidiam) [reply openid-4-verifiable-presentations-1_0.md]: @tplooker Is it going to be any different than `client_id_scheme` for https-based urls?

Comment (jogu) [reply openid-4-verifiable-presentations-1_0.md]: Just to be clear, we are not promoting federation as a preferred mechanism. We are merely acknowledging that federation is already the long established way of client's using self-allocated client_ids beginning with https: and that there's not really any option open to us of changing that position.

The alternative (following in the vein of this PR) is that federation is also prefixed, but that introduces interoperability issues / confusion and still doesn't solve the 'clean solution' problem.

I think if we do want to do something that is a generic scheme based on https urls (and I think that could be a good idea), we need to do that with a prefix, otherwise we're going to overlap and/or conflict with at least two other different specifications. It's simply not sustainable or sensible to have multiple specs trying to define self-assert client ids that are pure https urls. Even federation probably shouldn't be doing it, but that ship has mostly sailed.

File: openid-4-verifiable-presentations-1_0.md (hunk 9 above)
Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:304]: @danielfett 

> If one of the trust frameworks uses federation and one uses x509_san_dns, then you cannot have the same name in both

The goal I have in mind  is to have a single deployment using multiple trust frameworks.

Technically, I can deploy a system using both metadata exchanged only via TLS and well-known endpoints, as well as OpenID Federation, and also x509 certificates using SAN. The first method aids in discovery and cryptographic key fetching, while the latter is used solely for cryptographic key fetching, with metadata fetching as an optional addition (if required, see ISO 18013-5, which operates without any metadata discovery).

Therefore, the name is always unique when based on DNS and can be more detailed (enabling a more flexible way to deploy multitenancy) using HTTPS URLs with web paths within the entity ID value. It is up to the deployers to decide their deployment strategy to match all the trust frameworks they may want to use. In the technical specification, we should enable this kind of flexibility.

Comment (peppelinux) [reply openid-4-verifiable-presentations-1_0.md:304]: technically, using X.509 certificates and SAN

X509v3 extensions:
    X509v3 Subject Alternative Name:
        URI:https://example.com/rp/entity1

Comment (jogu) [reply openid-4-verifiable-presentations-1_0.md:304]: @peppelinux 

>  In the technical specification, we should enable this kind of flexibility.

As I think I mentioned in another comment somewhere, we're not preventing people defining new client id schemes that have behaviour like this (whether it is a good idea for people to do so or not is a separate matter and likely depends on many details that haven't yet been talked about). If anything this PR is enabling that extensions like that in a way that doesn't result in security issues and confusion. If someone manages to come up with a single client id scheme that works for every single ecosystem/trust framework, then that's an even better outcome!

File: openid-4-verifiable-presentations-1_0.md [hunk 12]
```diff
@@ -434,24 +433,52 @@ When the Verifier is sending a Request Object as defined in [@!RFC9101], the `au
...
 
-The following is a non-normative example of a request when `client_id` equals `redirect_uri`.
+For example, an Authorization Request might contain `client_id=verifier_attestation:example-client` to indicate that the `verifier_attestation` Client Identifier Scheme is to be used and that within this scheme, the Verifier can be identified by the string `example-client`. The presentation would contain the full `verifier_attestation:example-client` string as the audience (intended receiver) and the same full string would be used as the Client Identifier anywhere in the OAuth flow.
+
+Without the prefix, `example-client` in the example above would be interpreted as referring to a pre-registered client, as defined below. Therefore, Wallets MUST always use the full Client Identifier, including the prefix if provided, within the context of the Wallet or its responses to identify the client. This refers in particular to places where the Client Identifier is used in [@!RFC6749] and in the presentation returned to the Verifier.
```
danielfett, line 453:
```suggestion
For example, an Authorization Request might contain `client_id=verifier_attestation:example-client` to indicate that the `verifier_attestation` Client Identifier Scheme is to be used and that within this scheme, the Verifier can be identified by the string `example-client`. The presentation would contain the full `verifier_attestation:example-client` string as the audience (intended receiver) and the same full string would be used as the Client Identifier anywhere in the OAuth flow.

Without the prefix, `example-client` in the example above would be interpreted as referring to a pre-registered client, as defined below.

Confusing Verifiers using a Client Identifier Scheme with those using none can lead to attacks. Therefore, Wallets MUST always use the full Client Identifier, including the prefix if provided, within the context of the Wallet or its responses to identify the client. This refers in particular to places where the Client Identifier is used in [@!RFC6749] and in the presentation returned to the Verifier.
```

lj-raidiam (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -282,27 +277,31 @@ The following additional considerations are given for pre-existing Authorization
...
   response_type=vp_token
-  &client_id=https%3A%2F%2Fclient.example.org%2Fcb
+  &client_id=redirect_uri:https%3A%2F%2Fclient.example.org%2Fcb
   &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
   &presentation_definition=...
   &nonce=n-0S6_WzA2Mj HTTP/1.1
 ```
 
-The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_id_scheme` and `client_metadata` parameters): 
+The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional`client_metadata` parameters):
```
danielfett, line 300:
```suggestion
The following is a non-normative example of an Authorization Request with a `request_uri_method` parameter (including the additional `client_metadata` parameters): 
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: I completely restructured the section, please re-review.

danielfett:
FYI, I restructured the section on client id schemes to <strike>tighten the exp...</strike> improve readability.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1596,15 +1624,14 @@ try {
...
-* `client_id_scheme`
 * `response_type`
 * `response_mode`
 * `nonce`
 * `presentation_definition`
 * `client_metadata`
 * `request`
 
-The `client_id` and `client_id_scheme` MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the Client Identifier from the origin as asserted by the Web Platform and/or app platform. The transport of the request and origin from the Web Platform and/or app platform to the Wallet is platform-specific and is out of scope of this profile.
+The `client_id` MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the Client Identifier from the origin as asserted by the Web Platform and/or app platform. The transport of the request and origin from the Web Platform and/or app platform to the Wallet is platform-specific and is out of scope of this profile.
```
bc-pi, line 1634: Per my (possibly flawed) understanding, this effectively means that an unsigned request will have a Client Identifier determined to be something like "https://verifiers.domain.com" which is a value that "indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation" https://github.com/openid/OpenID4VP/pull/263/files#diff-3118c9756a1d7b361bb53af8bd9d65666476a27cfeeced56c47a5dccc38eac55R494 

It is not that. 

This bit of text was arguably a bit unspecified before this PR. 

But now it's wrong.

bc-pi (changes requested):
I am philosophically opposed to special treatment of DIDs and Federation. 

If we're gonna prefix, prefix all the new stuff equally. And ":" needn't be the prefix indicator. 

Also it causes breakage in the Appendix A. OpenID4VP profile for the W3C Digital Credentials API  https://github.com/openid/OpenID4VP/pull/263#discussion_r1773794765

openid-4-verifiable-presentations-1_0.md
peppelinux: since the client_id using HTTPS URL doesn't make any conflicts with any other trust establishment mechanisms that might come or be hinted in the future, I can approve this PR thinking that fixed namespaces will not scale, while https url would be more flexible.

I'm not very happy about this, but happiness doesn't necessarily have to be found in github reviews :-)

peppelinux (approved):

openid-4-verifiable-presentations-1_0.md
peppelinux: @tlodderstedt client_id_scheme as single valued array didnt scale as well, it would have been an array to scale. 
More specifically about your question: It was not a comparison between the two but a vision about improve the specification with something more flexible. 

approved, let's try to move forward lightly.

openid-4-verifiable-presentations-1_0.md
danielfett: As discussed on the call, an option would be that the wallet adds a prefix indicating that it derived the client id from the web platform, e.g., `api-orign:`.

Sakurann:
discussed on the WG call. need to clarify the behavior of client id in an unsigned request in the browser API. please make suggestions.

openid-4-verifiable-presentations-1_0.md
bc-pi: This option was more or less the agreed to path forward in the call (as I remember it anyway).

bc-pi: I still object to special treatment of DIDs and Federation. And do not buy the argument about DIDs not being special. But in the interest of moving forward, I've conceded on that point and agreed on fixing things here. And also came to realize that #213 is related.

bc-pi: Here is an attempt at some suggested text, which I half promised to do during the call.

```suggestion
The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the effective Client Identifier from the origin as asserted by the Web Platform and/or app platform. The effective Client Identifier is composed of a synthetic Client Identifier Scheme of `web-orign` and the origin itself. For example, an origin of `https://verifier.example.com` would result in an effective Client Identifier of `web-orign:https://verifier.example.com` The transport of the request and origin from the Web Platform and/or app platform to the Wallet is platform-specific and is out of scope of this profile.
```

The text in the {#unsigned_request} section likely needs a bit of an update too. But that content is too far out of scope from the changes in the PR to make a suggestion.

bc-pi: WRT the text in {#unsigned_request} perhaps,

> The Verifier MAY send all the OpenID4VP request parameters as members in the `request` member passed to the API. In this case, the Wallet will use the Verifier's origin as asserted by the user agent as the Verifer's Client Identifier. 

to

> The Verifier MAY send all the OpenID4VP request parameters as members in the `request` member passed to the API. In this case, the Wallet will use the Verifier's origin as asserted by the user agent to determine the Verifier's effective Client Identifier.

or something

Sakurann:
I think this PR also resolves https://github.com/openid/OpenID4VP/issues/29?

jogu:
> I think this PR also resolves #29?

I don't think so. I'll comment on the issue.

danielfett:
It indeed resolves #29 as it changes the title of the section that @jogu identified as confusing.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -461,9 +491,9 @@ Location: https://client.example.org/universal-link?
...
+* `https`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Since the Entity Identifier is already defined to start with `https:`, this Client Identifier Scheme MUST NOT be prefixed additionally. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `https://federation-verifier.example.com`.
```
danielfett, line 494: Following Brian's comments, I was wondering if we should open this up to other client ID schemes using `https` without an additional prefix. For example:

```suggestion
* `https`: This value indicates that the Client Identifier is an HTTPS URL. Since the Entity Identifier is already defined to start with `https:`, this Client Identifier Scheme MUST NOT be prefixed additionally. Unless another mechanism for resolving the Client Identifier and the metadata is defined within a profile of this specification, the following rules apply: 
     * The Client Identifier MUST be treated as an OpenID Federation [@!OpenID.Federation] Entity Identifier. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `https://federation-verifier.example.com`.
```

Sakurann: that's like saying "here are the rules we define for prefixing, but other rules for the same prefix can be defined outside this specification", so I don't think that's a good idea.

openid-4-verifiable-presentations-1_0.md
bc-pi: Thank you Dr. @danielfett

openid-4-verifiable-presentations-1_0.md
jogu: I agree with Kristina, I'm reluctant to open the door to allowing other schemes to use https without prefixing. It just creates a mess and confusion and interoperability problems.

I'm only reluctantly accepting that Federation should be allowed to use https urls without a prefix. I cannot see any strong reason to let others do so. Having a prefix really causes no technical issue unless something is already in production, and I'm pretty sure federation is the only spec that can even attempt the 'already in production' argument.

openid-4-verifiable-presentations-1_0.md
danielfett: Resolving this as the discussion is continued in #248.

bc-pi (approved):
I *still* object to special treatment of OpenID Federation and W3C Decentralized identifiers but am approving in the interest of making progress.

javereec (approved):

Sakurann:
I note that I relate to Brian's uncomfortableness to OpenID Federation's "special" treatment, while recognizing that changing openid federation document to enable `openidfederation:https://...` is, unfortunately, probably much much harder...

jogu:
This is quite a significant change, and it's taken well over 6 months to get to this point - but we now have 8 approvals on it and no request for changes. We've discussed this particular approach I believe on all the working group meetings for at least 2 weeks.

Not everyone is happy but, other than/despite the above debate over whether federation should have the "https" prefix or not (and similar for 'did'), I believe we've reached a solid consensus that prefixing is a workable approach. We did ask Oliver for feedback too, but he's not added anything more, and I also sent a message on the mailing list on 23rd September asking for feedback by 1st Oct, and (other than comments on this PR) we've not received anything further - so as discussed on today's WG call - merging!

A big thank you to everyone that contributed here - even if the WG didn't go with your suggestions I genuinely believe that all the different ideas we've had helped guide the working group towards what we ended up with!

bc-pi:
> changing openid federation document to enable openidfederation:https://... is, unfortunately, probably much much harder...

Is it though? That openid federation document has been under development for more than eight years now and is on its 39th draft. And apparently a federation wallet architecture specification is needed for it to federation to work with wallets anyway. Changes aren't possible there? While introducing a breaking change to everything else (except for DIDs of course)...   

I honestly don't see any justification for special treatment.  

But I've approved anyway https://github.com/openid/OpenID4VP/pull/263#pullrequestreview-2341135999

bc-pi:
> I note that I relate to Brian's uncomfortableness to OpenID Federation's "special" treatment, while recognizing that changing openid federation document to enable `openidfederation:https://...` is, unfortunately, probably much much harder...

In

> > changing openid federation document to enable openidfederation:https://... is, unfortunately, probably much much harder...
> 
> Is it though? That openid federation document has been under development for more than eight years now and is on its 39th draft. And apparently a federation wallet architecture specification is needed for it to federation to work with wallets anyway. Changes aren't possible there? While introducing a breaking change to everything else (except for DIDs of course)...
> 
> I honestly don't see any justification for special treatment.
> 
> But I've approved anyway [#263 (review)](https://github.com/openid/OpenID4VP/pull/263#pullrequestreview-2341135999)

It's worth nothing, much after the fact but for posterity, that no change to the openid federation document is even needed in order for OpenID4VP document to be self-consistent and reasonable.

PR #270, opened 2024-09-26
clarify base64url definition

### LINKED ISSUE #236: Clarify that the `vp_token` is **URL Safe** Base64 encoded, rather than a Base 64 encoded
> I ran into this issue while parsing the response by reading the spec, so figured others would too.

From:

https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-6.1

It says:

> Each Verifiable Presentation MUST be represented as a JSON string (that is a Base64url encoded value) 

Which is correct, but incomplete. I think it should say something like:

> Each Verifiable Presentation MUST be represented as a JSON string (that is a [URL Safe] Base64url encoded value) 

For implementers, that's the difference between decoding as:

```javascript
// From https://stackoverflow.com/questions/28100601/decode-url-safe-base64-in-javascript-browser-side
atob(data.replace(/_/g, '/').replace(/-/g, '+'))
```

Rather than:

```javascript
atob(data)
```

Sakurann:
resolves #236

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -64,7 +64,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 # Terminology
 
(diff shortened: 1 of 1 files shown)


c2bo (commented):
I like the change - makes it pretty clear imho

2 parts that I think should also be slightly adjusted / wasn't sure about that also reference base64url encoding.

761:
> The following is a non-normative example of the payload of the JWT used in the example above before base64url encoding and signing:

--> something like
> The following is a non-normative example of the payload of the JWT used in the example above before getting base64url-encoded and signed:

1073 - not sure about this one:
> (1) The Verifier produces a `nonce` value by generating at least 16 fresh, cryptographically random bytes with sufficient entropy, associates it with the session and base64url encodes it.

bc-pi (approved):

tplooker (approved):

Sakurann:
@c2bo i think it is sufficiently clear that those parts you mention also refer to base64url-encoded.

PR #271, opened 2024-09-26
Move the response encryption key remark outside of the new parameters list

### LINKED ISSUE #176: Move the response encryption key remark outside of the new parameters list
This fragment cuts the list into two fragments. It is unclear whether the `request_uri_method` that follows is still part of the same list and why the description of something that is not a new parameter appears here.

> A public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)). It MAY be passed by the Verifier using the `jwks` or the `jwks_uri` claim within the `client_metadata` or `client_metadata_uri` request parameter. 

I would reword it slightly ...

_If encryption of the Authorization Response is needed, the public key to be used by the Wallet as an input to the key agreement to encrypt Authorization Response (see (#jarm)) MAY be passed by the Verifier using the jwks or the jwks_uri claim in the Client Metadata._

... and move it after the `request_uri_method`. 
https://github.com/openid/OpenID4VP/blob/9a8ac76f93d0240767d92152085ae08c420d71cc/openid-4-verifiable-presentations-1_0.md?plain=1#L276

Any opinions?

Sakurann:
resolves #176

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -263,10 +263,10 @@ This specification defines the following new request parameters:
 `client_metadata`:
 : OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded.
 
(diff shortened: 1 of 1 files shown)


jogu:
It looks like this is already fixed by https://github.com/openid/OpenID4VP/pull/233/files

jogu:
I think we can close this as per comments on the issue - thanks Kristina!

PR #280, opened 2024-10-14
Remove urn:ietf:params:oauth urls

jogu:
As an alternative fix to the one in:

https://github.com/openid/OpenID4VP/pull/279

Remove the paragraph that appears to create the controverial URNs.

If these URNs are needed, the federation-wallet spec:

https://github.com/peppelinux/federation-wallet

seems to be a more appropriate place to define them.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1033,8 +1033,6 @@ The following is a non-normative example of the terms of use that may be defined
 }
 ```
 
-Federations that conform to those specified in [@OpenID.Federation] are identified by the `type` `urn:ietf:params:oauth:federation`. Individual federations are identified by the Entity Identifier of the trust anchor. If the federation decides to use trust marks as signs of whether an entity belongs to a federation or not then the federation is identified by the `type` `urn:ietf:params:oauth:federation_trust_mark` and individual federations are identified by the Entity Identifier of the trust mark issuer.
-
 Trust schemes that conform to the TRAIN [@TRAIN] trust scheme are identified by the `type` `https://train.trust-scheme.de/info`. Individual federations are identified by their DNS names.
 
 The following is a non-normative example of a `claims` parameter containing a `presentation_definition` that filters VCs based on their federation memberships:


bc-pi:
If this is an option, then that whole section should just be deleted.

jogu:
> If this is an option, then that whole section should just be deleted. The TRAIN reference is suspect at best and I'm not too sure the 'claims' parameter example is up to date.

I'm definitely happy to look at that, but I suspect it takes more discussion and there's a house of cards here - these OAuth urls in particular are holding up the IANA considerations PR, which is in turn going to conflict with other PRs. We either need to merge one of the changes, unblock things and open an issue to clean up further or drop them all from the upcoming ID.

bc-pi:
> > If this is an option, then that whole section should just be deleted. The TRAIN reference is suspect at best and I'm not too sure the 'claims' parameter example is up to date.
> 
> I'm definitely happy to look at that, but I suspect it takes more discussion and there's a house of cards here - these OAuth urls in particular are holding up the IANA considerations PR, which is in turn going to conflict with other PRs. We either need to merge one of the changes, unblock things and open an issue to clean up further or drop them all from the upcoming ID.
> 
> By contrast I think if this text around federation is needed (I'm honestly not sure) then it seems more sensible that it lives in the same place as other content in the federation-wallet spec (wherever that content ends up), so removing it here seems like it should be uncontentious.

See also https://github.com/openid/OpenID4VP/pull/274#discussion_r1790491130 and https://github.com/openid/OpenID4VP/pull/274#discussion_r1790466696 i guess but it is definitely a house of cards here. I thought just changing the URNs to something less wrong and importantly less permanent would be uncontentious too https://github.com/openid/OpenID4VP/pull/279 but that thought/hope proved wrong. 

Maybe I should create an alternative to this alternative...

bc-pi:
alternative to this alternative now at https://github.com/openid/OpenID4VP/pull/282

jogu:
no approvals at all on this one so I think we can close it in favour of https://github.com/openid/OpenID4VP/pull/282

PR #266, opened 2024-09-18
Introduce new query language (Approach 3)

danielfett:
This approach uses ids and a separate definition of valid sets of claims/credentials to return.

<strike>**The draft text has not been adapted yet, let's discuss the JSON examples first!**</strike>

Text is now updated as well, please review.

List of issues to solve in separate PRs:
- [x] Introduce transformed claims?
- [x] Determine if we need any other format-specific parameters
- [x] Define interaction with existing query lang (see https://github.com/openid/OpenID4VP/issues/211)
- [x] Define how requesting presentations with `scope` works with the new QL. 
- [x] Define new format for the VP Token
- [x] Maybe re-introduce purpose?

Fixes #178
Fixes #249
Fixes #255 (or rather, makes it obsolete)
Fixes #193
Fixes #161 (that was the discussion on the requirements for this new language)
Fixes #157 (mandatory to implement features of the query language)
Fixes #164 
Fixes #160 
Fixes #148

PR #281, opened 2024-10-15
Proposed alterations to how optionality is handled in query language (Approach 3)

tplooker:
As per https://github.com/openid/OpenID4VP/pull/266#issuecomment-2411134477 I also believe that optionality at the credential level as well as the claims level leads to difficult UX challenges and doesn't provide much value. 

What I propose we do is elevate things so there is in effect only one level of optionality that a wallet would need to communicate to a user about, which is at the "credential set query" level (not sold on that name I believe "usecase" might be a better descriptor here).

The main things I have done in this PR is
1. Remove optionality at the claim and credential level
2. Expanded the structure at the `credential_sets` level to handle optional queries instead. 

The easiest way to grok the difference from a syntax perspective IMO is through modification of [the credentials_alternatives.json example](https://github.com/openid/OpenID4VP/compare/danielfett/vp-query-language-3...tl/vp-query-language-3+?expand=1#diff-92252873aa616ade36134861650c7822cab4027498a3c68057a0157292ebcbeaR65-R80)

**Note** - I've included the `purpose` property mainly to help in understanding the examples, it hasn't been defined in spec text and we will likely want to remove it (if/before) this PR is to be merged or shift away from a free text field as way of conveying purpose.

### DIFF
--- a/examples/query_lang/claims_alternatives.json
+++ b/examples/query_lang/claims_alternatives.json
@@ -19,11 +19,10 @@
         {"id": "f", "path": ["email"]}
       ],
       "claim_sets": [ // defines the rules
-        // postal code or (locality and region), last_name, and date_of_birth are mandatory; email is optional 
-        ["a", "b", "e", "f?"],
-        ["a", "c", "d", "e", "f?"]
-        
-        // Note: the ? to mark a claim as optional is syntactical sugar; the rules could be rewritten without it, but that would require a lot of repetition
+        // postal code or (locality and region), last_name, and date_of_birth are mandatory; email is required if present on the matched credential 
+        ["a", "b", "e", "f?!"],
+        ["a", "c", "d", "e", "f?!"]
+        // Note: the ?! to mark a claim as `required if present` is syntactical sugar; the rules could be rewritten without it, but that would require a lot of repetition
       ]
     }
   ]
--- a/examples/query_lang/complex_mdoc.json
+++ b/examples/query_lang/complex_mdoc.json
@@ -1,0 +1,94 @@
+// Adding some comments here to ensure Brian isn't sad ;)
+{
+  "credentials": [
+    {
+      "id": "mdl",
+      "format": "mso_mdoc",
+      "meta": {
+        "doctype_values": ["org.iso.18013.5.1.mdl"]
+      },
+      "claims": [
+        {
+          "id": "age_over_18",
+          "namespace": "org.iso.18013.5.1",
+          "claim_name": "age_over_18"
+        },
+        {
+          "id": "birth_date",
+          "namespace": "org.iso.18013.5.1",
+          "claim_name": "birth_date"
+        },
+        {
+          "id": "portrait",
+          "namespace": "org.iso.18013.5.1",
+          "claim_name": "portrait"
+        }
+      ],
+      "claim_sets": [
+        [ "portrait", "age_over_18" ],
+        [ "portrait", "birth_date" ]
+      ]
+    },
+    {
+      "id": "photo_card",
+      "format": "mso_mdoc",
+      "meta": {
+        "doctype_values": ["org.iso.23220.photoid"]
+      },
+      "claims": [
+        {
+          "id": "age_over_18",
+          "namespace": "org.iso.23220.1",
+          "claim_name": "age_over_18"
+        },
+        {
+          "id": "birth_date",
+          "namespace": "org.iso.23220.1",
+          "claim_name": "birth_date"
+        },
+        {
+          "id": "portrait",
+          "namespace": "org.iso.23220.1",
+          "claim_name": "portrait"
+        }
+      ],
+      "claim_sets": [
+        [ "portrait", "age_over_18" ],
+        [ "portrait", "birth_date" ]
+      ]
+    },
+    {
+      "id": "loyalty_card",
+      "format": "mso_mdoc",
+      "meta": {
+        "doctype_values": ["loyalty_card"]
+      },
+      
+      "claims": [
+        {
+          "namespace": "namespace.loyalty_card",
+          "claim_name": "loyalty_card_number"
+        }
+      ]
+    }
+  ],
+  // The intention with the credential_sets array is that all "credential set queries" (objects in this array)
+  // are responded too, if they are either marked `required:true` explicitly or implicitly.
+  // Any "credential set queries" marked with `required:false` are optional for the wallet to return
+  "credential_sets": [
+    {
+      "purpose": "Prove your age",
+      "query_options": [ // Choose one of the elements in this array as the way you want to satisfy this credential set query
+        [ "mdl" ],
+        [ "photo_card" ] // All conditions for all elements in this inner array MUST be met, reason we have this inner array is incase a given usecase can be satisfied across multiple credentials.
+      ]
+    },
+    {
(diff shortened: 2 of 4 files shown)


jogu:
I'm not clear how the use cases I mentioned in https://github.com/openid/OpenID4VP/pull/266#issuecomment-2413821242 would work here.

> What I propose we do is elevate things so there is in effect only one level of optionality that a wallet would need to communicate to a user about, which is at the "credential set query" level (not sold on that name I believe "usecase" might be a better descriptor here).

I think this actually makes some cases harder, by forcing them to happen at credential selection time. In particular in the browser API because the credential selection has already happened by the time the wallet gets invoked, it seems like it leaves the wallet with nothing it needs to communicate with the user about as the only choice at that stage is "release selected credentials or don't". Trying to process the approach in this PR so that responsibility is divided between the wallet and the system credential selector actually feels harder for the wallet to me than the current approach.

I'm personally currently finding the "select credentlals then deal with what claims will be shared" model in the existing PR mentally easier as well as being more expressive / covering more use cases.

tplooker:
I've included an example here https://github.com/openid/OpenID4VP/pull/266#issuecomment-2414705934

Sakurann:
discussed on the wg call. i think the asks were:
- please keep discussions on this topic (UX issue, optionality, etc.) in this PR so that there is one place to look for
- please update the rationale/context for this proposed change
- please change the syntax so that no optionality is allowed at all at the claims level - currently it still does
- please update the examples to show how existing examples can still be represented with this new proposal

tplooker:
Post the WG call this morning I've done the following

- Added text around how the `claim_sets` array is to be processed, clarifying that its not an array of options that the user is expected to decide between, instead the wallet is to pick the highest preference option it can satisfy. Simply removing the `claim_sets` functionality would mean alternative claims become impossible to convey.
- Removed the "purpose" attribute for now
- Examples have been refined to show how it works

tplooker:
Additionally here is some more detailed rationale for this change

With the syntax in the original proposal because optionality is expressed at both the claim and credential level several UX challenges emerge.

The main issue boils down to how complex the consenting process would have to be in order to support optionality at both levels. For example optional claims would likely mean three possible different options

1. The wallet makes a blanket rule to not support optional claims and doesn't even render an option to the user, rejecting all user claims - which is a critical loss of functionality.
2. The wallet makes a blanket rule to support all optional claims and return them always without rendering an option to the user, which is potential privacy issue (hopefully wallets wouldn't do this in practise).
3. The wallet invests in a complex consent UX that allows a user to toggle optional claims in the consent process, arguably this is a poor user experience and critically it would lack conveying the context to the user as to for why these claims are being optionally requested AND perhaps how some of the optional claims relate to one and other.

Instead if we step back and consider when a verifier is making a request for some claims that are optional to return, the group of optional claims being requested likely has a purpose (in the case there is more then one) and being able to group these optional claims and in future attach a purpose to that request is a useful UX simplification.

Requests can shift from

I want your "first_name", "last_name" and "birth_date" and optionally your "resident_address", "resident_state" and "resident_country" attributes from your mDL.

To

I want "proof of your ID" which involves sharing the "first_name", "last_name" and "birth_date" from your mDL AND optionally I want "proof of your address" which involves sharing the "resident_address", "resident_state" and "resident_country" attributes from your mDL.

David-Chadwick:
I will argue that the revised reframing `I want "proof of your ID" which involves sharing the "first_name", "last_name" and "birth_date" from your mDL AND optionally I want "proof of your address" which involves sharing the "resident_address", "resident_state" and "resident_country" attributes from your mDL.` is insufficient on its own as the user has no idea why the optional attributes are being requested. Thus the reframing must include the purpose (i.e. extended service) that is being offered. 

I want "proof of your ID" which involves sharing the "first_name", "last_name" and "birth_date" from your mDL AND optionally I want "proof of your address" _so that we can post you a receipt_ which involves sharing the "resident_address", "resident_state" and "resident_country" attributes from your mDL.

In addition the original service could be included as the purpose for requesting the mandatory attributes, but this may not be necessary if we can assume that the user already knows the service that s/he is requesting.

examples/query_lang/claims_alternatives.json
```diff
@@ -19,11 +19,10 @@
         {"id": "f", "path": ["email"]}
       ],
       "claim_sets": [ // defines the rules
```
danielfett, line 21: @tplooker I think you wanted to remove claim_sets? For a fair comparison, please modify all examples so that it is not being used any longer.

examples/query_lang/credentials_alternatives.json
```diff
@@ -54,18 +54,27 @@
         "vct_values": ["https://company.example/company_rewards"]
       },
       "claims": [
-        {"id": "current", "path": ["rewards_number"]},
-        {"id": "legacy", "path": ["legacy_system_rewards_number"]}
+        {"id": "current", "path": ["rewards_number"]}
       ],
       "claim_sets": [
-        ["current", "legacy?"]
+        ["current"]
```
danielfett, line 60: As long as claim_sets is preserved, the `?` semantic is still here, but hidden. Either we remove claim_sets or we should keep `?` as an option.

examples/query_lang/complex_mdoc.json
```diff
@@ -0,0 +1,105 @@
...
+  "credentials": [
+    {
+      "id": "mdl-id",
+      "format": "mso_mdoc",
+      "meta": {
+        "doctype_values": ["org.iso.18013.5.1.mdl"]
+      },
+      "claims": [
+        {
+          "id": "given_name",
```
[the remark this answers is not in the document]
tlodderstedt: I agree.

tlodderstedt (changes requested):
I think the claim_set can and should be dropped. The combinations can be expressed on the ' credentials' level. That is sufficient and would simplify the syntax.

Sakurann:
Agreed on the WG call that in a new query language, we will:

1. remove claim_sets
2. introduce equivalent of ?! in credentials object (`provide_if_available` or something)
3. change `credential_sets` to array of objects 
4. add purpose to the object in credential_sets array - it can be a string or an integer, but it has to be shown only if RP is authenticated. (from processing, probably easier to defined purpose_code that is an integer separately? shouldn't matter, we can discuss in the PR)

for the sake of making progress, merging this PR and @danielfett will apply full set of these changes in his PR.

no consensus on value matching.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -714,12 +726,22 @@ The following rules apply for selecting claims via `claims` and `claim_sets`:
...
-  Claims postfixed by `?` are requested optionally, i.e., it is up to the Wallet
-  to send the claim in the response or not. Claims postfixed by `?!` have to be
-  provided by the Wallet if they exist in the Credential.
-
-If the Wallet cannot deliver all non-optional claims requested by the Verifier
+  `claim_sets`. Claims listed without a postfix are requested unconditionally. 
+  Claims postfixed by `?!` have to be provided by the Wallet if they exist on 
+  the matched Credential. The order of the options conveyed in the `claim_sets`
+  array expresses the Verifiers preference for what is returned. A wallet MUST 
+  prioritise returning the most preferred option.
```
danielfett, line 733:
```suggestion
  prioritize returning the most preferred option.
```

PR #294, opened 2024-10-23
Use dcql in #browser_api_request example rather than pe 

bc-pi:
Use dcql in #browser_api_request example rather than pe, which is an editorial change to a non-normative example that maybe kinda should have gone in https://github.com/openid/OpenID4VP/pull/254/files#r1811264551

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1987,7 +1987,7 @@ The following is a non-normative example of an unsigned OpenID4VP request (when 
   response_mode: "w3c_dc_api",
   nonce: "n-0S6_WzA2Mj",
   client_metadata: {...},
-  presentation_definition: {...}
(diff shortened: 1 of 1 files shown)


awoie (approved):

samuelgoto (approved):
I think this would be much appreciated by browser vendors ... much simpler to understand the `dcql_query` than PE ...

Any chance we can make this required, so that browser vendors don't have to parse/support both?

c2bo (approved):

jogu:
> I think this would be much appreciated by browser vendors ... much simpler to understand the `dcql_query` than PE ...
> 
> Any chance we can make this required, so that browser vendors don't have to parse/support both?

Can you open an issue for this please Sam, if we don't already have one? If we have sufficient implementation experience it's definitely something we could ask the working group about before 1.0.

jogu (approved):

PR #297, opened 2024-10-25
Update examples to reference the W3C VCDM v2, Data Integrity, and VC JOSE COSE

decentralgabe:
Fix #5 

Spec references:
* https://www.w3.org/TR/vc-data-model-2.0
* https://www.w3.org/TR/vc-jose-cose/
* https://www.w3.org/TR/vc-data-integrity/

There may be some small errors in the examples but I hope this enough to get started.


decentralgabe:
cc: @msporny @brentzundel @or13

OR13 (approved):

msporny (approved):
Overall, LGTM.

Non-blocking feedback... I wonder if the prefix should be "di" instead of "dip", we've never really used "dip" before, we tend to use "di".

/cc @dlongley @davidlehn

dlongley:
+1 and I agree that `di` is the more frequently used prefix.

examples/credentials/dip_vc.json
```diff
@@ -1,15 +1,16 @@
...
   "type": [
     "VerifiableCredential",
     "IDCredential"
   ],
   "issuer": {
-    "id": "did:example:issuer"
+    "id": "https://issuer.example/issuers/14",
+    "name": "Example Issuer"
   },
   "issuanceDate": "2010-01-01T19:23:24Z",
```
davidlehn, line 15:
```suggestion
  "validFrom": "2010-01-01T19:23:24Z",
```

David-Chadwick (approved):

examples/response/presentation_submission_multiple_vps.json
```diff
@@ -4,16 +4,16 @@
...
       "path_nested": {
-        "format": "ldp_vc",
+        "format": "di_vc",
         "path": "$.verifiableCredential[0]"
       }
     },
     {
       "id": "Example credential disclosing only address",
-      "format": "vc+sd-jwt",
+      "format": "sdjwt_vc",
```
jogu, line 16: Can you explain this change please? It appears to be out of scope for this issue and PR.

schema/request.json
```diff
@@ -30,7 +30,7 @@
     "format": {
       "type": "object",
       "patternProperties": {
-        "^jwt$|^jwt_vc$|^jwt_vp$": {
+        "^jwt$|^jose_vc$|^jose_vp$|^cose_vc$|^cose_vp$|^sdjwt_vc$|^sdjwt_vp$": {
```
jogu, line 33: Similar question as above about including `sdjwt_vc` and `sdjwt_vp` here.

jogu (changes requested):

schema/request.json
decentralgabe: updating to the VCDM v2 requires selecting a securing mechanism. one of those mechanisms is [VC JOSE COSE](https://www.w3.org/TR/vc-jose-cose/) which defines securing mechanisms for JOSE, COSE, and SD-JWT.

this change adds corresponding format properties for the new data model.

if you have a suggestion on how to break this into 2 PRs I'm open to it.

examples/response/presentation_submission_multiple_vps.json
decentralgabe: this one can be reverted it was a mistake

decentralgabe, line 16:
```suggestion
      "format": "vc+sd-jwt",
```

selfissued (approved):

decentralgabe:
For a myriad of reasons, I'm unable to make further changes to this PR for the foreseeable future. I would appreciate if someone else could pick it up on my behalf.

selfissued:
As I said on today's DCP call, I'm willing to own this PR while @decentralgabe needs to step away for a bit.

selfissued (approved):

schema/request.json
selfissued: @jogu does @decentralgabe's answer above satisfy you or do you want to request changes?  If you're satisfied, can you please approve?  Thanks.

bc-pi:
@OR13 had some thoughts related to this https://lists.openid.net/pipermail/openid-specs-digital-credentials-protocols/Week-of-Mon-20241028/000525.html

Sakurann (changes requested):
let's discuss in the issue #5 first how we want to transition from VCDM v1.1 to v.2.0. 

I originally opened that issue, but I do not believe we should overwrite v.1.1 text/examples in OID4VP as there are still implementations who use VCDM v1.1 as defined currently. 
if we want to support v2.0, we could add a separate section on VCDM 2.0, but we should discuss in a WG first

jogu:
As per previous comment, let's discuss this further in the issue - closing for now to keep the PR list showing PRs we want people to actively be looking at, we can reopen once we're ready to work on it again.

PR #306, opened 2024-11-05
change credential format references from the appendix of openid4vci to openid4vp

### LINKED ISSUE #256: vp_token definition referencing OpenID4VCI
The vp_token definition requires that the OpenID4VCI specification is followed for some of the details of the verifiable presentation content. This has a few issues. :

* It is strange that an issuance specification defines the details for how to do presentation, instead of specifying the presentation details in the presentation specification. 
* For some document formats, both the VCI spec as well Appendix B of OpenID4VP have requirements for the credential response. The current text explicitly says that OpenID4VCI must be followed if it specifies something for the credential response, this seems to be a contradiction to also having credential response requirements in OpenID4VP.
* For MSO_mdoc the Credential Response requirements in the OpenID4VCI spec do not make sense in the context of OpenID4VP, we should make it clear what exactly is meant with “If Appendix A of [OpenID.VCI] defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the vp_token response parameter.”

Can we remove the reference / mandate of using the OpenID4VCI specification for credential response encoding requirements?

c2bo:
fixes #256 

regarding this comment:

> I think removing the reference to VCI here and adding definitions for the contents of vp_token in Appendix B would be good, as Martijn notes this has already been done in the mdl section in Appendix B.

_Originally posted by @jogu in https://github.com/openid/OpenID4VP/issues/256#issuecomment-2341441861_
            

Not entirely sure if I got everything, but it looks to me like the vp_token is well defined in all of the appendix B sections, so I don't think there is anything else to do?


Sakurann (commented):
Interesting that the only reason we had VCI reference was appendix A

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2674,6 +2658,7 @@ The technology described in this specification was made available from contribut
 
    -22
 
+   * remove references to the openid4vci credential format section
```
jogu, line 2661: This should be in the -23 change history section (which isn't even showing up here, I suspect main needs to be merged into this PR first)

Sakurann: merging main probably does solve this. we can merge with a note to move this to -23 later and I can fix it in a PR later. a lot of PRs now open need change log anyway

```suggestion
   * remove references to the openid4vci credential format section (needs to be moved to -23)
```

awoie (approved):

danielfett (approved):
Somewhat concerned the changelog entry will forever live under the wrong version number :-D But otherwise looks good to me!

openid-4-verifiable-presentations-1_0.md
jogu: It seemed easiest to just move it, I've done that now - approving!

jogu (approved):

PR #308, opened 2024-11-05
Add Multi RP Credentials/Authentication capability

### LINKED ISSUE #248: How can verifiers that support multiple trust models/ecosystems know how to authenticate to the wallet?

tlodderstedt:
This PR changes the way requests are signed to the Digital Credentials API from compact serialization (single RP) to JWS JSON Serialization (Multiple RPs).
Questions to the WG:
- Do you want to keep (re-add) the JWS Compact Serialization to the Digital Credentials API profile?
- Do we want the Multi RP capability to the traditional OID4VP flow?

resolves #248

### DIFF
--- a/examples/digital_credentials_api/signed_request_payload.json
+++ b/examples/digital_credentials_api/signed_request_payload.json
@@ -6,5 +6,19 @@
   "response_type": "vp_token",
   "response_mode": "dc_api.jwt",
   "nonce": "n-0S6_WzA2Mj",
-  "dcql_query": {...}
+  "dcql_query": {...},
+  "client_metadata": {
+    "jwks": {
+      "keys": [
+        {
+          "kty": "EC",
+          "crv": "P-256",
+          "x": "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
+          "y": "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
+          "use": "enc",
+          "kid": "1"
+        }
+      ]
+    }
+  }
 }
--- a/examples/digital_credentials_api/signed_request_payload.json
+++ b/examples/digital_credentials_api/signed_request_payload.json
@@ -1,5 +1,4 @@
 {
-  "client_id": "https://client.example.org",
   "expected_origins": [
     "https://origin1.example.com",
     "https://origin2.example.com"
@@ -7,25 +6,5 @@
   "response_type": "vp_token",
   "response_mode": "dc_api.jwt",
   "nonce": "n-0S6_WzA2Mj",
-  "client_metadata": {
-    "vp_formats": {
-      "dc+sd-jwt": {
-        "sd-jwt_alg_values": [ "PS256" ],
-        "kb-jwt_alg_values": [ "PS256" ]
-      }
-    },
-    "jwks": {
-      "keys": [
-        {
-          "kty": "EC",
-          "crv": "P-256",
-          "x": "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
-          "y": "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
-          "use": "enc",
-          "kid": "1"
-        }
-      ]
-    }
-  },
-  "presentation_definition": {...}
+  "dcql_query": {...}
 }
--- a/examples/digital_credentials_api/signed_request_payload_compact.json
+++ b/examples/digital_credentials_api/signed_request_payload_compact.json
@@ -1,0 +1,25 @@
+{
+  "expected_origins": [
+    "https://origin1.example.com",
+    "https://origin2.example.com"
+  ],
+  "client_id": "x509_san_dns:rp.example.com",
+  "client_metadata": {
+    "jwks": {
+      "keys": [
+        {
+          "kty": "EC",
+          "crv": "P-256",
+          "x": "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
+          "y": "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
+          "use": "enc",
+          "kid": "1"
+        }
+      ]
+    }
+  },
+  "response_type": "vp_token",
+  "response_mode": "w3c_dc_api.jwt",
+  "nonce": "n-0S6_WzA2Mj",
+  "dcql_query": {...}
+}
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2275,7 +2275,7 @@ The signed request allows the Wallet to authenticate the Verifier using one or m
 
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
+The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a Wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
 
 The following request parameters MUST be present in the protected header of the respective signature object: 
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2296,7 +2296,6 @@ The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Clie
 In this case, the following request parameters MUST be present in the protected header of the respective `signature` object in the `signatures` array defined in [@!RFC7515, section 7.2.1]:
 
 * `client_id`
-* `client_metadata`
 
 All other request parameters MUST be present in the `payload` element of the JWS object.
 
@@ -2329,21 +2328,7 @@ This is an example of a protected header:
     "MIICOjCCAeG...djzH7lA==",
     "MIICLTCCAdS...koAmhWVKe"
   ],
-  "client_id": "x509_san_dns:rp.example.com",
-  "client_metadata": {
-    "jwks": {
-      "keys": [
-        {
-          "kty": "EC",
-          "crv": "P-256",
-          "x": "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
-          "y": "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
-          "use": "enc",
-          "kid": "1"
-        }
-      ]
-    }
-  }
+  "client_id": "x509_san_dns:rp.example.com"
 }
 ```
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2325,7 +2325,6 @@ This is an example of a protected header:
 ```
 {
   "alg": "ES256",
-  "kid": "k2bdc",
   "x5c": [
     "MIICOjCCAeG...djzH7lA==",
     "MIICLTCCAdS...koAmhWVKe"
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2300,7 +2300,7 @@ This is an example of the payload of a signed OpenID4VP request used with the W3
 
 #### JWS JSON Serialization
 
-The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata for the same request. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
 
 In this case, the following request parameters MUST be present in the protected header of the respective signature object: 
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2285,11 +2285,11 @@ The following request parameters MUST be present in the protected header of the 
 Below is a non-normative example of such a request:
 
 ```js
-const credential = await navigator.identity.get({
+const credential = await navigator.credentials.get({
   digital: {
     providers: [{
       protocol: "openid4vp",
-      request: {
+      data: {
         "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
         "signatures": [
         {
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2303,26 +2303,19 @@ All other request parameters MUST be present in the `payload` element of the JWS
 Below is a non-normative example of such a request:
 
 ```js
-const credential = await navigator.credentials.get({
-  digital: {
-    providers: [{
-      protocol: "openid4vp",
-      data: {
-        "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
-        "signatures": [
-        {
-          "protected": "eyJhbGciOiAiRVMyNT..MiLCJraWQiOiAiMSJ9XX19fQ",
-          "signature": "PFwem0Ajp2Sag...T2z784h8TQqgTR9tXcif0jw"
-        },
-        {
-          "protected": "eyJhbGciOiAiRVMyNTY...tpZCI6ICIxIn1dfX19",
-          "signature": "irgtXbJGwE2wN4Lc...2TvUodsE0vaC-NXpB9G39cMXZ9A"
-        }
-      ]
-     }
-    }]
-  }
-});
+{
+  "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
+  "signatures": [
+    {
+      "protected": "eyJhbGciOiAiRVMyNT..MiLCJraWQiOiAiMSJ9XX19fQ",
+      "signature": "PFwem0Ajp2Sag...T2z784h8TQqgTR9tXcif0jw"
+    },
+    {
+      "protected": "eyJhbGciOiAiRVMyNTY...tpZCI6ICIxIn1dfX19",
+      "signature": "irgtXbJGwE2wN4Lc...2TvUodsE0vaC-NXpB9G39cMXZ9A"
+    }
+  ]
+}
 ```
 
 Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2334,7 +2334,7 @@ The following is a non-normative example of a content of a decoded protected hea
 }
 ```
 
-This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS JSON Serialization:
+The following is a non-normative example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS JSON Serialization:
 
 <{{examples/digital_credentials_api/signed_request_payload.json}} 
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2323,7 +2323,7 @@ Every object in the `signatures` structure contains the parameters and the signa
 
 The following is a non-normative example of a content of a decoded protected header:
 
-```
+```json
 {
   "alg": "ES256",
   "x5c": [
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -3079,6 +3079,22 @@ established by [@!RFC7515].
 * Change Controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
 * Specification Document(s): (#verifier_attestation_jwt) of this specification
 
+### client_id
+
+* Header Parameter Name: `client_id`
+* Header Parameter Description: This header contains a Client Identifier. 
+* Header Parameter Usage Location: JWS
+* Change Controller: IETF
+* Specification Document(s): [@!RFC6749]
+
+### client_metadata
+
+* Header Parameter Name: `jwt`
+* Header Parameter Description: This header contains a JWT. Processing rules MAY depend on the `typ` header value of the respective JWT. 
+* Header Parameter Usage Location: JWS
+* Change Controller: OpenID Foundation Digital Credentials Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
+* Reference: (#vp_token_request) of this specification
+
 ## Uniform Resource Identifier (URI) Schemes Registry
 
 This specification registers the following URI scheme
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -3055,14 +3055,6 @@ established by [@!RFC7515].
 * Change Controller: IETF
 * Specification Document(s): [@!RFC6749]
 
-### client_metadata
-
-* Header Parameter Name: `client_metadata`
-* Header Parameter Description: A JSON object containing the Verifier metadata values as defined in (#new_parameters).
-* Header Parameter Usage Location: JWS
-* Change Controller: OpenID Foundation Digital Credentials Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
-* Specification Document(s): (#vp_token_request) of this specification
-
 ## Uniform Resource Identifier (URI) Schemes Registry
 
 This specification registers the following URI scheme
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -3089,11 +3089,11 @@ established by [@!RFC7515].
 
 ### client_metadata
 
-* Header Parameter Name: `jwt`
-* Header Parameter Description: This header contains a JWT. Processing rules MAY depend on the `typ` header value of the respective JWT. 
+* Header Parameter Name: `client_metadata`
+* Header Parameter Description: A JSON object containing the Verifier metadata values as defined in (#new_parameters).
 * Header Parameter Usage Location: JWS
 * Change Controller: OpenID Foundation Digital Credentials Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
-* Reference: (#vp_token_request) of this specification
+* Specification Document(s): (#vp_token_request) of this specification
 
 ## Uniform Resource Identifier (URI) Schemes Registry
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2275,7 +2275,7 @@ The signed request allows the Wallet to authenticate the Verifier using one or m
 
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a wallet. This is to serve use cases where the Verifier requests credentials belong to different transt framework and thus must authenticate in the context of those trust frameworks.
+The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
 
 The following request parameter MUST be placed in the protect header of the respective signature object: 
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2279,7 +2279,9 @@ Verifiers can format signed Requests either using JWS Compact Serialization or J
 
 #### JWS Compact Serialization
 
-In case of the JWS Compact Serialization, all request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as the value of the `request` claim in the `data` element of the API call. This is illustated in the following example.
+When the JWS Compact Serialization is used to send the request, the Verifier can convey only one Trust Framework, i.e., the Verifier should know which trust frameworks the wallet supports. All request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as the value of the `request` claim in the `data` element of the API call. 
+
+This is illustrated in the following non-normative example.
 
 ```js
 { request: "eyJhbGciOiJF..." }
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2321,7 +2321,7 @@ Below is a non-normative example of such a request:
 
 Every object in the `signatures` structure contains the parameters and the signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
 
-This is an example of a protected header:
+The following is a non-normative example of a content of a decoded protected header:
 
 ```
 {
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2307,7 +2307,7 @@ const credential = await navigator.identity.get({
 });
 ```
 
-Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier.
+Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
 
 This is an example of a protected header:
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2318,7 +2318,7 @@ Below is a non-normative example of such a request:
 }
 ```
 
-Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
+Every object in the `signatures` structure contains the parameters and the signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
 
 This is an example of a protected header:
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2275,12 +2275,39 @@ The signed request allows the Wallet to authenticate the Verifier using one or m
 
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a Wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
+Verifiers can format signed Requests either using JWS Compact Serialization or JWS Serialization [@!RFC7515]). 
 
-The following request parameters MUST be present in the protected header of the respective signature object: 
+#### JWS Compact Serialization
+
+In case of the JWS Compact Serialization, all request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as value of the `request` claim in the `data` element of the API call. This is illustated in the following example.
+
+```js
+const credential = await navigator.credentials.get({
+  digital: {
+    providers: [{
+      protocol: "openid4vp",
+      data: {
+        { request: "eyJhbGciOiJF..." }
+     }
+    }]
+  }
+});
+```
+
+This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS Compact Serialization:
+
+<{{examples/digital_credentials_api/signed_request_payload_compact.json}} 
+
+#### JWS JSON Serialization
+
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata for the same request. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
+
+In this case, the following request parameters MUST be present in the protected header of the respective signature object: 
 
 * `client_id`
 * `client_metadata`
+
+All other request parameters MUST be present in the `payload` element of the JWS object.
 
 Below is a non-normative example of such a request:
 
@@ -2293,11 +2320,11 @@ const credential = await navigator.credentials.get({
         "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
         "signatures": [
         {
-          "protected": "eyJhbGciOiJFUzI1NiJ9",
+          "protected": "eyJhbGciOiAiRVMyNT..MiLCJraWQiOiAiMSJ9XX19fQ",
           "signature": "PFwem0Ajp2Sag...T2z784h8TQqgTR9tXcif0jw"
         },
         {
-          "protected": "eyJhbGciOiJFUzI1NiJ9",
+          "protected": "eyJhbGciOiAiRVMyNTY...tpZCI6ICIxIn1dfX19",
           "signature": "irgtXbJGwE2wN4Lc...2TvUodsE0vaC-NXpB9G39cMXZ9A"
         }
       ]
@@ -2337,7 +2364,7 @@ This is an example of a protected header:
 }
 ```
 
-This is an example of the payload of a signed OpenID4VP request used with the DC API:
+This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS JSON Serialization:
 
 <{{examples/digital_credentials_api/signed_request_payload.json}} 
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -3052,7 +3052,7 @@ established by [@!RFC7515].
 ### client_id
 
 * Header Parameter Name: `client_id`
-* Header Parameter Description: This header contains a Client Identifier. 
+* Header Parameter Description: This header contains a Client Identifier. A Client Identifier is used in OAuth to identify a certain client. It is defined in [@!RFC6749], section 2.2.
 * Header Parameter Usage Location: JWS
 * Change Controller: IETF
 * Specification Document(s): [@!RFC6749]
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2291,7 +2291,7 @@ This is an example of the payload of a signed OpenID4VP request used with the W3
 
 #### JWS JSON Serialization
 
-The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests Credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
 
 In this case, the following request parameters MUST be present in the protected header of the respective `signature` object in the `signatures` array defined in [@!RFC7515, section 7.2.1]:
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2282,16 +2282,7 @@ Verifiers can format signed Requests either using JWS Compact Serialization or J
 In case of the JWS Compact Serialization, all request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as the value of the `request` claim in the `data` element of the API call. This is illustated in the following example.
 
 ```js
-const credential = await navigator.credentials.get({
-  digital: {
-    providers: [{
-      protocol: "openid4vp",
-      data: {
-        { request: "eyJhbGciOiJF..." }
-     }
-    }]
-  }
-});
+{ request: "eyJhbGciOiJF..." }
 ```
 
 This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS Compact Serialization:
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2293,7 +2293,7 @@ This is an example of the payload of a signed OpenID4VP request used with the W3
 
 The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
 
-In this case, the following request parameters MUST be present in the protected header of the respective `signature` object: 
+In this case, the following request parameters MUST be present in the protected header of the respective `signature` object in the `signatures` array defined in [@!RFC7515, section 7.2.1]:
 
 * `client_id`
 * `client_metadata`
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2277,7 +2277,7 @@ The signed Request Object MAY contain all the parameters listed in (#browser_api
 
 The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
 
-The following request parameter MUST be placed in the protect header of the respective signature object: 
+The following request parameters MUST be present in the protected header of the respective signature object: 
 
 * `client_id`
 * `client_metadata`
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2275,7 +2275,7 @@ The signed request allows the Wallet to authenticate the Verifier using one or m
 
 The signed Request Object MAY contain all the parameters listed in (#dc_api_request), except `request`.
 
-Verifiers can format signed Requests either using JWS Compact Serialization or JWS Serialization [@!RFC7515]). 
+Verifiers can format signed Requests either using JWS Compact Serialization or JWS JSON Serialization [@!RFC7515]). 
 
 #### JWS Compact Serialization
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2273,7 +2273,7 @@ The Verifier MAY send a signed request, for example, when identification and aut
 
 The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) in addition to the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
 
-The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
+The signed Request Object MAY contain all the parameters listed in (#dc_api_request), except `request`.
 
 Verifiers can format signed Requests either using JWS Compact Serialization or JWS Serialization [@!RFC7515]). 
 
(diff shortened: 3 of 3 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
 
+The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
+
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-Below is a non-normative example of such a request sent over the W3C Digital Credentials API:
+The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a wallet. This is to serve use cases where the Verifier requests credentials belong to different transt framework and thus must authenticate in the context of those trust frameworks.
```
charsleysa, line 2032: typos
```suggestion
The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
```

peppelinux: wallet -> Wallet

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
+The following request parameter MUST be placed in the protect header of the respective signature object:
```
c2bo, line 2034:
```suggestion
The following request parameters MUST be present in the protected header of the respective signature object: 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+        }
+      ]
+     }
+    }]
+  }
+});
 ```
 
-This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API:
+Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier.
```
c2bo, line 2064: Not sure if we need this, but I think it helps to reference to the JWS RFC
```suggestion
Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
```

selfissued (changes requested):
The JWS Compact Serialization is supported by essentially all implementations, whereas many don't support the JSON Serialization, in part, because JWTs don't use it.

I believe switching to the JSON Serialization would present an unnecessary burden to developers and harm interoperability.

(For what it's worth, I believe that adding the JSON Serialization a dozen or so years ago at all was a mistake, but I realize that that's water under the bridge.)

David-Chadwick:
As I pointed out in the issue that this PR purports to resolve: when there is a trust infrastructure in place the request does not need to be signed as the wallet can determine the authenticity of the RP using the trust infrastructure. So we do not need multiple signatures and JSON serialisation.

tlodderstedt:
> As I pointed out in the issue that this PR purports to resolve: when there is a trust infrastructure in place the request does not need to be signed as the wallet can determine the authenticity of the RP using the trust infrastructure. So we do not need multiple signatures and JSON serialisation.

The profile supports unsigned requests. Would that be appropriate for your use cases?

David-Chadwick:
@Sakurann What you say is only one way of evaluating trust. The trust infrastructure can equally well decide if the RP is trustworthy or not based on its asserted name or ID. No keys are needed for this. If the asserted name/ID is deemed to be untrustworthy by the trust infrastructure, then the wallet knows not to trust the RP. If the name/ID is deemed to be trustworthy then there are two cases. 1. The RP is giving its true name, or 2.  the RP is masquerading as a trustworthy RP. The latter can be protected against by the trust infrastructure providing the return URI of the RP for the wallet's response, and then the masquerading RP will not glean any information from the wallet. This is what we implemented. To prevent a DOS attack on the real trustworthy RP, we could add an extra protection feature, namely, the RP provides a random number on its request, which forms the query parameter on the wallet's reply to the return URI. The RP checks this parameter and if it is not recognised throws the wallet's response away.

David-Chadwick:
@tlodderstedt Yes, not requiring the RP's request to be signed solves the use case, and it also solves the problem of requiring multiple signatures to be on the request, because no signatures are actually needed. The RP can provide a set of names/IDs, one for each trust infrastructure that it is a member of.  Then the wallet can match the presented trust infrastructures to the ones it is a member of, and where there is an intersection find out if the RP is trustworthy or not.

tlodderstedt:
> The JWS Compact Serialization is supported by essentially all implementations, whereas many don't support the JSON Serialization, in part, because JWTs don't use it.
> 
> I believe switching to the JSON Serialization would present an unnecessary burden to developers and harm interoperability.
> 
> (For what it's worth, I believe that adding the JSON Serialization a dozen or so years ago at all was a mistake, but I realize that that's water under the bridge.)

@selfissued The requirement we are asked to fulfill is to allow for multiple RP credentials being proven (through signatures) in a single request. We have extensively discussed this requirement in the hybrid workshop and agreed to support it. The design in the PR is, in my opinion, the simplest and most robust solution we can do it (alternative proposals involved canonicalization). So from my standpoint, what we can do is to also continue to support compact serialization. That's why I asked that question further up.

examples/digital_credentials_api/signed_request_payload.json
```diff
@@ -1,31 +1,10 @@
...
-          "x": "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
-          "y": "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
-          "use": "enc",
-          "kid": "1"
-        }
-      ]
-    }
-  },
-  "presentation_definition": {...}
+  "dcql_query": {...}
```
selfissued, line 9: Why were the `client_id`, `client_metadata`, and `jwks` deleted?  This change seems unrelated to the purpose of this PR.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
 
 The Verifier MAY send a signed request, for example, when identification and authentication of the Verifier is required.
 
+The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
```
selfissued, line 2028: How does the signed request enable the use of trust frameworks representation using OpenID Federation?  (It may already do that, but more details to that effect would be useful to implementers.)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-Below is a non-normative example of such a request sent over the W3C Digital Credentials API:
+The signed Request MUST use the JWS Serialization [@!RFC7515]). This allows the Verifier multiple Client Identifiers and corresponding key material and metadata when invoking a Wallet. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
```
selfissued, line 2032: We should discuss whether to always use the JSON Serialization or only to use it when there are multiple signatures.  (In this case, I realize that the "Keep simple things simple but make complex things possible" adage cuts both ways.)  I'd rather keep the simple case simple.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
+The following request parameters MUST be present in the protected header of the respective signature object: 
+
+* `client_id`
+* `client_metadata`
```
selfissued, line 2037: These aren't defined header parameters, so this is weird.  Why aren't these being sent as claims in the request?

examples/digital_credentials_api/signed_request_payload.json
tlodderstedt: This is indeed related to this PR, as this data elements are RP specific. Consequently, the PR moves them into the RP credential specific elements of the request.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: The PR does not change the way client id schemes work. Consequently, all definitions for federation still hold. true.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: Thanks for your feedback (that's what I asked for in the description of the PR).

openid-4-verifiable-presentations-1_0.md
tlodderstedt: because they are specific to a certain RP credential

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,75 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
-{ request: "eyJhbGciOiJF..." }
+const credential = await navigator.credentials.get({
+  digital: {
+    providers: [{
+      protocol: "openid4vp",
+      data: {
+        "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
+        "signatures": [
+        {
+          "protected": "eyJhbGciOiJFUzI1NiJ9",
```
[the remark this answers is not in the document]
tlodderstedt: not sure what you mean, see an header example (with kid and alg) below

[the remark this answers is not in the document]
selfissued: `eyJhbGciOiJFUzI1NiJ9` decodes to `{"alg":"ES256"}`, which has no `kid`.

bc-pi:
> Do you want to keep (re-add) the JWS Compact Serialization to the Digital Credentials API profile?

Yes 

> Do we want the Multi RP capability to the traditional OID4VP flow?

No

tlodderstedt:
WG call: the sense of the "room" is we should not remove compact serialization. No support for adding multiple RP authentication feature to the HTTPS flow.

examples/digital_credentials_api/signed_request_payload.json
selfissued: We should discuss where the `client_id`, `client_metadata`, and `jwks` claims should go when using the Compact Serialization.  These would normally be claims - not header parameters.

I understand the reasoning for making them header parameters when using the JSON Serialization, but in my view, that's not the normal case, and we should make the normal case natural.

Sakurann:
WG discussion:
- question was asked if this approach is too much given the direction ISO SC17 WG10 has chosen? It was clarified that 18013-5 1st edition and 18013-7 2nd edition meet this requirement by sending multiple requests, but 18013-5 2nd edition would allows to meet this requirement by sending one request with multiple signatures, so the outcome of the discussion was that the approach in this PR is still very much needed.

selfissued (approved):
Thanks for enabling the use of the JWS Compact Serialization.

hlozi:
Clarification: The 18013-5 second edition supports multiple signatures for a single request. The 18013-7 annex C supports this feature because it doesn’t define the request but it is based on the 18013-5 request.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+const credential = await navigator.credentials.get({
+  digital: {
+    providers: [{
+      protocol: "openid4vp",
+      data: {
+        { request: "eyJhbGciOiJF..." }
+     }
+    }]
+  }
+});
```
c2bo, line 2048: I don't think this change is intended? Do we just need to rebase on main?

leecam (approved):
The proposed design looks good to me.

hlozi (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
 
 The Verifier MAY send a signed request, for example, when identification and authentication of the Verifier is required.
 
+The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
```
awoie, line 2028: IMO, for signed requests, the wallet would still use the Web PKI or origin model. IMO, it is unsigned requests < signed requests.

```suggestion
The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) in addition to the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-Below is a non-normative example of such a request sent over the W3C Digital Credentials API:
+Verifiers can format signed Requests either using JWS Compact Serialization or JWS Serialization [@!RFC7515]). 
+
+#### JWS Compact Serialization
+
+In case of the JWS Compact Serialization, all request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as value of the `request` claim in the `data` element of the API call. This is illustated in the following example.
```
awoie, line 2036:
```suggestion
In case of the JWS Compact Serialization, all request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as the value of the `request` claim in the `data` element of the API call. This is illustated in the following example.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
 ```
 
-This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API:
+This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS Compact Serialization:
+
+<{{examples/digital_credentials_api/signed_request_payload_compact.json}} 
+
+#### JWS JSON Serialization
+
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata for the same request. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
```
awoie, line 2057:
```suggestion
The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, must authenticate in the context of those trust frameworks.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
+In this case, the following request parameters MUST be present in the protected header of the respective signature object:
```
awoie, line 2059:
```suggestion
In this case, the following request parameters MUST be present in the protected header of the respective `signature` object: 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+<{{examples/digital_credentials_api/signed_request_payload_compact.json}} 
+
+#### JWS JSON Serialization
+
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata for the same request. This is to serve use cases where the Verifier requests credentials belonging to different trust frameworks and thus must authenticate in the context of those trust frameworks.
+
+In this case, the following request parameters MUST be present in the protected header of the respective signature object: 
+
+* `client_id`
+* `client_metadata`
```
awoie, line 2062: Since these become protected headers, do we need an IANA registration for this @selfissued ?

awoie (approved):
Generally looks good!

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+const credential = await navigator.credentials.get({
+  digital: {
+    providers: [{
+      protocol: "openid4vp",
+      data: {
+        { request: "eyJhbGciOiJF..." }
+     }
+    }]
+  }
+});
```
c2bo, line 2048: per #254 I think we should keep only the payload in these examples?
```suggestion
{ request: "eyJhbGciOiJF..." }
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,102 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+        },
+        {
+          "protected": "eyJhbGciOiAiRVMyNTY...tpZCI6ICIxIn1dfX19",
+          "signature": "irgtXbJGwE2wN4Lc...2TvUodsE0vaC-NXpB9G39cMXZ9A"
+        }
+      ]
+     }
+    }]
+  }
+});
```
c2bo, line 2088: same as above (#254)?

```suggestion
{
  "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
  "signatures": [
    {
      "protected": "eyJhbGciOiAiRVMyNT..MiLCJraWQiOiAiMSJ9XX19fQ",
      "signature": "PFwem0Ajp2Sag...T2z784h8TQqgTR9tXcif0jw"
    },
    {
      "protected": "eyJhbGciOiAiRVMyNTY...tpZCI6ICIxIn1dfX19",
      "signature": "irgtXbJGwE2wN4Lc...2TvUodsE0vaC-NXpB9G39cMXZ9A"
    }
  ]
}
```

c2bo (commented):
#254 removed the DC API specific parts of the examples in favour of leaving only the data element - not sure if we want to re-introduce those here

openid-4-verifiable-presentations-1_0.md
selfissued: Yes, if you're going to create new header parameters, you should register them.

tlodderstedt: added IANA registration

**https://github.com/openid/OpenID4VP/pull/308/commits/e2fe62eea5b458e1cd6e402319a3db9db68728e7**

tlodderstedt:
@leecam @hlozi @c2bo 
In a discussion with @awoie we discovered two challenges: 

1. If the RP sends a request with multiple client identifiers, which one is used in the apu/apv header of an encrypted response? This is related to issue #380 (@bc-pi )
2. If the RP sends a request with multiple client identifiers for one or more mdocs, how is it supposed to figure out which client identifier to use for checking the session transcript in the respective device response? This is related to PR #374 (@awoie)

hlozi:
> @leecam @hlozi @c2bo
> In a discussion with @awoie we discovered two challenges:
> 
> If the RP sends a request with multiple client identifiers, which one is used in the apu/apv header of an encrypted response? This is related to issue https://github.com/openid/OpenID4VP/pull/380 (@bc-pi )
> If the RP sends a request with multiple client identifiers for one or more mdocs, how is it supposed to figure out which client identifier to use for checking the session transcript in the respective device response? This is related to PR https://github.com/openid/OpenID4VP/pull/374 (@awoie)

For 1), currently, the intention is to have APV/APU transmitted to the verifier (I understand there is no current way of doing it otherwise), the verifier needs to check the value of the client ID coming in the apv/apu in a way that makes sense (is it is the one of those in the params or a prefixed origin), and then it’s used in the decryption process. However, if we switch to a detached mode in the future, we may face the problem. one solution is when we do detached in the future we don't use client id in the KDF.

For 2), I believe the solution is to remove the client ID from the session transcript.  Is there a risk that is mitigated by including the client ID and that is not already mitigated by including origin and nonce? If not, let’s remove it.

awoie:
> @leecam @hlozi @c2bo In a discussion with @awoie we discovered two challenges:
> 
> 1. If the RP sends a request with multiple client identifiers, which one is used in the apu/apv header of an encrypted response? This is related to issue [Mandate the use of apu/apv in the JWE header of OpenID4VP encrypted responses #380](https://github.com/openid/OpenID4VP/pull/380) (@bc-pi )
> 2. If the RP sends a request with multiple client identifiers for one or more mdocs, how is it supposed to figure out which client identifier to use for checking the session transcript in the respective device response? This is related to PR [add session transcript for browser api #374](https://github.com/openid/OpenID4VP/pull/374) (@awoie)

There is a third issue with this PR regarding encryption. Since the client_metadata is specific to the signature object—or more specifically, conveyed in the protected header—there might be multiple client_metadata objects with varying encryption keys. Guidance is needed on which encryption key to use if multiple are provided. Note that there is only one JWE.

Options to address this:

- Make the encryption key a general request parameter included in the signed payload to avoid the confusion of having multiple encryption keys.
- Use a multi-recipient JWE for the encrypted response, encrypting the response with the encryption key specified in the respective `client_metadata` of the `signature` object. This approach would also address issue 1), as each recipient would have their own protected header (`apu`/`apv` values). Additionally, it could resolve the detached AAD issue, as JWE JSON serialization supports a dedicated AAD fragment.

awoie:
> > @leecam @hlozi @c2bo
> > In a discussion with @awoie we discovered two challenges:
> > If the RP sends a request with multiple client identifiers, which one is used in the apu/apv header of an encrypted response? This is related to issue #380 (@bc-pi )
> > If the RP sends a request with multiple client identifiers for one or more mdocs, how is it supposed to figure out which client identifier to use for checking the session transcript in the respective device response? This is related to PR #374 (@awoie)
> 
> For 1), currently, the intention is to have APV/APU transmitted to the verifier (I understand there is no current way of doing it otherwise), the verifier needs to check the value of the client ID coming in the apv/apu in a way that makes sense (is it is the one of those in the params or a prefixed origin), and then it’s used in the decryption process. However, if we switch to a detached mode in the future, we may face the problem. one solution is when we do detached in the future we don't use client id in the KDF.

If we switch to a detached mode in the future, the verifier would still need to understand which `client_id` was used in the `SessionTranscript`. The verifier could certainly test each provided `client_id` but that is a solution that is awkward.

> For 2), I believe the solution is to remove the client ID from the session transcript. Is there a risk that is mitigated by including the client ID and that is not already mitigated by including origin and nonce? If not, let’s remove it.

I agree that removing `client_id` from the `SessionTranscript` would probably solve the issue. If the web and mobile platform is trusted, I personally don't see an issue with this since the wallet will obtain the origin from a trusted source which is then validated by the verifier. I don't know what attack would be mitigated by including the `client_id` in the verifiable presentation in this specific case. I remember we had a discussion where the verifier wanted to know what trust framework the wallet used to verify the request. Perhaps @danielfett has some input on this?

tlodderstedt:
What if we say that in case of multiple RP authentication, the encryption and session transcript functions always use a certain client id and dedicated data in the response, e.g. the first object in the 'signatures' array?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2642,6 +2709,22 @@ established by [@!RFC7515].
...
+* Header Parameter Name: `client_id`
+* Header Parameter Description: This header contains a Client Identifier. 
+* Header Parameter Usage Location: JWS
+* Change Controller: IETF
+* Specification Document(s): [@!RFC6749]
+
+### client_metadata
+
+* Header Parameter Name: `jwt`
+* Header Parameter Description: This header contains a JWT. Processing rules MAY depend on the `typ` header value of the respective JWT.
```
bc-pi, line 2723: need to fix copy/paste

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
 
-The signed request allows the Wallet to authenticate the Verifier using a trust framework other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
+This is an example of a protected header:
+
+```
+{
+  "alg": "ES256",
+  "kid": "k2bdc",
+  "x5c": [
```
bc-pi, line 2083: why would you have a kid (with my initials noless) alongside x5c?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
 
 The Verifier MAY send a signed request, for example, when identification and authentication of the Verifier is required.
 
+The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) in addition to the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
+
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-Below is a non-normative example of such a request sent over the W3C Digital Credentials API:
+Verifiers can format signed Requests either using JWS Compact Serialization or JWS Serialization [@!RFC7515]).
```
bc-pi, line 2032: There is no such thing as "JWS Serialization"
```suggestion
Verifiers can format signed Requests either using JWS Compact Serialization or JWS JSON Serialization [@!RFC7515]). 
```

openid-4-verifiable-presentations-1_0.md
Sakurann: WG discussion to keep compact serialization https://github.com/openid/OpenID4VP/pull/308#issuecomment-2520863682

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
 The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
 
-Below is a non-normative example of such a request sent over the W3C Digital Credentials API:
+Verifiers can format signed Requests either using JWS Compact Serialization or JWS Serialization [@!RFC7515]). 
+
+#### JWS Compact Serialization
+
+In case of the JWS Compact Serialization, all request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as the value of the `request` claim in the `data` element of the API call. This is illustated in the following example.
```
Sakurann, line 2036: I would add a sentence here that clarifies that JWS Compact serialization can convey only one trust framework i.e. Verifier should know which trust framework the wallet supports

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
 ```
 
-This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API:
+This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS Compact Serialization:
+
+<{{examples/digital_credentials_api/signed_request_payload_compact.json}} 
+
+#### JWS JSON Serialization
+
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
```
Sakurann, line 2048:
```suggestion
Using JWS JSON Serialization [@!RFC7515]) to send the Request allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.

This is achieved by passing multiple signatures of the same paylod in the protected header.
```

this is just a suggestion, but please provide an overview of the solution here. otherwise there is a jump between this short description and the technical details below..

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
When the JWS Compact Serialization is used to send the request, the Verifier can convey only one Trust Framework, i.e., the Verifier should know which trust frameworks the wallet supports. All request parameters are encoded in a request object as defined in (#vp_token_request) and the JWS object is used as the value of the `request` claim in the `data` element of the API call. 

This is illustrated in the following non-normative example.
```

openid-4-verifiable-presentations-1_0.md
Sakurann, line 2723:
```suggestion
* Header Parameter Name: `client_metadata`
* Header Parameter Description: This header contains Client Metadata.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+    {
+      "protected": "eyJhbGciOiAiRVMyNTY...tpZCI6ICIxIn1dfX19",
+      "signature": "irgtXbJGwE2wN4Lc...2TvUodsE0vaC-NXpB9G39cMXZ9A"
+    }
+  ]
+}
+```
 
-<{{examples/digital_credentials_api/signed_request_payload.json}}
+Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
```
Sakurann, line 2075: I don't understand how this is JWE RFC7516 compliant..?

Sakurann (changes requested):
- `signature` object need to be defined.
- descriptions need to be a little bit more detailed/reader-friendly

openid-4-verifiable-presentations-1_0.md
Sakurann, line 2075: it would really help readability if this is moved before the content of the `signature` is described....

Sakurann:
```suggestion
```

Sakurann, line 2075:
```suggestion
Every object in the `signatures` structure contains the parameters and the signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
```
i believe otherwise the sentence does not make much sense..

Sakurann:
Why an't we use JWE protected header to indicate which client_id is being used in apv/apu, session transcript and which key is used to decrypt?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2013,19 +2013,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+  "alg": "ES256",
+  "kid": "k2bdc",
+  "x5c": [
+    "MIICOjCCAeG...djzH7lA==",
+    "MIICLTCCAdS...koAmhWVKe"
+  ],
+  "client_id": "x509_san_dns:rp.example.com",
+  "client_metadata": {
+    "jwks": {
+      "keys": [
```
bc-pi, line 2078: I know there's other related commentary so please excuse any redundancy but it seems pretty awkward to have what are probably ephemeral encryption keys in each of the headers.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2677,6 +2744,22 @@ established by [@!RFC7515].
 * Change Controller: OpenID Foundation Artifact Binding Working Group - openid-specs-ab@lists.openid.net
 * Specification Document(s): (#verifier_attestation_jwt) of this specification
```
bc-pi, line 2746: I am not a DE on the JSON Web Signature and Encryption Header Parameters Registry so I can only speculate but if I were, I'd likely have questions...

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2013,19 +2013,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
 
 The Verifier MAY send a signed request, for example, when identification and authentication of the Verifier is required.
 
-The signed Request Object MAY contain all the parameters listed in (#dc_api_request), except `request`.
+The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) in addition to the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
 
-Below is a non-normative example of such a request sent over the Digital Credentials API (DC API):
+The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
```
c2bo, line 2018:
```suggestion
The signed Request Object MAY contain all the parameters listed in (#dc_api_request), except `request`.
```

awoie:
> What if we say that in case of multiple RP authentication, the encryption and session transcript functions always use a certain client id and dedicated data in the response, e.g. the first object in the 'signatures' array?

Since this PR (i.e. support for multi RP auth) is only applicable to DC API and not to vanilla OID4VP, this should be fine although it seems a bit awkward. Depending on future platform APIs, this might lead to confusions in wallet implementations under the assumption that the mobile platform provides only selected sub-requests to a wallet if multiple wallets are involved to answer all sub-requests. 

IMO, there is also some guidance needed on what information to display to the user if multiple Client IDs are provided?

However, it raises the question why the `client_id` has to be included at all in the verifiable presentation for audience binding (inclusion in `SessionTranscript` for mdocs/ `aud` for SD-JWT VCs/ `domain` for W3C VCs) and in the `apv`/`apu` for the DC API since the origin is asserted by the web/app platform and validated by the verifier.

hlozi:
In my opinion, removing the client ID from the session transcript (and the equivalent for other formats) does not affect the value added by the client ID in signed requests. But, it may complicate matters not only for this PR but also for future ones. We might end up having the RP try multiple values to verify the signature, which is really not desirable situation. So, we should consider removing the client ID from the session transcript but keeping the origin and nonce.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: fixed

openid-4-verifiable-presentations-1_0.md
tlodderstedt: removed it

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2025,19 +2025,86 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+    }
+  ]
+}
+```
 
-<{{examples/digital_credentials_api/signed_request_payload.json}}
+Every `signature` object in the structure contains the parameters and signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
 
-The signed request allows the Wallet to authenticate the Verifier using a trust framework other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
+This is an example of a protected header:
```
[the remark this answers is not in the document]
tlodderstedt: I wasn't aware of this convention. Thanks for making a suggestion.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: fixed it with the text from the parameter definition + added reference to the definition

openid-4-verifiable-presentations-1_0.md
tlodderstedt: Would you prefer to have one key in the request body? that would certainly also solve the apv/apu challenge. However, as far as I remember, there was also the desire to have authenticated keys. I would assume those would be provided with a cert chain related to the trust framework related to a certain Client Identifier.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: like? ...

tlodderstedt:
Discussion in the WG meeting on the 14th of Jan:
- session transcript: there are two proposals on the table that were agreed to think through
 - @hlozi suggested to drop the Client Identifier from the session transcript (and have the origin only)
 - @awoie suggested to represent the relationship between a certain Client Identifier and the sub query / credential that Client Identifier is related to in the request and use that information to determine the expected Client Identifier for the corresponding Device Response object. 
- response encryption: conclusion was to move the `client_metadata` parameter into the payload (again), i.e. there is one encryption key per request.

awoie:
> Discussion in the WG meeting on the 14th of Jan:
> 
> 1. session transcript: there are two proposals on the table that were agreed to think through
> 
> * @awoie suggested to represent the relationship between a certain Client Identifier and the sub query / credential that Client Identifier is related to in the request and use that information to determine the expected Client Identifier for the corresponding Device Response object.

I don't think this is possible. I'm leaning towards removing the `client_id` from `SessionTranscript` but it raises the question why it is needed for other verifiable presentations in case of DC API.

paulbastian:
I summarized my views in this flow chart. I came to the conclusion that client_id is not needed in session transcript. Ultimately that would mean it makes sense to have `aud` in KB-JWT to represent origin in DC-API case.
image

openid-4-verifiable-presentations-1_0.md
bc-pi: I would assume in that case to the best of how I understand things, those keys would be authenticated by the signature over them and trust in the key that applied that signature is established by that a cert chain related to the trust framework related to a certain Client Identifier.

I don't think that's how some of those that have expressed the desire to have authenticated keys think it should or will work. But it's how I imagine it'd be done with the fewest breaking or impractical changes down the stack of things OpenID4VP builds on.

openid-4-verifiable-presentations-1_0.md
bc-pi: Well, the first question that would come to my mind is, "what the fuck is this?" but t think about it from the perspective that this will maybe eventually be put forwarded through various bureaucratic entities to be considered as a registered JWS header. JOSE is a general purpose thing that is *many* layers removed from this document and the DE's might not know or care about OpenID4VP or any of this stuff. What is `client_id` and what does it do and why does it need or deserve to be general reserved JOSE header? `This header contains a Client Identifier.` and a pointer to RFC6749 are not, in my narrow view as the DE for a different registry, helpful at all.

tlodderstedt:
> I don't think this is possible. I'm leaning towards removing the `client_id` from `SessionTranscript` but it raises the question why it is needed for other verifiable presentations in case of DC API.

Can you please explain why this is not possible? We can add a "credentials" array to every client identifier specific "signature" object and use that to link the Client Identifier.

paulbastian:
> > I summarized my views in this flow chart. I came to the conclusion that client_id is not needed in session transcript. Ultimately that would mean it makes sense to have `aud` in KB-JWT to represent origin in DC-API case.
> 
> That would mean the audience restriction would differ depending on the "transport". I would assume that makes implementations (at least in transition from traditional protocol to DC) harder. It also changes the level on which the restriction needs to be checked (client id/application vs. origin/transport).

That's already the reality for ISO mdoc format. I don't think it makes it much more complicated, because the input for credential formats, so mdoc and SD-JWT implementations don't need to change. The protocol implementations change anyway when people add DC-API, the change is how to build session_transcript/aud and that's what you stuff into the credential format library.

tlodderstedt:
> That's already the reality for ISO mdoc format.

Doesn't the device response contain the client id?

paulbastian:
> > That's already the reality for ISO mdoc format.
> 
> Doesn't the device response contain the client id?

Not sure what you mean with device response.

client_id is present at two places in ISO18013-7
- as Authorization Request parameter
- in the session_transcript

The first is removed with DC-API and the second may be as well following my discussion line here in this thread.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2013,19 +2013,70 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
 ```
 
-This is an example of the payload of a signed OpenID4VP request used with the DC API:
+This is an example of the payload of a signed OpenID4VP request used with the W3C Digital Credentials API in conjunction with JWS Compact Serialization:
+
+<{{examples/digital_credentials_api/signed_request_payload_compact.json}} 
+
+#### JWS JSON Serialization
+
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
```
peppelinux, line 2036:
```suggestion
The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests Credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
```

peppelinux (approved):
I can approve this PR because it enables an important flexibility in how a wallet evaluates the trust with an RP.

I have read and I found valuable the comments of other contributors, I have decided to approve it in the current status anyway to facilitate the process.

Regarding the question: would we be able to support both compact and json serialization, I want to answer that it would be in the scope of any good library detect and support both the formats, in full support of RFC 7515. I would not use openid4vp to limit the use of the serialization types (impl profile could decide this, if/when required).

tlodderstedt:
> The first is removed with DC-API and the second may be as well following my discussion line here in this thread.
>

Not sure what you mean, the client identifier is either present in the request (signed request) or determined from the origin (unsigned request). 

> and imo that makes sense, the client_id is kind of an origin equivalent, because OAuth couldn't achieve this without OS support.

They are different (in case of signed request) and the client identifier is more of a logical identifier that could be used from different origins.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2677,6 +2730,14 @@ established by [@!RFC7515].
 * Change Controller: OpenID Foundation Artifact Binding Working Group - openid-specs-ab@lists.openid.net
 * Specification Document(s): (#verifier_attestation_jwt) of this specification
 
+### client_id
+
+* Header Parameter Name: `client_id`
+* Header Parameter Description: This header contains a Client Identifier.
```
tlodderstedt, line 2736:
```suggestion
* Header Parameter Description: This header contains a Client Identifier. A Client Identifier is used in OAuth to identify a certain client. It is defined in [@!RFC6749], section 2.2.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2013,19 +2013,72 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+
+```
+{
+  "alg": "ES256",
+  "x5c": [
+    "MIICOjCCAeG...djzH7lA==",
+    "MIICLTCCAdS...koAmhWVKe"
+  ],
+  "client_id": "x509_san_dns:rp.example.com"
+}
```
Sakurann, line 2076: specific proposal for disambiguating credential requested and credential_id

```suggestion
{
  "alg": "ES256",
  "x5c": [
    "MIICOjCCAeG...djzH7lA==",
    "MIICLTCCAdS...koAmhWVKe"
  ],
  "client_id": "x509_san_dns:rp.example.com",
  "credentials": [ <same definition as transaction data> ]
}
```

Sakurann: credential_ids: REQUIRED. Array of strings each referencing a Credential requested by the Verifier that can be used to authorize this transaction. In [[DIF.PresentationExchange](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#DIF.PresentationExchange)], the string matches the id field in the Input Descriptor. In the Digital Credentials Query Language, the string matches the id field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for session transcript

Sakurann: alternative proposal:
wallet puts used client_id in the response next to the credential:
{
  "my_credential": {
     “payload”: “eyJhbGci...QMA”,
     “client_id”: “asdf”
   }
}

Sakurann: what prevents MITM replaying the request RP sent through the DC API:
- for unsigned requests, origin is in the returned credential because origin == client_id 
- for signed request, expected_origins is in the signed request that MITM cannot modify

bc-pi: > what prevents MITM replaying the request RP sent through the DC API:
> 
>     * for unsigned requests, origin is in the returned credential because origin == client_id
> 
>     * for signed request, expected_origins is in the signed request that MITM cannot modify

Thank you for this summary and apologies for struggling with the concept during the most recent DCP call. One nit just for posterity is that for unsigned request the client_id contains the origin and is prefixed (it doesn't strictly equal origin).

Sakurann:
WG discussion notes
issue: which client id to put in the session transcript - there are two proposals on the table that were agreed to think through:
1. drop the Client Identifier from the session transcript (and have the origin only)
2. represent the relationship between a certain Client Identifier and the sub query / credential that Client Identifier is related to in the request and use that information to determine the expected Client Identifier for the corresponding Device Response object.

arguments to go the second way:
- this means no more audience restriction and that has been a fundamental security feature. much larger change since openid4vp specification currently mandates audience restrictions for all credential format. which means this has impact on other credential formats too.
- second approach is also disambiguation

proposal and further discussion documented https://github.com/openid/OpenID4VP/pull/308/files#r1927297499

Sakurann:
We don't have another DCP WG mtg before ISO virtual mtg next week. Having heard DCP WG discussion and having re-read ISO requirement, what I would like to offer as a way forward is **to explain at the ISO virtual mtg that we are addressing ISO requirement of “mdoc authentication must be bound to the origin” not by including origin in the session transcript but by what is described below**. If this gets accepted, we can remove origin from session transcript as an option, and conclude this circular discussion we have been having:

> what prevents MITM replaying the request RP sent through the DC API:
>
> - for unsigned requests, origin is in the returned credential because origin == client_id
> - for signed request, expected_origins is in the signed request that MITM cannot modify

David-Chadwick:
Would it be better to rephrase the first bullet point to

for unsigned requests, the resulting credential is returned to the origin, because origin == client_id

openid-4-verifiable-presentations-1_0.md
c2bo: How about adding the `client_id` as an optional parameter (array of strings) to the Credential Query? That would simplify things quite a bit imho for the request:
- If client_id is not present in the Credential Query: you use only one client_id in the request and nothing changes to the status quo (without multi-rp)
- If client_id is present, the wallet can directly match which client ids are possible for each requested credential and choose accordingly

The DCQL `vp_token` gets an optional parameter `client_id` as well that is only included if more than one client_id was provided in the request (the DCQL query) and not present otherwise --> nothing changes for normal flows.

That way nothing changes for simple flows, but mult-rp flows extend the normal data structures without any unnecessary duplication unless I am missing something.

GarethCOliver:
Use Case: As a Verifier I want to request mDLs to verify KYC, and I accept mDLs issued in the EU, US.

US uses web PKI
EU uses an explicit trust framework 

This seems like it would be a problem, unless we allow a mechanism to know to use the origin for particular returned credentials. (A US wallet may not be able to validate an EU signature, and so can't verify the signature over expected_origins)

openid-4-verifiable-presentations-1_0.md
Sakurann: superseded by resolution in #395

openid-4-verifiable-presentations-1_0.md
Sakurann: @bc-pi

Sakurann (approved):

c2bo (approved):

examples/digital_credentials_api/signed_request_payload_compact.json
```diff
@@ -0,0 +1,25 @@
+{
+  "expected_origins": [
+    "https://origin1.example.com",
+    "https://origin2.example.com"
+  ],
+  "client_id": "x509_san_dns:rp.example.com",
```
Sakurann, line 6:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2271,19 +2271,72 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+In this case, the following request parameters MUST be present in the protected header of the respective `signature` object in the `signatures` array defined in [@!RFC7515, section 7.2.1]:
+
+* `client_id`
 
-The signed request allows the Wallet to authenticate the Verifier using a trust framework other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the Origin of the Verifier. This Origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered Origins.
+All other request parameters MUST be present in the `payload` element of the JWS object.
+
+Below is a non-normative example of such a request:
+
+```js
```
danielfett, line 2306:
```suggestion
```json
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2271,19 +2271,72 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+    }
+  ]
+}
+```
+
+Every object in the `signatures` structure contains the parameters and the signature specific to a particular Client Identifier. The signature is calculated as specified in section 5.1 of [@!RFC7515].
+
+The following is a non-normative example of a content of a decoded protected header:
+
+```
```
danielfett, line 2326:
```suggestion
```json
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2271,19 +2271,72 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
 
 The Verifier MAY send a signed request, for example, when identification and authentication of the Verifier is required.
 
+The signed request allows the Wallet to authenticate the Verifier using one or more trust framework(s) in addition to the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the web origin of the Verifier. This web origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered web origins.
+
 The signed Request Object MAY contain all the parameters listed in (#dc_api_request), except `request`.
 
-Below is a non-normative example of such a request sent over the Digital Credentials API (DC API):
+Verifiers can format signed Requests either using JWS Compact Serialization or JWS JSON Serialization [@!RFC7515]).
```
danielfett, line 2278:
```suggestion
Verifiers SHOULD format signed Requests using JWS Compact Serialization but MAY use JWS JSON Serialization [@!RFC7515]) to cater for use cases described below. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2271,19 +2271,72 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
...
+<{{examples/digital_credentials_api/signed_request_payload_compact.json}} 
 
-<{{examples/digital_credentials_api/signed_request_payload.json}}
+#### JWS JSON Serialization
+
+The JWS JSON Serialization [@!RFC7515]) allows the Verifier to use multiple Client Identifiers and corresponding key material and metadata to protect the same request. This serves use cases where the Verifier requests Credentials belonging to different trust frameworks and, therefore, needs to authenticate in the context of those trust frameworks.
+
+In this case, the following request parameters MUST be present in the protected header of the respective `signature` object in the `signatures` array defined in [@!RFC7515, section 7.2.1]:
+
+* `client_id`
```
danielfett, line 2300: What about client_metadata?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2996,6 +3049,14 @@ established by [@!RFC7515].
 * Change Controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
 * Specification Document(s): (#verifier_attestation_jwt) of this specification
 
+### client_id
+
+* Header Parameter Name: `client_id`
```
danielfett, line 3054: As above, what about client_metadata?

openid-4-verifiable-presentations-1_0.md
tlodderstedt: moved it into payload (again) because we wanted to convey the (ephm) response encryption key only once, removed metadata further up

Issue #310, opened 2024-11-06
Response Encryption : Why HPKE without JSON envelope


hlozi:
The encryption for The Digital Credential Response should be specified in detail to ensure interoperability. Since the API is one-shot and doesn’t allow for negotiation, there should be a single encryption method. Negotiation would lead to privacy leaks.

Why HPKE ?

- HPKE is format-agnostic, unlike JARM, which is JWT-specific. This means that when used in transmission modes where JWT isn’t appropriate, a different encryption mechanism is needed.
- JARM requires the transmission of all authenticated information, including APU, and AVP headers. However, there are security benefits in not sending all information and providing it on both sides.
- The JSON envelope adds complexity that provides little value when HPKE is used.
- HPKE as a mechanism has support on native platform libraries.
- In constrained environments like UICC, SE, etc., dealing with JSON structures can be very challenging.

jogu:
I think there's 3 separate points that are probably being mixed together here:

1. For the interoperability profiles we need to select a single encryption method (which is a pretty uncontentious statement and something OID4VP does expect people to do, e.g. ISO 18013-7 picked ECDH-ES I think)
2. The suggestion that HPKE should be that single encryption method (which is possible already without any breaking changes to OID4VP / will be possible when https://datatracker.ietf.org/doc/draft-ietf-jose-hpke-encrypt/ is further along in the IETF process. The suggestion to (optionally?) omit API/AVP headers is I presume something that could be suggested to the authors of this draft.)
3. The suggestion to not use JARM, which would require a breaking change to OID4VP itself, rather than simply profiling it.

paulbastian:
As discussed on DCP Call, currently the JARM encrypts the `vp_token`, which is a JSON structure. A possible (breaking) optimization could be to only encrypt individual Credentials, that could enable passing the encrypted credential to a SecureElement without it doing any JSON parsing.

c2bo:
As discussed on the WG call today, I'd propose to initially add a small note in the JARM section that clarifies the different options from JOSE and explicitly points to HPKE as one such option.

Switching from the current variant to encryption on the level of individual credentials is something that needs a bit more discussion and also has some dependencies on features currently getting discussed like wallet attestation as part of the authorization response (and where exactly it resides).

paulbastian:
To illustrate, currently we do:
Auth Response -> JARM JWE -> vp_token -> credential
                                                                         -> credential

The suggestion is to:
[...]

PR #315, opened 2024-11-11
note on encryption choices (JARM)

c2bo:
Editorial note as briefly discussed in last weeks DCP calls and mentioned in #310 to mention HPKE as one of the choices for encryption.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1155,6 +1155,8 @@ To obtain Verifier's public key for the input to the key agreement to encrypt th
 
 To sign the Authorization Response, the Wallet MUST use a private key that corresponds to a public key made available in its metadata.
 
+Note: For encryption, implementers can utilize the different options provided by JOSE, including variants like Hybrid Public Key Encryption (HPKE) as introduced in [@I-D.ietf-jose-hpke-encrypt].
+
 ### Response Mode "direct_post.jwt" {#direct_post_jwt}
 
 This specification also defines a new Response Mode `direct_post.jwt`, which allows for JARM to be used with Response Mode `direct_post` defined in (#response_mode_post).


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1155,6 +1155,8 @@ To obtain Verifier's public key for the input to the key agreement to encrypt th
 
 To sign the Authorization Response, the Wallet MUST use a private key that corresponds to a public key made available in its metadata.
 
+Note: For encryption, implementers can utilize the different options provided by JOSE, including variants like Hybrid Public Key Encryption (HPKE) as introduced in [@I-D.ietf-jose-hpke-encrypt].
```
peppelinux, line 1158:
```suggestion
Note: For encryption, implementers have a variety of options available through JOSE, including the use of Hybrid Public Key Encryption (HPKE) as detailed in [@I-D.ietf-jose-hpke-encrypt]. 
```

peppelinux (approved):

Sakurann (approved):

Sakurann:
this can probably go into the implementers draft since it is technically a non-normative change

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1155,6 +1155,8 @@ To obtain Verifier's public key for the input to the key agreement to encrypt th
 
 To sign the Authorization Response, the Wallet MUST use a private key that corresponds to a public key made available in its metadata.
 
+Note: For encryption, implementers have a variety of options available through JOSE, including the use of Hybrid Public Key Encryption (HPKE) as detailed in [@I-D.ietf-jose-hpke-encrypt].
```
Sakurann, line 1158: don't we need to say which of the two modes we are recommending to use?

> This specification describes two modes of use for HPKE in JWE:
>
> - HPKE JWE Integrated Encryption, where HPKE is used to encrypt the plaintext.
>
> - HPKE JWE Key Encryption, where HPKE is used to encrypt a content encryption key (CEK) and the CEK is subsequently used to encrypt the plaintext.

c2bo: Not sure if we should or need to make that recommendation here - the choice which mode to use should be imho done by an ecosystem / a profile. From an implementers perspective, it is discoverable which mode was used via the JWE parameters (alg, enc, iv, tag), so this would mainly be about reducing options within openid4vp?

jogu (approved):

PR #318, opened 2024-11-12
Wallet Attestation: Option with new request/response parameters

### LINKED ISSUE #141: Verifier/Relying Parties need to be able to verify authenticity and validity of (European Digital Identity) Wallets
The eIDAS regulation has the following text: 
Article 5a, 5., (a), viii
'European Digital Identity Wallets shall, in particular (…) support common protocols and interfaces (…) for relying parties to verify the authenticity and validity of European Digital Identity Wallets;'

We need to come up with means to allow wallets to comply with this requirement. 

I assume EUDI Wallets will have a credential, referred to with different terms such as "Wallet Instance Attestation" or "Wallet Attestation" or "Wallet Trust Anchor", which it could present to the Verifier along with a Proof of Possession (PoP) of the respective key in the credential. I will use the term Wallet Attestation in this text. The Wallet Attestation is issued by a trusted party, which can assert the authenticity and validity of the wallet. 
I also assume the Verifier explicitly requests the authenticity/validity check. 

I see the following options to build support into the protocol: 

1. The Wallet sends a Wallet Attestation with a PoP in the POST request to the request URI of the Verifier. That's the earliest point where the authentication can happen. It would require an extension to the POST request. The Verifier could indicates in the authorization request that it requires Wallet authentication along with a fresh nonce the PoP must incorporate. It would also be possible the Verifier responds with a 401 in the POST request (providng a nonce) and the Wallet needs to send the request again, this time with the Wallet Attestation and PoP. The syntax could leverage syntax as defined in the "attestation based client authentication" draft. 
2. The Verifier requests the Wallet Attestation like any other Attestation in the Request. The Wallet Attestation is provided in the `vp_token`. This option would not require any change to the spec. 
3. The Verifier requests the Wallet Attestation with a special parameter in the request, the Wallet Attestation's PoP key is used to sign the Response. The Wallet Attestation is provided in a JOSE header of the JARM response structure.

c2bo:
Closes #141.
This is the option of new dedicated request & response parameters.


openid-4-verifiable-presentations-1_0.md
```diff
@@ -118,6 +118,9 @@ VP Token:
 Wallet:
 :  An entity used by the Holder to receive, store, present, and manage Verifiable Credentials and key material. There is no single deployment model of a Wallet: Verifiable Credentials and keys can both be stored/managed locally, or by using a remote self-hosted service, or a remote third-party service. In the context of this specification, the Wallet acts as an OAuth 2.0 Authorization Server (see [@!RFC6749]) towards the Credential Verifier which acts as the OAuth 2.0 Client.
 
+Wallet Attestation:
+: An attestation designed to prove authenticity of the wallet towards parties it is interacting with, in the context of this specification, the Verifier. The overall concept, format and data structure of a Wallet Attestation is defined in [@!OpenID.VCI].
```
peppelinux, line 122:
```suggestion
: A signed proof crafted to demonstrate the authenticity of the Wallet to the parties it interacts with, specifically the Verifier in this specification. The comprehensive concept, format, and data structure of a Wallet Attestation are outlined in [@!OpenID.VCI].
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -310,6 +313,9 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`wallet_attestation`:
+: OPTIONAL. A boolean value that indicates whether or not a wallet attestation is requested by the Verifier. If the parameter is not present, the default value is `false`. If the value is `true`, a wallet attestation is expected to be part of the response (see `wallet_attestation` in (#response-parameters)).
```
peppelinux, line 317:
```suggestion
: OPTIONAL. A boolean value that indicates whether or not a Wallet Attestation is requested by the Verifier. If the parameter is not present, the default value is `false`. If the value is `true`, a Wallet Attestation is expected to be part of the response (see `wallet_attestation` in (#response-parameters)).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1631,12 +1642,16 @@ If no End-User interaction is required before sending the request, it is easy to
...
+Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount of information possible, and in particular, without any HTTP headers identifying the software used for the request (e.g., HTTP libraries or their versions). The Wallet MUST NOT send PII or any other data that could be used for fingerprinting to the Request URI in order to prevent End-User tracking.
 
 ## Authorization Error Response with the `wallet_unavailable` error code
 
 In the event that another component is invoked instead of the Wallet, the End-User MUST be informed and give consent before the invoked component returns the `wallet_unavailable` Authorization Error Response to the Verifier.
 
+## Wallet Attestation
+
+Wallet Attestations could be an additional correlation factor that could allow End-User tracking. Wallet Attestations SHOULD use privacy-preserving mechanisms similar to the ones that are applied to credentials in an ecosystem, e.g., use of one-time usage Wallet Attestations if the same applies for credentials.
```
peppelinux, line 1653:
```suggestion
Wallet Attestations could be an additional correlation factor that could allow End-User tracking. Wallet Attestations SHOULD use privacy-preserving mechanisms similar to the ones that are applied to Credentials in an ecosystem, e.g., use of one-time usage Wallet Attestations if the same applies for Credentials.
```

peppelinux:
interesting approach, my points below:

1. How can an RP request a specific wallet attestation over another, given that wallet attestations are tied to the trust frameworks in use and, like digital credentials, may have varying schemes and identifiers?

2. This PR would benefit from an example where the presentation_submission includes a Wallet Attestation, specifying to the Verifier the index position within the vp_token array where the Wallet Attestation is provided.

3. Considering point 1, why isn't the Wallet Attestation included in the presentation_definition/duckle query like other digital credentials? Why would using a request parameter be considered superior to a standard presentation query?

I support this work

c2bo:
> interesting approach, my points below:
> 
> 1. How can an RP request a specific wallet attestation over another, given that wallet attestations are tied to the trust frameworks in use and, like digital credentials, may have varying schemes and identifiers?
> 2. This PR would benefit from an example where the presentation_submission includes a Wallet Attestation, specifying to the Verifier the index position within the vp_token array where the Wallet Attestation is provided.
> 3. Considering point 1, why isn't the Wallet Attestation included in the presentation_definition/duckle query like other digital credentials? Why would using a request parameter be considered superior to a standard presentation query?
> 
> I support this work

There are 2 approaches right now 
- special parameters for request/response (this also means attestation is not in vp_token, but directly in the dedicated response paramter) -> this is the option I've tried to describe in this PR and was the one that most people agreed with the last time this was discussed
- integregate wallet attestation into dcql query and vp_token. I am planning to do a second drat PR with this option (which I personally prefer). I will likely choose the variant with a dedicated request parameter that links to a credential_id in the dcql request as described here: https://github.com/openid/OpenID4VP/issues/141#issuecomment-2461946150

openid-4-verifiable-presentations-1_0.md
```diff
@@ -978,6 +985,9 @@ If Presentation Exchange was used for the request, the `presentation_submission`
 
 Including the `presentation_submission` parameter as a separate response parameter allows the Wallet to provide the Verifier with additional information about the format and structure in advance of the processing of the VP Token, and can be used even with the Credential formats that do not allow for the direct inclusion of `presentation_submission` parameters inside a Credential itself.
 
+`wallet_attestation`:
+: REQUIRED if the `wallet_attestation` parameter was present and its value `true` in the Authorization Request (see (#vp_token_request) for more details). The Wallet Attestation presentation MUST be bound to the nonce that was provided in the Authorization Request. The expected format of a Wallet Attestation and validation rules are defined in section TODO of [@!OpenID.VCI]. If the validation of the Wallet Attestation fails, the response is rejected. Additional claims for individual implementations and ecosystems can be added to a Wallet Attestation.
```
Sakurann, line 989:
```suggestion
`wallet_attestation`:
: REQUIRED if the `wallet_attestation` parameter was present and its value `true` in the Authorization Request (see (#vp_token_request) for more details). The Wallet Attestation presentation MUST be bound to the nonce that was provided in the Authorization Request. The expected format of a Wallet Attestation and validation rules are defined in [@!OpenID.VCI]. If the validation of the Wallet Attestation fails, the response is rejected. Additional claims for individual implementations and ecosystems can be added to a Wallet Attestation.
```

Sakurann (approved):

martijnharing:
I don't see the benefit of the approach from this PR over using using the mechanisms that are already present (e.g. DCQL) for requesting wallet attestation. As far as I'm aware using the request and response logic using existing credentials can support all the features required for the wallet attestation use case without the need of any special handling.

Using the approach in this PR doesn't support certain features that the DCQL does support like making queries that say "for this credential I need a wallet attestation, but for this one I don't" or support of wallet attestations in different formats than the one considered in the OpenID4VCI spec. We could of course add these features, but then we are just making another version of DCQL.

It was mentioned that one of the reasons to make a special parameter was that it allows wallets to treat the request for wallet attestation different than other credentials. But since wallet UI / UX is out of scope of the specification there is nothing in the current specification that stops a wallet from implementing that logic based on the credential type indicator that is already present in DCQL (e.g. vct_values).

openid-4-verifiable-presentations-1_0.md
c2bo: My plan was to wait for the merge of the VCI PR and adjust this PR with the correct section

c2bo:
> I don't see the benefit of the approach from this PR over using using the mechanisms that are already present (e.g. DCQL) for requesting wallet attestation. As far as I'm aware using the request and response logic using existing credentials can support all the features required for the wallet attestation use case without the need of any special handling.
> 
> Using the approach in this PR doesn't support certain features that the DCQL does support like making queries that say "for this credential I need a wallet attestation, but for this one I don't" or support of wallet attestations in different formats than the one considered in the OpenID4VCI spec. We could of course add these features, but then we are just making another version of DCQL.
> 
> It was mentioned that one of the reasons to make a special parameter was that it allows wallets to treat the request for wallet attestation different than other credentials. But since wallet UI / UX is out of scope of the specification there is nothing in the current specification that stops a wallet from implementing that logic based on the credential type indicator that is already present in DCQL (e.g. vct_values).

I wouldn't agree that UI/UX is fully out of scope - there are enough parts in the OpenID4VC protocols where certain aspects were introduced to make a better UX possible. Examples being things like the display metadata in VCI & new proof type to avoid user interaction for re-iussuance in VCI.

On the topic: I do believe there should be some kind of part of the request (whether a request parameter or part of DCQL) that marks an attestation request explicitly as such. At the end of the day, a Wallet needs to be able to deal with an attestation accordingly, but we should rather help implementations to fail fast if they do now know an attestation by making it explicit imho.

martijnharing:
> > I don't see the benefit of the approach from this PR over using using the mechanisms that are already present (e.g. DCQL) for requesting wallet attestation. As far as I'm aware using the request and response logic using existing credentials can support all the features required for the wallet attestation use case without the need of any special handling.
> > Using the approach in this PR doesn't support certain features that the DCQL does support like making queries that say "for this credential I need a wallet attestation, but for this one I don't" or support of wallet attestations in different formats than the one considered in the OpenID4VCI spec. We could of course add these features, but then we are just making another version of DCQL.
> > It was mentioned that one of the reasons to make a special parameter was that it allows wallets to treat the request for wallet attestation different than other credentials. But since wallet UI / UX is out of scope of the specification there is nothing in the current specification that stops a wallet from implementing that logic based on the credential type indicator that is already present in DCQL (e.g. vct_values).
> 
> I wouldn't agree that UI/UX is fully out of scope - there are enough parts in the OpenID4VC protocols where certain aspects were introduced to make a better UX possible. Examples being things like the display metadata in VCI & new proof type to avoid user interaction for re-iussuance in VCI.
> 
> On the topic: I do believe there should be some kind of part of the request (whether a request parameter or part of DCQL) that marks an attestation request explicitly as such. At the end of the day, a Wallet needs to be able to deal with an attestation accordingly, but we should rather help implementations to fail fast if they do now know an attestation by making it explicit imho.

You are correct, we definitely design features to support good UI/UX. I meant that there isn't an explicit requirement to actually implement that in a certain way. And In this particular case the credential type (vct_value, doctype) achieves the goal of identifying that something is a credential attestation.

David-Chadwick:
The other defining feature is that the subject of the credential is the wallet app, and not the user. Given that wallets should be designed to hold credentials of multiple subjects e.g. children, pets, wallet apps etc. then any good wallet GUI will have tabs for the different subjects so that the holder can see which credentials s/he holds for whom.
For this reason I can see no good reason to differentiate the wallet app credential from any other credential.

tlodderstedt (approved):
lgtm

tlodderstedt:
> then any good wallet GUI will have tabs for the different subjects so that the holder can see which credentials s/he holds for whom.

You are making a good point. I would not assume that the user interface shows wallet attestations to the user, it's a technical mechanism like RP credentials to ensure the security of the ecosystem. To me this is a good reason to treat the wallet attestation differently.

David-Chadwick:
`I would not assume that the user interface shows wallet attestations to the user, it's a technical mechanism like RP credentials to ensure the security of the ecosystem. To me this is a good reason to treat the wallet attestation differently.`

On the contrary, I think it is a very good idea to show this to the user since it will help to provide the user with confidence and trust in the wallet app, and will differentiate this wallet app from a cheap substitute (that isn't certified). After all, consumers like to see trust marks of various kinds on physical items that they purchase, to give them confidence that they are safe to use, so why not for virtual items that they obtain?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -310,6 +313,9 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`wallet_attestation`:
+: OPTIONAL. A boolean value that indicates whether or not a Wallet Attestation is requested by the Verifier. If the parameter is not present, the default value is `false`. If the value is `true`, a Wallet Attestation is expected to be part of the response (see `wallet_attestation` in (#response-parameters)).
```
danielfett, line 317: I'm fine with this approach if we can ensure that we leave the door open to later (in future versions) add a DCQL query or other complex query here if needed.

Not sure if this is the best way:

```suggestion
: OPTIONAL. A value that indicates whether or not a Wallet Attestation is requested by the Verifier. If the parameter is not present, or is `false`, no Wallet Attestation is requested. Otherwise, a Wallet Attestation is expected to be part of the response (see `wallet_attestation` in (#response-parameters)).
```

openid-4-verifiable-presentations-1_0.md
danielfett, line 989:
```suggestion
: REQUIRED if the `wallet_attestation` parameter was requested in the Authorization Request (see (#vp_token_request) for more details). The Wallet Attestation presentation MUST be bound to the nonce that was provided in the Authorization Request. The expected format of a Wallet Attestation and validation rules are defined in section TODO of [@!OpenID.VCI]. If the validation of the Wallet Attestation fails, the response MUST be rejected. Additional claims for individual implementations and ecosystems can be added to a Wallet Attestation.
```

openid-4-verifiable-presentations-1_0.md
danielfett: Maybe add something like "Note: Instead of offering just the boolean values `true`/`false`, extensions of this specifications may define custom queries to be used in this parameter. Therefore, any value other than `false` MUST be treated as a request for a Wallet Attestation."

danielfett (approved):
Approved under the condition that we can leave the door open for more complex queries in the future.

c2bo:
> Approved under the condition that we can leave the door open for more complex queries in the future.

Do I understand your comment correctly that you would expect more complex queries to exist in the `wallet_attestation` parameter instead of directly being part of the DCQL query?

David-Chadwick:
I am also opposed to this PR. I maintain that treating the wallet attestation like any other credential is the right approach, and then it can selected by the RP in the same way that it selects any other credential to be returned to it.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -959,6 +966,9 @@ If Presentation Exchange was used for the request, the `presentation_submission`
 
 Including the `presentation_submission` parameter as a separate response parameter allows the Wallet to provide the Verifier with additional information about the format and structure in advance of the processing of the VP Token, and can be used even with the Credential formats that do not allow for the direct inclusion of `presentation_submission` parameters inside a Credential itself.
 
+`wallet_attestation`:
+: REQUIRED if the `wallet_attestation` parameter was requested in the Authorization Request (see (#vp_token_request) for more details). The Wallet Attestation presentation MUST be bound to the nonce that was provided in the Authorization Request. The expected format of a Wallet Attestation and validation rules are defined in section TODO of [@!OpenID.VCI]. If the validation of the Wallet Attestation fails, the response MUST be rejected. Additional claims for individual implementations and ecosystems can be added to a Wallet Attestation.
```
c2bo, line 970:
```suggestion
: REQUIRED if the `wallet_attestation` parameter was requested in the Authorization Request (see (#vp_token_request) for more details). The Wallet Attestation presentation MUST be bound to the nonce that was provided in the Authorization Request. The expected format of a Wallet Attestation and validation rules are defined in Appendix E of [@!OpenID.VCI]. If the validation of the Wallet Attestation fails, the response MUST be rejected. Additional claims for individual implementations and ecosystems can be added to a Wallet Attestation.
```

danielfett:
> > Approved under the condition that we can leave the door open for more complex queries in the future.
> 
> Do I understand your comment correctly that you would expect more complex queries to exist in the `wallet_attestation` parameter instead of directly being part of the DCQL query?

yes

openid-4-verifiable-presentations-1_0.md
```diff
@@ -620,6 +626,7 @@ The following is a non-normative example of a request object:
   "presentation_definition": {...},
   "nonce": "n-0S6_WzA2Mj",
   "wallet_nonce": "qPmxiNFCR3QTm19POc8u",
+  "wallet_attestation": true,
```
peppelinux, line 629: ok, I want to find a solution but, sincerely, this appears very naif.

different trust frameworks may defines many different wallet attestations. By only requesting a "wallet attestation" (any of them, at least one) would end in a useless/weak parameter.

The RP needs to ask one or more (considering at least one of a set of supported one) wallet attestation.

For this reason, using presentation_definition or duckle appeared to me the best way to fully specify the wallet attestation, by type and properties.

peppelinux (changes requested):

peppelinux:
I can say that the words spent by @David-Chadwick and @martijnharing are also mine.

andprian:
@Sakurann my two cents on the questions you addressed:

1. in the EUDIW context we previously had 2 attestations (WTE and WIA) that were merged into one (WUA), but I am not sure this is going to stay like this. In other standards we keep these 2 concepts separate (WTE and WIA) and leave room to implement this as one or two attestations until this topic is stabilized. I think here we should offer the same flexibility. But if there will be only one attestation, it will serve different purposes and we should support selective disclosure to differentiate when it is presented to a Issuer or to a RP. The format should be the same as for PID or the usual EAA, so do not add the burden to support another format.
2. i do not see the point in asking the user if he wants to share this because:
- the average user (so the majority of the population) will not understand what this is
- technical stuff is usually shared in a transparent way - you do not ask the user if he wants to share his public key for instance
- the RP service cannot function without this attestation

tlodderstedt (approved):

Sakurann:
European Commission seems to be dropping the requirement to send wallet attestation to the RPs, to remove the burden on the RPs and improve the adoption. It’s Issuer and wallet provider’s responsibility to make sure that credentials in the wallet are valid during the presentation to the verifier: we are not sure which of the alternative solutions are better and more scalable: chain revocation or short lived credentials (less than 24h), but there is an agreement that sending wallet attestation to the RPs is the least preferred one

PR #330, opened 2024-11-19
define where in mdocs to include transaction_data_hashes

### LINKED ISSUE #259: define where in mdoc presentation to include transaction data
for sd-jwt vc, it is in key binding jwt. for mdoc, is it session_transcript?

Sakurann:
blocked by #421 

resolves #259

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2234,6 +2234,14 @@ See ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 Annex C [@ISO.
 
 The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as defined in ISO/IEC 18013-5 [@ISO.18013-5] or ISO/IEC 23220-4 [@ISO.23220-4]. Essentially, the `DeviceResponse` CBOR structure contains a signature or MAC over the `SessionTranscript` CBOR structure including the OpenID4VP-specific `Handover` CBOR structure.
 
+The `transaction_data_hashes` response parameter MUST be returned in DeviceSigned using the `transaction_data_hashes` DataElement defined below:
+
+| Namespace | Identifier | Description | Encoding format |
+| --------- | ---------- | ----------- | --------------- |
+| `net.openid.openid4vc` | `transaction_data_hashes` | as defined in Section 7.4 of this specification | [ tstr* ] |
+
+If KeyAuthorization for the requested transaction authorization is not present, the Wallet MUST return an error.
+
 See ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 Annex C [@ISO.23220-4] for the latest examples on how to use the `presentation_submission` parameter and how to generate the Authorizaton Response for presenting Credentials in the mdoc format.
 
 ## IETF SD-JWT VC


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2234,6 +2234,14 @@ See ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 Annex C [@ISO.
 
 The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as defined in ISO/IEC 18013-5 [@ISO.18013-5] or ISO/IEC 23220-4 [@ISO.23220-4]. Essentially, the `DeviceResponse` CBOR structure contains a signature or MAC over the `SessionTranscript` CBOR structure including the OpenID4VP-specific `Handover` CBOR structure.
 
+The `transaction_data_hashes` response parameter MUST be returned in DeviceSigned using the `transaction_data_hashes` DataElement defined below:
+
+| Namespace | Identifier | Description | Encoding format |
+| --------- | ---------- | ----------- | --------------- |
+| `net.openid.openid4vc` | `transaction_data_hashes` | as defined in Section 7.4 of this specification | [ tstr* ] |
+
+If KeyAuthorization for the requested transaction authorization is not present, the Wallet MUST return an error.
```
Sakurann, line 2243: need to define proper data element identifier for KeyAuthorization.

martijnharing: When using the mdoc to sign data, there will often be security considerations is to whether the particular mdoc can / should / may sign a particular data element. Whether there are any limitations strongly dependent on the specific document type and use case. But typically the Issuer indicates with a particular namespace and data element identifier which exact things the mdoc can / should / may sign.
So from a security perspective we should define the namespace and data element dependent on the use case and not as a generic identifier.

GarethCOliver: as I understand it, transaction_data contains a 'type' parameter. If we want issuers to have more granular control over what type of transaction_data is allowed to be returned by the wallet we could define a set of DataElements:

namespace: `net.openid.openid4vc.transactiondata`
identifier: must be the `type` of the transaction_data

The wallet MUST return split the transaction_data_hashes so each is returned as part of the value array in a DataElement with a namespace net.openid.openid4vc.transactiondata and identifier of transaction_data.type. A wallet MUST return an error if a request contains transaction_data with a type that is not present in the KeyAuthorizations. 

psuedocode example: 
paymentsTransactionData = { type = "payments", ... }
fooTransactionData1 = { type = "foo" , ...}
fooTransactionData2  = { type = "foo", ... }

request {
  ...
  transaction_data [ paymentsTransactionData, fooTransactionData1, fooTransactionData2 ]
 ..
}

then the response would have in DeviceSigned: 


...
`net.openid.openid4vc.transactiondata`:
  [ 
    { `payments` -> [ base64(hash(paymentsTransactionData))) }
   { `foo` ->  [ base64(hash(fooTransactionData1)), base64(hash(fooTransactionData2)) ] }
 ]
}
..

additionally the KeyAuthorizations would either need to authorize the entire `net.openid.openid4vc.transactiondata` namespace, or the identifiers `payments` and `foo` within the `net.openid.openid4vc.transactiondata` namespace.

@martijnharing this would provide the granularity you are looking for? 
@Sakurann does this roughly make sense?

andprian:
I believe the hash algorithm should also be protected by the signature of the mdoc. In the case of transaction data, signing implies commitment from the user on some data. Or the commitment is not complete if it is made on the data representation (hash) without committing to the algorithm as well.

GarethCOliver:
> I believe the hash algorithm should also be protected by the signature of the mdoc. In the case of transaction data, signing implies commitment from the user on some data. Or the commitment is not complete if it is made on the data representation (hash) without committing to the algorithm as well.

Yeah, you are right. Based on the outcome of https://github.com/openid/OpenID4VP/issues/443 it should either be a single, additional data element, or in a wrapper cbor structure (if alg remains per-transaction-data).
{
  hash = hash(...)
  alg = ALG
}

sander:
See also: https://github.com/openid/OpenID4VP/issues/259#issuecomment-2714622940

openid-4-verifiable-presentations-1_0.md
awoie: If the type was included, it would mean the issuer controls what types of transactions the wallet can sign with the credential. Is this really an issuer decision?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2234,6 +2234,14 @@ See ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 Annex C [@ISO.
 
 The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as defined in ISO/IEC 18013-5 [@ISO.18013-5] or ISO/IEC 23220-4 [@ISO.23220-4]. Essentially, the `DeviceResponse` CBOR structure contains a signature or MAC over the `SessionTranscript` CBOR structure including the OpenID4VP-specific `Handover` CBOR structure.
 
+The `transaction_data_hashes` response parameter MUST be returned in DeviceSigned using the `transaction_data_hashes` DataElement defined below:
```
awoie, line 2237:
```suggestion
The `transaction_data_hashes` response parameter MUST be returned in `DeviceSigned` using the `transaction_data_hashes` data element defined below:
```

openid-4-verifiable-presentations-1_0.md
awoie: Also, since the mdoc can use device signature or device mac, i would argue it should be mostly device signature since for transaction signing non-repudiation is probably a requirement, isn't it?

sander: > If the type was included, it would mean the issuer controls what types of transactions the wallet can sign with the credential. Is this really an issuer decision?

@awoie in my opinion, yes. See also https://github.com/openid/OpenID4VP/issues/259#issuecomment-2714622940. Of course, an issuer may decide to include a device-signed attribute with arbitrary CBOR content, confirming “any” transaction. But I do not expect for example EUDIW PID issuers to do so.

sander:
Note https://github.com/openid/OpenID4VP/pull/421 contains an alternative proposal.

openid-4-verifiable-presentations-1_0.md
martijnharing: The approach of @GarethCOliver on where to put the data mostly makes sense to me. I think we should define the openid namespace as an option that profiles can use, but not mandate it as the only option.

tlodderstedt:
Are the changes suggested in this PR alternative/additional with this proposed change? https://github.com/openid/OpenID4VP/pull/421/files#diff-3118c9756a1d7b361bb53af8bd9d65666476a27cfeeced56c47a5dccc38eac55R2257

sander:
> Are the changes suggested in this PR alternative/additional with this proposed change? https://github.com/openid/OpenID4VP/pull/421/files#diff-3118c9756a1d7b361bb53af8bd9d65666476a27cfeeced56c47a5dccc38eac55R2257

@tlodderstedt @Sakurann As I see it, it’s a conflicting alternative right now:

- #421 proposes that the mdoc type rulebook specifies which parameters from which transaction data types map to which device-signed items.
- #330 proposes a single device-signed item that any mdoc type rulebook can include to support any transaction data type.

I don’t expect that mdoc issuers will adopt the latter for open systems. If however OIDF still wants to provide the option, I suggest to rephrase it it as an addition instead:

- propose an optional `net.openid.openid4vc` namespace for mdoc items that enable the proposed approach to “stash the transaction data JSON into a single item”.

Then some mdoc type rulebooks can choose to support this as a catch-all mechanism, while other mdoc type rulebooks can apply the more idiomatic and granular mapping mechanism, without breaking any requirement as currently proposed in this PR.

Sakurann:
closing since #421 was merged

Issue #305, opened 2024-11-05
Presentation of non-mdoc credentials (e.g. sd-jwt vc) offline


tlodderstedt:
We had a good discussion at IIW #39 about ways to enable offline presentation of non-mdoc credentials.
Options discussed:
- Extension of ISO 18013-5
- Pursue OID4VP over BLE
- Use CTAP/HYBRID as transport for OID4VP requests and responses. 
The consensus in the session was the CTAP/HYBRID seems to be the most attractive approach as it leveraged CTAP/HYBRID (implementations and community and roadmap) and also leverages OID4VP feature for offline (e.g. transaction data).

c2bo:
My current understanding would be this:

We have a similar problem as we have with the Digital Credentials Browser API, where an extension to CTAP needs to be defined that would allow for an OpenID4VP Request to be conveyed using proximity protocols. Currently exists support for the DC API in Hybrid mode with the following supported flows that are relevant for credentials:

- dcp | credential presentation (Digital Credentials API)
- dci | credential issuance (Digital Credentials API)

Hybrid mode = flow where proof of proximity (e.g., BLE beacon) and data transport (e.g., tunneled over the internet) are decoupled.

For the feature discussed in this issue, we would need support for something similar, where data is transported using the proximity protocols (NFC, BLE, UWB, ..) as well instead of tunneling through the internet. Currently communicating directly over BLE also requires persistent pairing for both parties in CTAP (which makes a lot of sense in the scenarios authenticators are currently used). 

---

I think this would make sense:

- OpenID4VP defines a profile for flows over a secure transport protocol (without expecting any kind of online connection). This should be pretty similar to the current DC API profile. We could adjust the current DC API profile to allow for the different modes of wallet invocation / origins and then have 3 subsections for:
  - Invocation over browser API (origin = website as provided by the browser) 
  - Invocation by another APP (origin = origin app)
  - Invocation via CTAP "offline" (origin = physical transport; not sure what to best put here - just an identifier for the used transport protocol?)
- CTAP defines how to create an ad-hoc secure transport channel (without requiring pairing) and defines how the OpenID4VP messages are transmitted (with some kind of identifier for the protocol) - this could just be an extension to hybrid mode with new modes
  - additional initiation: NFC
  - additional transport: NFC(?) - do we want this?
  - additional transport: BLE
  ... Wi-Fi direct, UWB, or whatever technologies might make sense

[...]

PR #337, opened 2024-11-20
Generalize W3C Digital Credentials API references

timcappalli:
Based on the discussion on yesterday's call, this changes the following:

- Adds a new term in Terminology for "The Digital Credentials API" which talks about both the Web Platform API (W3C) and the native platform equivalents (generically)
- Changes most occurrences of "W3C Digital Credentials API" to either "Digital Credentials API (DC API)" or "DC API"
- Changes the `response_mode` to `dc_api`
- Includes "app origin" along with web origin
- Fixes xref for W3C DC API

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1963,7 +1963,7 @@ This section defines a profile of OpenID4VP for use with the Digital Credentials
 
 The name "Digital Credentials API" (DC API) encomposes the W3C Digital Credentials API [@!W3C.Digital_Credentials_API]
 as well as its native App Platform equivalents in operating systems (such as [Credential Manager on Android](https://developer.android.com/jetpack/androidx/releases/credentials)).
-The DC API allows web sites and apps acting as Verifiers to request the presentation of verifiable credentials.
+The DC API allows web sites and native apps acting as Verifiers to request the presentation of verifiable credentials.
 The API itself does not define a Credential exchange protocol but can be used with multiple protocols.
 The Web Platform, working in conjunction with other layers, such as the app platform/operating system, and based on the permission of the End-User, will send the request data along with the web or app origin of the Verifier to the End-User's chosen Wallet.
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -558,7 +558,7 @@ Body
 
 * `x509_san_uri`: When the Client Identifier Scheme is `x509_san_uri`, the Client Identifier MUST be a URI and match a `uniformResourceIdentifier` Subject Alternative Name (SAN) [@!RFC5280] entry in the leaf certificate passed with the request. The request MUST be signed with the private key corresponding to the public key in the leaf X.509 certificate of the certificate chain added to the request in the `x5c` JOSE header [@!RFC7515] of the signed request object. The Wallet MUST validate the signature and the trust chain of the X.509 certificate. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter. If the Wallet can establish trust in the Client Identifier authenticated through the certificate, e.g., because the Client Identifier is contained in a list of trusted Client Identifiers, it may allow the client to freely choose the `redirect_uri` value. If not, the `redirect_uri` value MUST match the Client Identifier without the prefix `x509_san_uri:`. Example Client Identifier: `x509_san_uri:https://client.example.org/cb`.
 
-* `web-origin`: This Client Identifier Scheme is defined in (#browser_api_request). It MUST NOT be used in the Authorization Request.
+* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). It MUST NOT be used in the Authorization Request.
 
 To use the Client Identifier Schemes `https`, `did`, `verifier_attestation`, `x509_san_dns`, and `x509_san_uri`, Verifiers MUST be confidential clients. This might require changes to the technical design of native apps as such apps are typically public clients.
 
@@ -1984,7 +1984,7 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 For the profile defined in this section, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
 
-## Request {#browser_api_request}
+## Request {#dc_api_request}
 
 The Verifier MAY send all the OpenID4VP request parameters to the Digital Credentials API (DC API).
 
@@ -2036,7 +2036,7 @@ The Verifier MAY send all the OpenID4VP request parameters as members in the req
 
 The Verifier MAY send a signed request, for example, when identification and authentication of the Verifier is required.
 
-The signed Request Object MAY contain all the parameters listed in (#browser_api_request), except `request`.
+The signed Request Object MAY contain all the parameters listed in (#dc_api_request), except `request`.
 
 Below is a non-normative example of such a request sent over the Digital Credentials API (DC API):
 
@@ -2528,7 +2528,7 @@ established by [@!RFC6749].
 * Name: `expected_origins`
 * Parameter Usage Location: authorization request
 * Change Controller: OpenID Foundation Artifact Binding Working Group - openid-specs-ab@lists.openid.net
-* Reference: (#browser_api_request) of this specification
+* Reference: (#dc_api_request) of this specification
 
 ## OAuth Extensions Error Registry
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2005,18 +2014,18 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
...
+The value of the `response_mode` parameter MUST be `dc_api` when the response is neither signed nor encrypted and `dc_api.jwt` when the response is signed and/or encrypted as defined in (#jarm).
 
 In addition to the above-mentioned parameters, this profile introduces a new parameter:
 
-* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the W3C Digital Credentials API [@!w3c.digital_credentials_api]. An array of strings, each string representing an origin of the Verifier that is making the request. The Wallet can detect replay of the request from a malicious Verifier by comparing values in this parameter to the origin asserted by the Web Platform.
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). An array of strings, each string representing an origin of the Verifier that is making the request. The Wallet can detect replay of the request from a malicious Verifier by comparing values in this parameter to the origin asserted by the Web Platform.
```
leecam, line 2021: Do we need to call out these can be web origins or native app origins?

leecam (commented):
Left one comment about the origin but this looks great to me!

openid-4-verifiable-presentations-1_0.md
timcappalli: I added app origins earlier in the section:

Line 1968:
> The Web Platform, working in conjunction with other layers, such as the app platform/operating system, and based on the permission of the End-User, will send the request data along with the web or app origin of the Verifier to the End-User's chosen Wallet.

But can be more explicit here if you think it isn't clear.

leecam: I did see that yeah (looks great) but I think it wouldn't hurt to clarify in this section too. I think most folks think of an origin only as a web origin so saying "representing a web origin or native app origin..." might help

timcappalli: Would the client ID schemes need to be updated as well? 

> * `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). It MUST NOT be used in the Authorization Request.

https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-5.10.4

Should there be two? `web-origin` and `app-origin` or should it just be generalized to `origin`?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -73,6 +73,9 @@ This specification also defines the following terms. In the case where a term ha
 Credential:
 :  A set of one or more claims about a subject made by a Credential Issuer. Note that the definition of the term "Credential" in this specification is different from that in [@!OpenID.Core].
 
+Digital Credentials API:
+:  The Digital Credentials API (DC API) refers to the W3C Digital Credentials API [@!W3C.Digital_Credentials_API] on the Web Platform and its equivalent native APIs on App Platforms (such as [Credential Manager on Android](https://developer.android.com/jetpack/androidx/releases/credentials)).
```
Sakurann, line 77: is it ok to point to a specific implementation in the definition..? re `Credential Manager on Android`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1951,19 +1957,22 @@ In the event that another component is invoked instead of the Wallet, the End-Us
...
+This section defines a profile of OpenID4VP for use with the Digital Credentials API.
 
-The W3C Digital Credentials API defines a Web Platform API which allows web sites acting as Verifiers
-to request the presentation of Verifiable Credentials. The API itself does not define a Credential exchange protocol
-but can be used with multiple protocols. The Web Platform, working in conjunction with other layers, such as the app platform/operating system, and based on the permission of the End-User, will send the request data along with the web origin of the Verifier to the End-User's chosen Wallet.
+The name "Digital Credentials API" (DC API) encomposes the W3C Digital Credentials API [@!W3C.Digital_Credentials_API]
+as well as its native App Platform equivalents in operating systems (such as [Credential Manager on Android](https://developer.android.com/jetpack/androidx/releases/credentials)).
+The DC API allows web sites and apps acting as Verifiers to request the presentation of verifiable credentials.
+The API itself does not define a Credential exchange protocol but can be used with multiple protocols.
```
Sakurann, line 1967:
```suggestion
The API itself is agnostic to the Credential exchange protocol and can be used with different protocols.
```

Sakurann (approved):
looks good - minor nits.

openid-4-verifiable-presentations-1_0.md
timcappalli: @Sakurann thoughts on this? ^

openid-4-verifiable-presentations-1_0.md
jogu: I think it's probably okay, it does seem useful to point developers in the right directions. Ideally it would be good to include more than one (Apple?) but is Google the only one that has written anything public right now?

We can always add more references for other OSes later, potentially even in errata after we've gone final.

openid-4-verifiable-presentations-1_0.md
jogu: Is there a concrete example of the likely formats that the app origins will arrive in? It seems like it might be important to ensure that web origins and app origins across all OSes (Windows/Mac/iOS/Android/linux/whatever) cannot be confused with each other? e.g. a rogue app installed on one OS shouldn't be able to end up impersonating an actual app on a different OS.

c2bo: We need to differentiate the different origins (and with the possible addition of "raw" CTAP flows as discussed in #305, we will likely have a third origin type/option for offline interactions with readers). Whether we do that via some kind of prefix and have one parameter `origin` or we have different origin types doesn't matter too much imho as long as we are able to detect & distinguish those properly.

timcappalli: > We need to differentiate the different origins (and with the possible addition of "raw" CTAP flows as discussed in #305, we will likely have a third origin type/option for offline interactions with readers). Whether we do that via some kind of prefix and have one parameter `origin` or we have different origin types doesn't matter too much imho as long as we are able to detect & distinguish those properly.

In WebAuthn, there is a single `origin` value, which can be a web or app origin.

c2bo: > > We need to differentiate the different origins (and with the possible addition of "raw" CTAP flows as discussed in #305, we will likely have a third origin type/option for offline interactions with readers). Whether we do that via some kind of prefix and have one parameter `origin` or we have different origin types doesn't matter too much imho as long as we are able to detect & distinguish those properly.
> 
> In WebAuthn, there is a single `origin` value, which can be a web or app origin.

Is my understanding correct that the origin values for WebAuthn would look like this?
- android:apk-key-hash:<some hash>
- ios:bundle-id:<some id>
- https://example.com

and you would parse this via prefix matching?

c2bo (approved):
Not entirely sure what the best approach for the origin is, but given that webauthn already uses one `origin` parameter  for browser-initiated or native app calls, it feels like that would be the logical way to go as well?

Looks good to me once the origin discussion is resolved.

openid-4-verifiable-presentations-1_0.md
timcappalli: Apple platforms use web origin. 

Here's the exact text from the security considerations of WebAuthn: https://w3c.github.io/webauthn/#sctn-validating-origin

> A web application with a companion native application might allow [origin](https://w3c.github.io/webauthn/#dom-collectedclientdata-origin) to be an operating system dependent identifier for the native application. For example, such a [Relying Party](https://w3c.github.io/webauthn/#relying-party) might require that [origin](https://w3c.github.io/webauthn/#dom-collectedclientdata-origin) exactly equals some element of the list `["https://example.org", "example-os:appid:204ffa1a5af110ac483f131a1bef8a841a7adb0d8d135908bbd964ed05d2653b"]`.

leecam: Yeah you don't actually need to parse the origin strings as a verifier. You just need to string match them against a static list of approved/expected origins. Logically you can think of them as not having structure or any machine readable information within them. They are just opaque strings that the OS/Browser provides to the verifier in a consistent and trusted way.

So in that sense we probably don't need call out the different variants in the spec? Just to call out that origin here means more than just web-origin. It can be any platform provided origin string depending on the client context. I think its fine to give some examples, such as web origins and android app a origins.

If the word origin is too loaded, maybe we could say "originating-context" to be more general.

c2bo: I agree the verifier doesn't need to understand - just checking that it matches expected values seems fine there. I was thinking more in the direction that depending on circumstances, this might be something that should/might be shown to the user (on a credential selector / wallet level)? Not entirely sure tbh - maybe not necessary?

Sakurann: discussed in the WG. would like to merge this PR to get it into the implementers draft, but will turn this into an issue to make sure there is consensus.

leecam (approved):

tlodderstedt (approved):

danielfett (approved):

openid-4-verifiable-presentations-1_0.md
Sakurann: #351

PR #338, opened 2024-11-20
add intent to retain to claims query in DCQL

### LINKED ISSUE #321: Add intent_to_retain to DCQL
The presentation exchange supports the intent_to_retain parameter to indicate whether the RP intends to retain the received data element / claim. A similar parameter should be added to the DCQL.

A solution would be to add it to the claims query, either as a generic option or as an mdoc specific option.

Sakurann:
resolves #321


openid-4-verifiable-presentations-1_0.md
```diff
@@ -740,6 +740,9 @@ in the mdoc, e.g., `first_name`.
 If the `values` property is present, the Wallet SHOULD return the claim only if the
 type and value of the claim both match for at least one of the elements in the array. Details of the processing rules are defined in (#selecting_claims).
 
+`intent_to_retain`
+: OPTIONAL. A boolean that indicates the Verifier intends to retain the Claim's data that is being requested.
```
danielfett, line 744:
```suggestion
: OPTIONAL. A boolean that indicates the Verifier intends to store the Claim's data that is being requested beyond the processing of the response, for example, for purposes of recording a proof of the transaction or to maintain a user profile.
```
Not sure if this is the best wording, but we should try to explain what the 'retain' in `intent_to_retain` means.

nemqe (approved):

paulbastian (dismissed):
I'm not in favor of adding this and I think there have been good reasons/argumentations in the Issue. In short, the feature as-is is under-specified and likely of not much use and it seems more reasonable to convey this via the `purpose` field.

Sakurann:
WG discussion: agreement that this is not perfect (PR text should be improved), but probably good enough as a starting point to meet the requirements and have a transition path for those who us intent_to_retain in 18013-5/PE.

nemqe:
A potential issue I see is that if this option is added in this form it will probably need to be maintained in the version it is introduced as it could be a breaking change (until a parallel mechanism is introduced and the option is marked for removal in the next version)

Other than that if the group is short on time and this option is required it may be a good compromise until a better mechanism is found.

Sakurann:
> if this option is added in this form it will probably need to be maintained in the version it is introduced as it could be a breaking change to remove it

the feature is `optional`. I don't think we plan to remove it, but if we come up with a better mechanism in v1.1 in the future, we can recommend that feature over intent_to_retain.

another option I considered when doing PR is to add intent_to_retain only to mdocs, but that did not feel in sprit of the specification, or the discussions we have been having..

openid-4-verifiable-presentations-1_0.md
```diff
@@ -744,6 +744,9 @@ within the Verifiable Credential, as defined in (#claims_path_pointer).
 If the `values` property is present, the Wallet SHOULD return the claim only if the
 type and value of the claim both match for at least one of the elements in the array. Details of the processing rules are defined in (#selecting_claims).
 
+`intent_to_retain`
+: OPTIONAL. A boolean that indicates the Verifier intends to store the Claim's data that is being requested beyond the processing of the response. For example, to meet regulatory requirements of retaining records of transactions or to maintain user profiles. If omitted, the default value is `false`.
```
Sakurann, line 748:
```suggestion
: OPTIONAL. A boolean that indicates the Verifier intends to store the Claim's data that is being requested beyond the processing of the response. For example, to meet regulatory requirements of retaining records of transactions or to maintain user profiles.
```

Sakurann: + move it to the mdoc section as a starting point.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2243,6 +2243,23 @@ The following is an ISO mdoc specific parameter in the `meta` parameter in a Cre
...
+`namespace`:
+: REQUIRED if the Credential Format is based on the mdoc format defined in [@ISO.18013-5]; MUST NOT be present otherwise. 
+The value MUST be a string that specifies the namespace of the data element
+within the mdoc, e.g., `org.iso.18013.5.1`.
+
+`claim_name`:
+: REQUIRED if the Credential Format is based on mdoc format defined in [@ISO.18013-5]; MUST NOT be present otherwise. 
+The value MUST be a string that specifies the data element identifier of the data element within the provided namespace
+in the mdoc, e.g., `first_name`.
+
```
charsleysa, line 2259: These parameters were removed as part of https://github.com/openid/OpenID4VP/pull/368 that has been merged

```suggestion
```

peppelinux (approved):
good one!

c2bo (approved):
I am still not fully convinced that this feature on its own makes sense, but as it is an optional credential format specific feature that just matches the current capabilities in the ISO protocols, I guess it is fine to merge.

We can (and probably should imho) provide another solution in the direction of purpose in the long run.

awoie (approved):

awoie (commented):
IMO, this would cause interop issues. I know that certain members are in favor of providing some guidance to the user but it would cause confusion if this was done only for mdocs.

tlodderstedt (approved):

PR #352, opened 2024-11-28
change PE version to 2.1.1

### LINKED ISSUE #231: Advice on the presentation exchange protocol version
In the oid4vp spec, every reference to the presentation exchange protocol seems to point to the [presentation exchange version 2.0](https://identity.foundation/presentation-exchange/spec/v2.0.0/). For implementers, I am not sure what is the semantic of this version. Does this means that any implementation should pin the presentation exchange protocol to version 2.0.0. I'm asking this because there is a newer version 2.1.1 now and whether this should be bumped in the oid4vp spec or pinned to specific version.

If it should be bumped in the oid4vp spec, what would be the mechanism to discover what version the implementation is using? :thinking:

Sakurann:
resolves #231


jogu:
I couldn't find any list of the changes between v2.0.0 and v2.1.1, but I did find a list of changes between 1.x and 2.1.1: 

https://identity.foundation/presentation-exchange/spec/v2.1.1/#what-is-new

And I think (if you scroll down to `spec/spec.md` and expand it) this is the actual changes between the two versions:

https://github.com/decentralized-identity/presentation-exchange/compare/b747b897144172f29ff1c331c568989c5d1174a0...7cbe949c95fe1e19413b24c9b49dc34f76f5d76a#diff-c4795c497b83a8c03e33535caf0fb0e1512cecd8cb448f62467326277c152afe

In particular the only thing that was added to the changelog between 2.0.0 and 2.1.1 was "Introduction of an optional `name` property for the [[ref:Input Descriptor Objects]]."

So overall this doesn't really look like it's a scary change.

jogu (approved):

Sakurann:
yes, I don't think DIF introduces breaking changes to the specs from 2.x to 2.y

nemqe (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -948,14 +948,14 @@ When a VP Token is returned, the respective response includes the following para
...
+ * In case [@!DIF.PresentationExchange] was used, it is a JSON String or JSON object that MUST contain a single Verifiable Presentation or an array of JSON Strings and JSON objects each of them containing a Verifiable Presentations. Each Verifiable Presentation MUST be represented as a JSON string (that is a base64url-encoded value) or a JSON object depending on a format as defined in (#format_specific_parameters).  When a single Verifiable Presentation is returned, the array syntax MUST NOT be used.  If (#format_specific_parameters) defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the `vp_token` response parameter. Otherwise, this specification does not require any additional encoding when a Credential format is already represented as a JSON object or a JSON string.
```
nklomp, line 951:
```suggestion
 * In case [@!DIF.PresentationExchange] was used, it is a JSON String or JSON object that MUST contain a single Verifiable Presentation or an array of JSON Strings and JSON objects each of them containing a Verifiable Presentation. Each Verifiable Presentation MUST be represented as a JSON string (that is a base64url-encoded value) or a JSON object depending on a format as defined in (#format_specific_parameters).  When a single Verifiable Presentation is returned, the array syntax MUST NOT be used.  If (#format_specific_parameters) defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the `vp_token` response parameter. Otherwise, this specification does not require any additional encoding when a Credential format is already represented as a JSON object or a JSON string.
```

nklomp (approved):
Small editorial comment

PR #355, opened 2024-11-30
Require typ in request objects

### LINKED ISSUE #268: Require JAR typ value to be used in request objects
https://www.rfc-editor.org/rfc/rfc9101.html defines a `typ` value to be used in the header of request objects, `oauth-authz-req+jwt`, but doesn't go as far as requiring it to be used, offering various cases where it can be omitted or the value `JWT` used instead for historical compatibility. But does say:

> requiring explicit typing would be a good idea for new OAuth deployment profiles where compatibility with existing deployments is not a consideration.

Given any existing code or deployments is going to need a large number of changes to become a wallet or verifier, there doesn't seem to be any need for historical compatibility with existing deployment - hence it would be a "good idea" for VP to require explicit typing of request objects, and that should ensure there's never any possibility that request objects are confused with any other kind of object.

I suggest we add text along the lines:

> Verifiers MUST include the `typ` header in request objects with the value defined in RFC9101, `oauth-authz-req+jwt`. Wallets MUST reject request objects where the `typ` header is not present or does not have the value `oauth-authz-req+jwt`

(Mentioned by Kristina, as I believe this caused an interoperability problem for someone testing with the conformance suite - the conformance suite currently doesn't send the `typ` header as it's not required by the spec.)

jogu:
closes #268

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -246,7 +246,7 @@ Presentation of Verifiable Credentials using OpenID for Verifiable Presentations
 
 The Authorization Request follows the definition given in [@!RFC6749] taking into account the recommendations given in [@!I-D.ietf-oauth-security-topics].
 
-The Verifier MAY send an Authorization Request as a Request Object either by value or by reference, as defined in the JWT-Secured Authorization Request (JAR) [@RFC9101].
+The Verifier MAY send an Authorization Request as a Request Object either by value or by reference, as defined in the JWT-Secured Authorization Request (JAR) [@RFC9101]. Verifiers MUST include the `typ` Header Parameter in Request Objects with the value defined in [@RFC9101], `oauth-authz-req+jwt`. Wallets MUST NOT process Request Objects where the `typ` Header Parameter is not present or does not have the value `oauth-authz-req+jwt`.
 
 This specification defines a new mechanism for the cases when the Wallet wants to provide to the Verifier details about its technical capabilities to
 allow the Verifier to generate a request that matches the technical capabilities of that Wallet.
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2709,6 +2709,7 @@ The technology described in this specification was made available from contribut
 
    -24
 
+   * require `typ` value in request object to be `oauth-authz-req+jwt`
 
    -23
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -246,7 +246,7 @@ Presentation of Verifiable Credentials using OpenID for Verifiable Presentations
 
 The Authorization Request follows the definition given in [@!RFC6749] taking into account the recommendations given in [@!I-D.ietf-oauth-security-topics].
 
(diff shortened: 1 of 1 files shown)


Sakurann (commented):
this is a text from RFC9101, which is much more relaxed than the current PR text. I understand that mandating `oauth-authz-req+jwt` helps with the interop, but I think we need to discuss implications on the implementations. For example, I am aware of the implementations that use `jwt` as a `typ` value.

> Another way to prevent cross-JWT confusion is to use explicit typing, as described in [Section 3.11](https://www.rfc-editor.org/rfc/rfc8725#section-3.11) of [[RFC8725](https://datatracker.ietf.org/doc/html/rfc8725)]. One would explicitly type a Request Object by including a typ Header Parameter with the value oauth-authz-req+jwt (which is registered in [Section 9.4.1](https://datatracker.ietf.org/doc/html/rfc9101#MediaContents)). Note, however, that requiring explicitly typed Request Objects at existing authorization servers will break most existing deployments, as existing clients are already commonly using untyped Request Objects, especially with OpenID Connect [[OpenID.Core](http://openid.net/specs/openid-connect-core-1_0.html)]. However, requiring explicit typing would be a good idea for new OAuth deployment profiles where compatibility with existing deployments is not a consideration

jogu:
That text in RFC9101 was there to not breaking existing OpenID Connect implementations (from what I remember the OAuth WG weren't keen to allow that exception) - as I mentioned on the issue no one can use an existing OpenID Connect implementation to request credentials, and we already have other breaking changes happening (E.g. around DCQL) there doesn't seem to be any large issue with following the JWT BCP and doing explicit required typing, and hence requiring implementations to update.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -246,7 +246,7 @@ Presentation of Verifiable Credentials using OpenID for Verifiable Presentations
 
 The Authorization Request follows the definition given in [@!RFC6749] taking into account the recommendations given in [@!I-D.ietf-oauth-security-topics].
 
-The Verifier MAY send an Authorization Request as a Request Object either by value or by reference, as defined in the JWT-Secured Authorization Request (JAR) [@RFC9101].
+The Verifier MAY send an Authorization Request as a Request Object either by value or by reference, as defined in the JWT-Secured Authorization Request (JAR) [@RFC9101]. Verifiers MUST include the `typ` Header Parameter in Request Objects with the value defined in [@RFC9101], `oauth-authz-req+jwt`. Wallets MUST NOT process Request Objects where the `typ` Header Parameter is not present or does not have the value `oauth-authz-req+jwt`.
```
paulbastian, line 249:
```suggestion
The Verifier MAY send an Authorization Request as a Request Object either by value or by reference, as defined in the JWT-Secured Authorization Request (JAR) [@RFC9101]. Verifiers MUST include the `typ` Header Parameter in Request Objects with the value `oauth-authz-req+jwt`, as defined in [@RFC9101]. Wallets MUST NOT process Request Objects where the `typ` Header Parameter is not present or does not have the value `oauth-authz-req+jwt`.
```

paulbastian (approved):

selfissued (approved):

deshmukhrajvardhan (approved):

nemqe (approved):

c2bo (approved):

babisRoutis (approved):

PR #356, opened 2024-11-30
Require wallets to ignore client_id if it's present in unsigend DC API requests

### LINKED ISSUE #276: Clarify behaviour of client_id with unsigned browser requests
The browser API appendix currently says:

> The client_id parameter MUST be omitted in unsigned requests defined in

I think it'd be helpful to specify the behaviour the wallet needs to have. The obvious options are that the wallet should either return an error, or it should ignore the client_id. Specifying the behaviour would allow it to be tested for in the conformance tests.

I'd suggest saying something like:

> Wallets MUST ignore `client_id` if it is included.

jogu:
closes #276

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1978,7 +1978,7 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
 * `transaction_data`
 * `dcql_query`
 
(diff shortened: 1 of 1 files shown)


martijnharing:
Shouldn't we also extend this to scenarios where the wallet does not verify the signature but still decides to continue, e.g. because it cannot, because it doesn't want to or another reason.

bc-pi (approved):

jogu:
> Shouldn't we also extend this to scenarios where the wallet does not verify the signature but still decides to continue, e.g. because it cannot, because it doesn't want to or another reason.

Hmm. Good question. I don't think the VP spec in general caters for this in general. e.g. There's language like this:

> The Wallet MUST validate the signature and the trust chain of the X.509 certificate.

I think we should discuss it separately - I've opened https://github.com/openid/OpenID4VP/issues/362 - let's continue the discussion on that issue.

Sakurann (approved):
we need to discuss more the point Martijn is raising - let's merge this one first, without expanding its scope.

c2bo (approved):

paulbastian (approved):

PR #360, opened 2024-12-02
Clarify that non-JARM error response may be sent even when JARM requested

### LINKED ISSUE #135: Clarify JWT envelope for direct_post.jwt in error cases
`direct_post.jwt` requires the wallet to send a JWT to the `response_uri`. This also applies to an error response. In this case, the error parameters are included in a JWT.

What happens in the following case:

A specific OID4VP profile requires encrypting the response (using JARM) but the provided client metadata does not contain encryption keys, so the response cannot be encoded as a JWE. 

- Should the wallet not send a response at all? Probably safest.
- Should the wallet send the response with alg=none? Potentially dangerous.
- Should the wallet sign the response, and if so what key should be used? Probably makes no sense.

Do you think we should add this to the implementation considerations? IMO, this would be useful.

jogu:
Even if the Verifier requests direct_post.jwt, the Wallet may be unable to do so for various reasons including the Verifier not supplying keys or encryption algorithms. Allow a direct_post error response to allow for those and other cases.

closes #135

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1144,6 +1144,8 @@ This specification also defines a new Response Mode `direct_post.jwt`, which all
 
 The Response Mode `direct_post.jwt` causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in (#response_mode_post). The Wallet adds the `response` parameter containing the JWT as defined in Section 4.1. of [@!JARM] and (#jarm) in the body of an HTTP POST request using the `application/x-www-form-urlencoded` content type. The names and values in the body MUST be encoded using UTF-8.
 
+If a Wallet is unable to send a JARM response it MAY send an error response as per (#response_mode_post).
+
 The following is a non-normative example of a response using the `presentation_submission` and `vp_token` values from (#jwt_vc). (line breaks for display purposes only):
 
 <{{examples/response/jarm_jwt_vc_json_post.txt}}
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1144,7 +1144,7 @@ This specification also defines a new Response Mode `direct_post.jwt`, which all
 
 The Response Mode `direct_post.jwt` causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in (#response_mode_post). The Wallet adds the `response` parameter containing the JWT as defined in Section 4.1. of [@!JARM] and (#jarm) in the body of an HTTP POST request using the `application/x-www-form-urlencoded` content type. The names and values in the body MUST be encoded using UTF-8.
 
-If a Wallet is unable to send a JARM response it MAY send an error response as per (#response_mode_post).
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1144,6 +1144,8 @@ This specification also defines a new Response Mode `direct_post.jwt`, which all
 
 The Response Mode `direct_post.jwt` causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in (#response_mode_post). The Wallet adds the `response` parameter containing the JWT as defined in Section 4.1. of [@!JARM] and (#jarm) in the body of an HTTP POST request using the `application/x-www-form-urlencoded` content type. The names and values in the body MUST be encoded using UTF-8.
 
+If a Wallet is unable to send a JARM response it MAY send an error response as per (#response_mode_post).
```
Sakurann, line 1147: [Response Mode "direct_post"](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-response-mode-direct_post) does not really define error responses clearly....

I might have misunderstood the discussion in the issue, but I think this sentence should be clearer?

```suggestion
If a Wallet is unable to generate a JARM response, it MAY send an error response without using JARM.
```

jogu: > [Response Mode "direct_post"](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-response-mode-direct_post) does not really define error responses clearly....

It is where the non-JARM way of sending responses is defined, and It has an example of an error being sent at least. I'd prefer to keep the reference to make it clear what is meant. Happy to change send to generate though, does this work for you:

```suggestion
If a Wallet is unable to generate a JARM response, it MAY send an error response as per (#response_mode_post).
```

Sakurann: umm, i really think it should be explicitly stated that an error response can be without using JARM, which I think is the point of this PR, right>

jogu: Yes - that is the point, the `(#response_mode_post)` is a link to the non-JARM response definition.

awoie (approved):

openid-4-verifiable-presentations-1_0.md
Sakurann: yes, so if that's the point of that PR, I would like to see the language to be more explicit

jogu:
```suggestion
If a Wallet is unable to generate a JARM response, it MAY send a non-JARM error response as per (#response_mode_post).
```

@Sakurann does explicitly mentioning non-JARM like this this solve it?

Sakurann: what is a non-JARM error response? why can't we say "error response without using JARM."

jogu:
```suggestion
If a Wallet is unable to generate a JARM response, it MAY send an error response without using JARM as per (#response_mode_post).
```

Changed to use Kristina's suggested wording.

Sakurann (approved):

PR #368, opened 2024-12-12
Use `path` for mdoc

danielfett:


openid-4-verifiable-presentations-1_0.md
```diff
@@ -827,33 +825,36 @@ these constraints. The Wallet is not controlled by the Verifier and the Verifier
...
 
-## Claims Path Pointer {#claims_path_pointer}
+# Claims Path Pointer {#claims_path_pointer}
+
+A claims path pointer is a pointer into the Verifiable Credential, identifying one or more claims.
+A claims path pointer MUST be a non-empty array of strings, nulls and non-negative integers.
+A claims path pointer can be processed, which means it is applied to a credential. The results of
+processing are the referenced claims.
+
+## Semnatics for JSON-based credentials
```
danielfett, line 835:
```suggestion
## Semantics for JSON-based credentials
```

Sakurann (approved):

c2bo (approved):

bc-pi (approved):

leecam (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -867,15 +868,36 @@ In detail, the array is processed by the Wallet from left to right as follows:
...
        return an error. If the index does not exist in a selected array, remove
        that array from the selection.
-  3. If the set of elements currently selected is empty, abort processing and
-     return an error.
+3. If the set of elements currently selected is empty, abort processing and
+   return an error.
+
+The result of the processing is the set of selected JSON elements.
+
+## Semantics for ISO mdoc-based credentials
```
paulbastian, line 876: Why is this section here and not in the mdoc credential format appendix? It's explicitly for ISO mdoc, so I don't understand why it should be here.

babisRoutis:
A question from my side:  In the mdoc case, is it valid to have a path with a single element (the namespace)?

If not, should somehow this constraint appear on the spec in the mdoc section?

Sakurann:
> A question from my side: In the mdoc case, is it valid to have a path with a single element (the namespace)?
>
> If not, should somehow this constraint appear on the spec in the mdoc section?

my understanding is that path for mdoc always has at least 2 elements, so we should make that clear

babisRoutis (approved):

openid-4-verifiable-presentations-1_0.md
Sakurann: i think for the symmetry with JSON-based credentials section. will merge as-is since there are enough approvals and no other blockers. we can fix this later if necessary, as we did with the mdoc section in the very first DCQL PR.

Issue #371, opened 2024-12-12
Encryption details for OpenID4VP over the digital credentials API


Sakurann:
copy from https://github.com/openid/oid4vc-haip/issues/131#issuecomment-2539528136

WG discussion:
- agreement to what was said in the end of the last call: **a viable way forward now seems to be to start with option 1 with what we have to meet the deadlines while signaling that we are working on HPKE and add HPKE once it is ready, since that is more desirable**.
    - noting that @martijnharing said they believe HPKE is the better algorithm, better library support, easier, can be implemented in a way that is more secure. (no agreement on this)
- @awoie to open an issue in IETF JOSE-HPKE repo why detached AAD is needed (and say it applies to JWE with ECDH-ES too). should not define that in this WG.
- lazy verifier problem does not apply for OpenID4VP over the digital credentials API?
- @martijnharing said he would like this feature to make the protocol more robust and more secure. agreement that this is not catch-all solution to lazy verifier problem.
- @ve7jtb if the alg supports detached AAD, it should be used to help with lazy verifier check - solution might work with ECDH and HPKE.

next step (@bc-pi volunteered): 
- do a PR in OID4VP to say "use JWE with ECDH-ES". that PR will also say parameters in AAD that currently go into the protected header (such as `apv` and `apu`), and one of them has to be nonce and the other client_id (or origin when DC API is not signed). will try frame it in a way that does not lose alg agility once detached AAD becomes available
- will need a small PR in HAIP mandating a specific alg/curve, once we agree on it. (https://github.com/openid/oid4vc-haip/issues/112)

PR #380, opened 2025-01-10
discuss the use of apu/apv in the JWE header of OpenID4VP encrypted responses

bc-pi:
Touching 3 & 5 of https://docs.google.com/document/d/1AJDDWuRG_b-MOBrAwhBoQV3dhH3LD31WNEQKzOB36SY/edit?tab=t.0 and https://github.com/openid/OpenID4VP/issues/371

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -7,7 +7,7 @@ keyword = ["security", "openid", "ssi"]
 
 [seriesInfo]
 name = "Internet-Draft"
-value = "openid-4-verifiable-presentations-1_0-24"
+value = "openid-4-verifiable-presentations-1_0-25"
 status = "standard"
 
 [[author]]
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1150,14 +1150,8 @@ If the JWT is only a JWE, the following processing rules MUST be followed:
 - `iss`, `exp` and `aud` MUST be omitted in the JWT Claims Set of the JWE, and the processing rules as per [@!JARM] Section 2.4 related to these claims do not apply.
 - The processing rules as per [@!JARM] Section 2.4 related to JWS processing MUST be ignored.
 
-Whenever the response is encrypted, the following additional processing directives apply to JWE. This enables binding response encryption to the Verifier's identifier and its respective transaction with the End-User:
-
-- The `apu` (Agreement PartyUInfo) header parameter MUST be set to the value of the `nonce` from the Authorization Request.
-- The `apv` (Agreement PartyVInfo) header parameter MUST be set to the value of the `client_id` of the Verifier.
-
 Note that for the ECDH JWE algorithms (from section 4.6 of [@!RFC7518]), the `apu` and `apv` values are inputs
-into the key derivation process that is used to derive the content encryption key. However, regardless of algorithm used, the values are always part of the AEAD tag computation so will still be bound to the encrypted response.
-
+into the key derivation process that is used to derive the content encryption key. Regardless of algorithm used, the values are always part of the AEAD tag computation so will still be bound to the encrypted response.
 The following is a non-normative example of the payload of a JWT used in an Authorization Response that is encrypted and not signed:
 
 <{{examples/response/jarm_jwt_enc_only_vc_json_body.json}}
@@ -2775,7 +2769,7 @@ The technology described in this specification was made available from contribut
 
    -24
 
-   * Mandate the use of apu/apv in the JWE header of encrypted responses
+   * add a note on the use of apu/apv in the JWE header of encrypted responses
    * require `typ` value in request object to be `oauth-authz-req+jwt`
    * add `SessionTranscript` requirements 
    * use claims path pointer for mdoc based credentials
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1128,7 +1128,7 @@ If the JWT is only a JWE, the following processing rules MUST be followed:
 - `iss`, `exp` and `aud` MUST be omitted in the JWT Claims Set of the JWE, and the processing rules as per [@!JARM] Section 2.4 related to these claims do not apply.
 - The processing rules as per [@!JARM] Section 2.4 related to JWS processing MUST be ignored.
 
-Whenever the response is encrypted, the following additional processing directives apply to bind the Verifier's identifier and its respective transaction with the End-User into the JWE:
+Whenever the response is encrypted, the following additional processing directives apply to JWE. This enables binding response encryption to the Verifier's identifier and its respective transaction with the End-User:
 
 - The `apu` (Agreement PartyUInfo) header parameter MUST be set to the value of the `nonce` from the Authorization Request.
 - The `apv` (Agreement PartyVInfo) header parameter MUST be set to the value of the `client_id` of the Verifier.
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1119,7 +1119,7 @@ This section defines how an Authorization Response containing a VP Token can be 
 
 To sign, encrypt, or both sign and encrypt the Authorization Response, implementations MUST use the JWT Secured Authorization Response Mode for OAuth 2.0 (JARM) [@!JARM], and when only encrypting, the JARM extension described below.
 
-This specification also defines how to encrypt an unsigned Authorization Response by extending the mechanisms defined in [@!JARM]. The JSON containing the Authorization Response parameters can be encrypted as the payload of the JWE.
+This specification also defines how to encrypt an unsigned Authorization Response by adapting the mechanisms defined in [@!JARM]. The JSON containing the Authorization Response parameters can be encrypted as the payload of the JWE.
 
 The advantage of an encrypted but not signed Authorization Response is that it prevents the signing key from being used as a correlation factor. It can also be a challenge to establish trust in the signing key to ensure authenticity. For security considerations with encrypted but unsigned responses, see (#encrypting_unsigned_response).
 
@@ -1127,6 +1127,14 @@ If the JWT is only a JWE, the following processing rules MUST be followed:
 
 - `iss`, `exp` and `aud` MUST be omitted in the JWT Claims Set of the JWE, and the processing rules as per [@!JARM] Section 2.4 related to these claims do not apply.
 - The processing rules as per [@!JARM] Section 2.4 related to JWS processing MUST be ignored.
+
+Whenever the response is encrypted, the following additional processing directives apply to bind the Verifier's identifier and its respective transaction with the End-User into the JWE:
+
+- The `apu` (Agreement PartyUInfo) header parameter MUST be set to the value of the `nonce` from the Authorization Request.
+- The `apv` (Agreement PartyVInfo) header parameter MUST be set to the value of the `client_id` of the Verifier.
+
+Note that for the ECDH JWE algorithms (from section 4.6 of [@!RFC7518]), the `apu` and `apv` values are inputs
+into the key derivation process that is used to derive the content encryption key. However, regardless of algorithm used, the values are always part of the AEAD tag computation so will still be bound to the encrypted response.
 
 The following is a non-normative example of the payload of a JWT used in an Authorization Response that is encrypted and not signed:
 
@@ -2709,6 +2717,7 @@ The technology described in this specification was made available from contribut
 
    -24
 
+   * Mandate the use of apu/apv in the JWE header of encrypted responses
 
    -23
 


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1128,6 +1128,14 @@ If the JWT is only a JWE, the following processing rules MUST be followed:
 - `iss`, `exp` and `aud` MUST be omitted in the JWT Claims Set of the JWE, and the processing rules as per [@!JARM] Section 2.4 related to these claims do not apply.
 - The processing rules as per [@!JARM] Section 2.4 related to JWS processing MUST be ignored.
 
+Whenever the response is encrypted, the following additional processing directives apply to bind the Verifier's identifier and its respective transaction with the End-User into the JWE:
+
+- The `apu` (Agreement PartyUInfo) header parameter MUST be set to the value of the `nonce` from the Authorization Request.
+- The `apv` (Agreement PartyVInfo) header parameter MUST be set to the value of the `client_id` of the Verifier.
```
awoie, line 1134: Should we use only `apv` for this, i.e., client_id:nonce? Since it is the recipient info and nonce was provided by the recipient. This would allow us to put other things into `apu` if needed.

selfissued (changes requested):
Given that the nonce is part of the encrypted payload, it already contributes to the cryptographic output.  Therefore also using it as the "apu" value is redundant.

Likewise, the Client ID is part of the encrypted content, and so need not also be in "apv".

bc-pi:
> Given that the nonce is part of the encrypted payload, it already contributes to the cryptographic output. Therefore also using it as the "apu" value is redundant.
> 
> Likewise, the Client ID is part of the encrypted content, and so need not also be in "apv".

Arguably none of this is needed or useful but this is my attempt to carry out the will of the WG in a compromise that isn't a complete departure from the specification stack this all builds on and hopefully put an end to months of confusing and confused but very heated discussion on the topic.

awoie (approved):
Approving this change since it implements what was agreed in the DCP WG call. I want to note that this will not work with the Multi RP PR https://github.com/openid/OpenID4VP/pull/308 since the response might be encrypted to different client IDs.

Sakurann:
@selfissued 

> Given that the nonce is part of the encrypted payload, it already contributes to the cryptographic output. Therefore also using it as the "apu" value is redundant.
> 
> Likewise, the Client ID is part of the encrypted content, and so need not also be in "apv".

the purpose of using nonce/client_id as apu/apv values is to ensure interoperability so that the verifier can decrypt, and NOT audience restriction or session binding

bc-pi:
> Approving this change since it implements what was agreed in the DCP WG call. I want to note that this will not work with the Multi RP PR #308 since the response might be encrypted to different client IDs. After the Multi RP RP got merged, this text will need to be revised accordingly.

Noted. Though I'm not sure what one response encrypted to different client IDs would look like...

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1128,6 +1128,14 @@ If the JWT is only a JWE, the following processing rules MUST be followed:
 - `iss`, `exp` and `aud` MUST be omitted in the JWT Claims Set of the JWE, and the processing rules as per [@!JARM] Section 2.4 related to these claims do not apply.
 - The processing rules as per [@!JARM] Section 2.4 related to JWS processing MUST be ignored.
 
+Whenever the response is encrypted, the following additional processing directives apply to JWE. This enables binding response encryption to the Verifier's identifier and its respective transaction with the End-User:
+
+- The `apu` (Agreement PartyUInfo) header parameter MUST be set to the value of the `nonce` from the Authorization Request.
+- The `apv` (Agreement PartyVInfo) header parameter MUST be set to the value of the `client_id` of the Verifier.
```
paulbastian, line 1134: In ISO18013-7 the `apv` is the nonce instead of `apu`, does that make more or less sense?

tlodderstedt:
Discussion in the WG call on the 14th of Jan 2025: 

- With the latest change in #308, we can no longer assume the encryption key to be related to a certain Client Identifier. We might change this into the origin. @paulbastian raised the question how the RP would determine the correct origin if the request contains multiple `expected_origins`. 
- @Mike pointed out use of apv/apu would preclude algorithms that do not have those features.

paulbastian:
I think I agree with Mike, that defining new top-level parameters beside the JWE may be more future-proof than using `apu` and `apv`, as long as those things are not directly involved in the decryption itself

awoie:
> I think I agree with Mike, that defining new top-level parameters beside the JWE may be more future-proof than using `apu` and `apv`, as long as those things are not directly involved in the decryption itself

Tend to agree as well since all protected headers are part of the AAD which would include these newly defined headers. This would also meet the ISO requirement:
> Response encryption authentication must be bound to the origin, e.g. RP URL.

openid-4-verifiable-presentations-1_0.md
bc-pi: None of this makes much sense to me, to be honest. Per JWA - apu is information about the producer and apv is information about the recipient. ISO18013-7 has (or so I'm told) mdocGeneratedNonce for apu and nonce for apv which makes a certain amount of sense if you squint at it right. The wallet creates the mdocGeneratedNonce/apu so I guess it's information about the producer while the nonce/apv is generated by the verifier so I guess it's information about the recipient. If you squint at it a different way, it doesn't make any sense at all. I went with nonce as apu because it's about the the transaction or session the producer (Wallet) has with the Verifier and so seemed close enough to being about the producer and there are only two KDF contributing headers and client_id in apv made sense to me as information about the recipient because it's an identifier for the recipient.

That's a lot of words that probably don't make much sense either. 

The upside though, as I said in https://github.com/openid/OpenID4VP/pull/380#issuecomment-2583266050, is that I don't think any of this is needed or useful.

Sakurann:
so am I getting it right that the proposed solution is to add origin to the protected header? 
since the requirement is `Response encryption authentication must be bound to the origin, e.g. RP URL.` I assume there is no need to add any other information to the protected header? 

Does it also mean that we do not use apu and apv values? if we do not define them, there will be no way for both parties to be able to use the same values.

bc-pi:
> @selfissued
> 
> > Given that the nonce is part of the encrypted payload, it already contributes to the cryptographic output. Therefore also using it as the "apu" value is redundant.
> > Likewise, the Client ID is part of the encrypted content, and so need not also be in "apv".
> 
> the purpose of using nonce/client_id as apu/apv values is to ensure interoperability so that the verifier can decrypt, and NOT audience restriction or session binding

bc-pi:
> the purpose of using nonce/client_id as apu/apv values is to ensure interoperability so that the verifier can decrypt, and NOT audience restriction or session binding

that was not my understanding but perhaps my not understanding is part of the problem

bc-pi:
> Discussion in the WG call on the 14th of Jan 2025:
> 
>     * With the latest change in [Add Multi RP Credentials/Authentication capability #308](https://github.com/openid/OpenID4VP/pull/308), we can no longer assume the encryption key to be related to a certain Client Identifier. We might change this into the origin. @paulbastian raised the question how the RP would determine the correct origin if the request contains multiple `expected_origins`.

The rational for multiple `expected_origins` was questioned from it's introduction but it was explained that at IIW it was determined as needed because the backend infrastructure thing that signs these requests might not know the exact origin the whole page or whatever the script invoking the API is being served on. Which I can kinda see. But it'll have to know the set of possible origins in order to make the `expected_origins` list. I think an RP/Verifer would just need to ensure that the origin in the JWE header is one of those that it expects. 

> 
>     * @mike pointed out use of apv/apu would preclude algorithms that do not have those features.

I don't understand what that means exactly but what it seems to say on the face of things isn't correct.

bc-pi:
> I think I agree with Mike, that defining new top-level parameters beside the JWE may be more future-proof than using `apu` and `apv`, as long as those things are not directly involved in the decryption itself

I apologize for missing the call where this all was discussed and for my not being able to understand much of this but I don't understand how or why defining new top-level parameters beside the JWE would contribute anything to anything.

bc-pi:
> > I think I agree with Mike, that defining new top-level parameters beside the JWE may be more future-proof than using `apu` and `apv`, as long as those things are not directly involved in the decryption itself
> 
> Tend to agree as well since all protected headers are part of the AAD which would include these newly defined headers. This would also meet the ISO requirement:
> 
> > Response encryption authentication must be bound to the origin, e.g. RP URL.

"protected headers" are different than "top-level parameters beside the JWE" and maybe I'm being too literal but it's a meaningful distinction. Also `apu` and `apv` are protected headers. All headers are in the compact serialization (but please please don't take that as any kind of support for the JSON serialization). So yeah, the AAD would include these newly defined headers but it also includes the `apu` and `apv` headers (which also contribute to the KDF in ECDH). The text in the PR even tries to say this. So I am indeed sorry but I am pretty confused about what was discussed yesterday and what the concerns are and what to do with any of it.

bc-pi:
> so am I getting it right that the proposed solution is to add origin to the protected header? since the requirement is `Response encryption authentication must be bound to the origin, e.g. RP URL.` I assume there is no need to add any other information to the protected header?

Is this where the WG wants to take this? Asking 'cause I honestly don't know. 

> 
> Does it also mean that we do not use apu and apv values? if we do not define them, there will be no way for both parties to be able to use the same values.

We don't define them now, they don't need to be used and default behavior when they aren't there is specified.  So I'm not sure why it would be a problem. But I've very likely misunderstood something.

Sakurann:
Documented WG discussions say that[ `do a PR in OID4VP to say "use JWE with ECDH-ES"`.](https://github.com/openid/oid4vc-haip/issues/131#issuecomment-2539528136). we should either say that in this PR, or rediscuss.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1150,6 +1150,14 @@ If the JWT is only a JWE, the following processing rules MUST be followed:
...
+
+Note that for the ECDH JWE algorithms (from section 4.6 of [@!RFC7518]), the `apu` and `apv` values are inputs
+into the key derivation process that is used to derive the content encryption key. However, regardless of algorithm used, the values are always part of the AEAD tag computation so will still be bound to the encrypted response.
+
```
Sakurann, line 1160:
```suggestion
Note that for the ECDH JWE algorithms (from section 4.6 of [@!RFC7518]), the `apu` and `apv` values are inputs
into the key derivation process that is used to derive the content encryption key. Regardless of algorithm used, the values are always part of the AEAD tag computation so will still be bound to the encrypted response.

```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2767,6 +2775,7 @@ The technology described in this specification was made available from contribut
 
    -24
 
+   * Mandate the use of apu/apv in the JWE header of encrypted responses
```
Sakurann, line 2778:
```suggestion
   * add a note on the use of apu/apv in the JWE header of encrypted responses
```
also needs to be moved to -25

Sakurann (changes requested):
after the virtual ISO mtg, there has been an agreement not to define apu/apv values to be client_id/nonce, etc. 
option 1 is to close this PR.
option 2 is to refactor this PR to add a note on apv/apu.
I made suggestions for option 2 which could be slightly nicer, but do not have strong opinion

awoie:
> The "apu" and "apv" values
>    MUST be distinct, when used.

I think the important part of the sentence is "when used". Since we are not using them, it is up to more specific profiles to defined these values "when used".

awoie:
> RFC 7518 also says that:
> 
> ```
> Applications need to specify how the "apu" and "apv" Header
>    Parameters are used for that application.  The "apu" and "apv" values
>    MUST be distinct, when used.  Applications wishing to conform to
>    [[NIST.800-56A](https://datatracker.ietf.org/doc/html/rfc7518#ref-NIST.800-56A)] need to provide values that meet the requirements of
>    that document, e.g., by using values that identify the producer and
>    consumer.
> ```
> 
> So I think we need to normatively specify how the apu and apv header are used (which could be that they must be empty). I'm not sure what the exact implications are for the requirements regarding NIST 800-56A. Is anyone familiar with making the use of ECDH-ES compliant to NIST 800-56A?

It is a fair point. I believe the implications are that profiles who care about NIST 800-56A should set the `apu`/`apv` values accordingly. For example, HAIP could decide to address this requirement.

openid-4-verifiable-presentations-1_0.md
bc-pi: That'd be okay by me FWIW

Sakurann (approved):

Sakurann:
need to update the -25 in the title in the top (can also be another PR)

selfissued (changes requested):
The note, while true, doesn't add anything useful to the specification.

Thus, I'd prefer to see this closed without merging.  Thanks.

Sakurann:
@selfissued not everyone implementing OpenID4VP specs are JWE experts and we have received continuous feedback that these kind of notes are helpful, so please do not block unless you disagree with the content of the note.

selfissued (approved):

PR #381, opened 2025-01-10
Add version and request type to protocol identifier for DC API

### LINKED ISSUE #326: Adding request type to the protocol identifier
# Problem Statement

The top level properties of the request object that are passed into the Digital Credentials API (via the `data` property), are different for signed and unsigned requests. For example, a signed request only has `payload` and `signature` while an unsigned request has a completely different set of properties.

For implementers, the only way to know whether the request is signed or unsigned is to do some implicit matching based on the properties which are present at the top level of the `data` object (e.g. look for `payload` or `signature`). This doesn't scale well, can be error prone, and thus makes implementation more challenging.

# Proposal

The loose consensus from the 2024-11-13 Digital Credentials API call was to bring this issue forward for discussion in the DCP WG: https://github.com/WICG/digital-credentials/issues/185.

The proposal in the issue is to use different protocol strings when the top level structure is different. So for OID4VP, for signed vs unsigned requests, `openid4vp-signed` and `openid4vp-unsigned` respectively.

```javascript
const credential = await navigator.credentials.get({
  digital: {
    providers: [{
      protocol: "openid4vp-signed",
      data: {
        "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
        "signatures": [
          {
            "protected": "eyJhbGciOiJFUzI1NiJ9",
            "header": {
              "client_id": "987647789",
              "client_id_scheme": "x509_san_dns"
            },
            "signature": "PFwem0Ajp2Sag...T2z784h8TQqgTR9tXcif0jw"
          }
        ]
      }
    }
    ]
  }
});
```
## Alternative Proposal

An alternative would be to have an additional top level property for the DigitalCredentialsRequest object, such as `requestType`, which would be an enum of values appropriate to the protocol (and would be part of the registry).

```javascript
const credential = await navigator.credentials.get({
  digital: {
    providers: [{
      protocol: "openid4vp",
      requestType: "signed",
      data: {
        "payload": "eyAiaXNzIjogImh0dHBzOi8...NzY4Mzc4MzYiIF0gfQ",
        "signatures": [
          {
            "protected": "eyJhbGciOiJFUzI1NiJ9",
            "header": {
              "client_id": "987647789",
              "client_id_scheme": "x509_san_dns"
            },
            "signature": "PFwem0Ajp2Sag...T2z784h8TQqgTR9tXcif0jw"
          }
        ]
      }
    }
    ]
  }
});
```

### LINKED ISSUE #363: Adding a version to the protocol identifier, and making it a URN
The W3C Digital Credentials API specification will have a registry. In this registry, protocols will be normatively mapped to their respective specifications.

In the API itself, the protocol identifier is passed in the request as the `protocol` field, currently: `openid4vp`.

In #326, we've been discussing how two types of requests (signed vs unsigned) may want to have different protocol identifiers, as the top level contents of the request differ significantly. So having a type appended to the identifier can help (ex: openid4vp-signed).

One thing that is missing is the protocol version. When we inevitably end up with different versions of the specification, the protocol identifier needs to accurately represent the request payload. One can imagine sending two versions of the same request in the event the user only has a wallet that speaks an older version of the protocol.

To support both a version number (this issue) and request type (#326), I've proposed that the DC API use URNs for protocol identifiers: WICG/digital-credentials/issues/191.

Example for a signed request v1.0: `urn:openid:spec:openid4vp:v1.0:signed`.

awoie:
This PR adds version and request type to the protocol identifier for DC API.

Fixes #326, fixes #363

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1960,7 +1960,7 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:v<version>:<request-type>`. The `<version>` field is a numeric value, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
 
-The value `1` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
+The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
 
 The following exchange protocol values are defined by this specification:
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1960,7 +1960,7 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:oid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
 
-This specification defines `1.0` as the value for the `<version>` field. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), use `unsigned`, and signed requests, as defined in (#signed_request), use `signed`.
+The value `1.0` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
 
 The following exchange protocol values are defined by this specification:
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1964,8 +1964,8 @@ The value `1` MUST be used for the `<version>` field to indicate the request and
 
 The following exchange protocol values are defined by this specification:
 
-* Unsigned requests: `urn:openid:protocol:openid4vp:v1:unsigned`
-* Signed requests: `urn:openid:protocol:openid4vp:v1:signed`
+* Unsigned requests: `urn:openid:dcapi-protocol:openid4vp:v1:unsigned`
+* Signed requests: `urn:openid:dcapi-protocol:openid4vp:v1:signed`
 
 ## Request {#dc_api_request}
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1958,14 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:v<version>:<request-type>`. The `<version>` field is a numeric value, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
 
-The value `1.0` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
+The value `1` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
 
 The following exchange protocol values are defined by this specification:
 
-* Unsigned requests: `urn:openid:protocol:openid4vp:1.0:unsigned`
-* Signed requests: `urn:openid:protocol:openid4vp:1.0:signed`
+* Unsigned requests: `urn:openid:protocol:openid4vp:v1:unsigned`
+* Signed requests: `urn:openid:protocol:openid4vp:v1:signed`
 
 ## Request {#dc_api_request}
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1958,14 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:oid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
 
 The value `1.0` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
 
 The following exchange protocol values are defined by this specification:
 
-* Unsigned requests: `urn:openid:protocol:oid4vp:1.0:unsigned`
-* Signed requests: `urn:openid:protocol:oid4vp:1.0:signed`
+* Unsigned requests: `urn:openid:protocol:openid4vp:1.0:unsigned`
+* Signed requests: `urn:openid:protocol:openid4vp:1.0:signed`
 
 ## Request {#dc_api_request}
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1958,7 +1958,7 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
 
 The value `1.0` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:oid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+
+This specification defines `1.0` as the value for the `<version>` field. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), use `unsigned`, and signed requests, as defined in (#signed_request), use `signed`.
+
+The following exchange protocol values are defined by this specification:
+
+* Unsigned requests: `urn:openid:protocol:oid4vp:1.0:unsigned`
+* Signed requests: `urn:openid:protocol:oid4vp:1.0:signed`
 
 ## Request {#dc_api_request}
 


awoie:
Is there any issue with setting the value to `unsigned` although the request was signed? Does this open up similar attack vectors as `alg=none`?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:oid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
```
c2bo, line 1961:
```suggestion
To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
+
+The value `1.0` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
+
+The following exchange protocol values are defined by this specification:
+
+* Unsigned requests: `urn:openid:protocol:oid4vp:1.0:unsigned`
+* Signed requests: `urn:openid:protocol:oid4vp:1.0:signed`
```
c2bo, line 1968:
```suggestion
* Unsigned requests: `urn:openid:protocol:openid4vp:1.0:unsigned`
* Signed requests: `urn:openid:protocol:openid4vp:1.0:signed`
```

c2bo (commented):
Not entirely sure about the `protocol`part in the urn - do we need that? What about `urn:openid:openid4vp:1.0:signed`?
Also we should stick with openid4vp (and openid4vci) instead of oid4vp imho

openid-4-verifiable-presentations-1_0.md
hlozi, line 1968: What if the wallet considers a request unsigned because it can’t verify the signature? Also, could you explain the impact of this on the session transcript? Does the wallet include the client ID from the parameters when the protocol identifier indicates a signed request, even if the signature isn’t verified? In my opinion, this will have some negative security implications. I believe this PR is not compatible with the session transcript PR (which was already approved). I would be okay with reopening the session transcript PR and removing the client ID from the session transcript if that is what you want to do.

hlozi (changes requested):
in general the request signature has value only if the Wallet can verify it. if the wallet can't verify the signature (for example in case of not trusting the chain the RP used) the request is considered by the wallet to be unsigned.

openid-4-verifiable-presentations-1_0.md
awoie: I'd assume that for signed requests, the wallet will need to include the `client_id` in the `SessionTranscript` in all cases. I'd further assume that if the wallet cannot verify the signed request according to the client ID scheme that was used, it would decline the request.

Are you suggesting that the wallet can choose to skip the verification of the signed request according to the client ID scheme in some cases?

hlozi: Assuming multiple wallets with different credentials can satisfy a request, and the RP doesn’t know which combination wallet-credential the user chooses, if the RP signs with the trusted schema of Wallet 1 under a context of issuing credential 1, and the user selects Wallet2-credential2 that doesn’t trust/know the signature scheme used by the RP, Wallet 2 should be able to skip verification and consider the request unsigned.

awoie: In my opinion, supporting this might require additional considerations. In your scenario, the RP does not know whether the wallet verifies the request. The RP provides a signed request along with a corresponding Client ID and expects this Client ID to be used in the response (via `SessionTranscript`) after the request has been verified according to the Client ID scheme.

For instance, if the wallet switches from x509_san_dns to web origin, this change would need to be communicated in the response through the protocol. Otherwise, the RP would not know which Client ID to use in the `SessionTranscript`. Right?

I’m not sure if this falls within the scope of this PR, the `SessionTranscript` PR, or if it should be treated as a separate feature altogether. Allowing the wallet to downgrade the verification procedure could impact security and potentially introduce attack vectors. Wouldn’t that require further discussion and clarification?

hlozi: II believe the RP shouldn’t rely on knowing if the Wallet successfully performed the signature verification. This could be risky. I also agree with the session transcript, which raises the question of whether we should include client ID in the session transcript at all.

If you think this isn’t within the scope of this PR, I’m happy to move the discussion elsewhere. For me, something is signed if the signature can be verified, not because the protocol says it is.

openid-4-verifiable-presentations-1_0.md
awoie: If there are no requirements, I would use the shortest version possible, e.g., `urn:openid4vp:1.0:signed`, `urn:openid4vp:1.0:unsigned`. What do you think @c2bo @timcappalli ?

openid-4-verifiable-presentations-1_0.md
awoie: I think this PR is really just about providing a version and request type to the request, so it makes parsing the request and generating the expected response easier which is why I would prefer to have this discussion elsewhere.

openid-4-verifiable-presentations-1_0.md
timcappalli: Fine with dropping protocol but I believe keeping the org name is common practice. If length is the concern, maybe just use "oidf"? So it would be `urn:oidf:openid4vp:1.0:signed`.

@selfissued should weigh in here.

c2bo: I am happy with either one.

Adding the `oidf` seems to be a bit cleaner in terms of namespacing, but I don't think it's likely to encounter a naming clash here, so I guess `urn:openid4vp:1.0:signed` would be fine as well if we need/want to optimize for size?

timcappalli: so there is some OIDF precedent. CIBA defines a URN-based grant type that includes both the organization and the type (in this case `params`): `urn:openid:params:grant-type:ciba`. There's also `urn:openid:params:jwt:claim:auth_req_id` and `urn:openid:params:jwt:claim:rt_hash`. 

So if we wanted to be consistent with other OIDF specs, I think we'd want it to be `urn:openid:protocol:oid4vp:1.0:signed`. 

@selfissued @ve7jtb any comments on this?

awoie: Current proposal is:
* Unsigned requests: `urn:openid:protocol:openid4vp:1.0:unsigned`
* Signed requests: `urn:openid:protocol:openid4vp:1.0:signed`

awoie:
> e chain the RP used) th

Could you please elaborate this a bit further? I'm not sure if this is related to this PR.

Sakurann (approved):

Sakurann:
@hlozi I think from the perspective of processing rules in OpenID4VP, if the verifier signed the request, and the wallet cannot validate it, the wallet needs to abort the request. when the request is signed it's not wallet's choice to validate it or not, the wallet muct validate it

c2bo (approved):

paulbastian:
> Not entirely sure about the `protocol`part in the urn - do we need that? What about `urn:openid:openid4vp:1.0:signed`?
> Also we should stick with openid4vp (and openid4vci) instead of oid4vp imho

Would agree here, aren't those always "protocols" that do into DC API?

timcappalli:
> > Not entirely sure about the `protocol`part in the urn - do we need that? What about `urn:openid:openid4vp:1.0:signed`?
> > Also we should stick with openid4vp (and openid4vci) instead of oid4vp imho
> 
> Would agree here, aren't those always "protocols" that do into DC API?

While I agree it seems unnecessary at face, isn't there value in being consistent with other [OIDF URNs](https://github.com/openid/OpenID4VP/pull/381#discussion_r1917432408)?

timcappalli (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
 
-To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+
+The value `1.0` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
+
+The following exchange protocol values are defined by this specification:
+
+* Unsigned requests: `urn:openid:protocol:openid4vp:1.0:unsigned`
+* Signed requests: `urn:openid:protocol:openid4vp:1.0:signed`
```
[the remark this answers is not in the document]
bc-pi: and what of #308?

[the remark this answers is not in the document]
awoie: IMO, the version identifier does not have to match the OID4VP version. In this regards, we might have to think about if this PR should really set the version to 1.0 already or if we should start with 0.1 first.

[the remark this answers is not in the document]
timcappalli: The version in the protocol identifier will need to be mapped back to a specification in the registry, so the version needs to at a minimum be defined somewhere in the spec, if its not going to the be the spec version itself.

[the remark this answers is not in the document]
bc-pi: Or, and hear me out, having that level of specificity in the identifier is not a good idea.

Sakurann, line 1968:
```suggestion
To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `urn:openid:protocol:openid4vp`.
```

timcappalli: Disagree with this change. One of the two core requirements of the structured protocol identifier is version number.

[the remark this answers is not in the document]
Sakurann: it increasingly seems better long-term not to include both version and signed/unsigned in the protocol identifier:
- no version, because for 1.0, we know there will be backward compatible 1.1, 1.2, etc. and having to maintain those most likely will turn out to be an unnecessary maintenance overhead.
- no signed/unsigned because wallets already need to differentiate signed/unsigned requests and is not worth the complexity of asking the platforms to support both.

```suggestion
To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `urn:openid:protocol:openid4vp`.
```

Sakurann, line 1968: it increasingly seems better long-term not to include both version and signed/unsigned in the protocol identifier:
- no version, because for 1.0, we know there will be backward compatible 1.1, 1.2, etc. and having to maintain those most likely will turn out to be an unnecessary maintenance overhead.
- no signed/unsigned because wallets already need to differentiate signed/unsigned requests and is not worth the complexity of asking the platforms to support both.

```suggestion
To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `urn:openid:protocol:openid4vp`.
```

Sakurann (changes requested):
having heard more opinions, i think version and signed/unsigned should be dropped

timcappalli:
This has been discussed extensively in both groups and consensus was reached. Why the sudden change?

bc-pi (dismissed):
having a minor version is a nonstarter for me 

signed vs. unsigned (vs. signed many times) seems unnecessary but I understand that some/many want it

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
Sakurann: In this thread please comment which option you support ie "+1 to 1 and B" etc.

agreement that there is a need to differentiate major versions.
1. version is part of URN, i.e., `:1.0`
2. version is part of protocol part, i.e. `:openid4vp` and `:openid4vp2` or something
3. not a URN, but just `openid4vp` and `openid4vp2`

no agreement to differentiate signed/unsigned in the URN
A. signed is part of URN, i.e., `:signed`, `:unsigned`
B. no indication of signed/unsigned in the URN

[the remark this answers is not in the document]
timcappalli: I think the questions are:

- Should version number be in the protocol identifier (y/n)
- What specificity should the version number be? (major, minor / just major)
- Should the request type (e.g. `signed` or `unsigned`) be included in the protocol identifier (y/n)
- What format should the protocol identifier be? ( urn, unstructured string)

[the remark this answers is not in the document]
timcappalli: 1. yes
2. major
3. yes
4. urn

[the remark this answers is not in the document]
David-Chadwick: Answers to @timcappalli questions
1. Yes
2. Major
3. Yes
4. urn

David-Chadwick: Seems much better for a wallet to not even to have to start parsing a request if it knows it cannot support the options being presented, rather than to always have to parse requests it subsequently finds out it cannot support

dyiop: If not in the protocol string (so that platforms/wallet don't have to support multiple "protocols"), I would like to see the signed vs unsigned indication somewhere in the request

It feels like bad practice to have wallets determine the type of request by failing to parse the request and then trying the other option. For example, this would make it hard to debug (and print easy-to-understand log messages) for truly malformed requests, since that could mean either: the request is malformed, or it is of the other type...

dyiop: For the version: I agree with the comment from [awoie](https://github.com/awoie) [yesterday](https://github.com/openid/OpenID4VP/pull/381/files#r1924169709) essentially stating that the OpenID4VP spec version and the "API" version might be different concepts. 

If theres agreement with that, we would just need to find some versioning system that each OpenID4VP Spec version is in charge of clearly indicating (and is semantically different from the OpenID4VP major.minor scheme to avoid having folks correlate the two concepts).

I liked `v1`, and any non-backwards compatible change in the spec can be update the "API" version to `v2`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
```
peppelinux, line 1961:
```suggestion
To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:<version>:<request-type>`. The `<version>` field adheres to semantic versioning, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
awoie: 1. yes
2. just major is fine
3. not sure (not opinionated)
4. urn sounds reasonable to achieve the goal

awoie:
I updated the PR to remove the minor version from the `<version>` field. Please review it again. So far, I've seen more support for the unsigned/signed distinction, so I haven't made any changes to that part.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
-To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:v<version>:<request-type>`. The `<version>` field is a numeric value, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+
+The value `1` MUST be used for the `<version>` field to indicate the request and response conform to this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
```
jogu, line 1963: I think we just need to be careful here as we hope that v1.1 of OID4VP will use '1' for version too, so I suggest instead of 'confirm to this version' we use 'compatible with this version':

```suggestion
The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1958,7 +1958,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
...
 
-To use OpenID4VP over the DC API, the value of the exchange protocol used with the Digital Credentials API (DC API), is `openid4vp`.
+To use OpenID4VP with the Digital Credentials API (DC API), the exchange protocol value has the following format: `urn:openid:protocol:openid4vp:v<version>:<request-type>`. The `<version>` field is a numeric value, and `<request-type>` explicitly specifies the type of request. This approach eliminates the need for Wallets to perform implicit parameter matching to accurately identify the version and the expected request and response parameters.
+
+The value `1` MUST be used for the `<version>` field to indicate the request and response are compatible with this version of the specification. For `<request-type>`, unsigned requests, as defined in (#unsigned_request), MUST use `unsigned`, and signed requests, as defined in (#signed_request), MUST use `signed`.
+
+The following exchange protocol values are defined by this specification:
+
+* Unsigned requests: `urn:openid:protocol:openid4vp:v1:unsigned`
+* Signed requests: `urn:openid:protocol:openid4vp:v1:signed`
```
Sakurann, line 1968:
```suggestion
* Unsigned requests: `urn:openid:dcapi-protocol:openid4vp:v1:unsigned`
* Signed requests: `urn:openid:dcapi-protocol:openid4vp:v1:signed`
```

jogu (approved):

bc-pi (approved):
I now "don't hate" this enough to approve.

c2bo:
I don't have a strong opinion on the signed vs unsigned suffix and from an implementation perspective, having an identifier for signed seems to make ones life easier with parsing which is good. 

One concern I have would be #395. If we agree that this kind of flow is a good idea and want to support such a planned downgrade in the protocol, then that would also mean that we are switching from a signed request to a response that would belong to an unsigned flow.

This wouldn't change the data structure of the response, but having two different identifiers for signed and unsigned and then switching between them in a request/response sequence would be pretty weird imho.

bc-pi (approved):
I now "don't hate" this enough to approve.

dyiop:
Will there be any part of the request which can tell the Wallet which minor version this request is using?

This could be useful for the Wallet to track errors for different minor versions coming in from various verifiers?

Agreed that it should not need to be part of the protocol identifier!

timcappalli (approved):

awoie:
> Will there be any part of the request which can tell the Wallet which minor version this request is using?

The protocol identifier won't have any minor version. I believe this was the consensus that was achieved in this PR and in the DCP WG meeting.

Sakurann (approved):
> I now "don't hate" this enough to approve.

same

@hlozi ?

hlozi (approved):
I’m still not convinced about making the identifier format too complicated. In reality, it’ll mostly be used to uniquely identify a specification, so the browser will still need to look at the payload regardless. I guess this is a discussion for the DC API group, not this group so I don't want to block this.

PR #387, opened 2025-01-21
client id scheme web-origin must not be used outside dc api

### LINKED ISSUE #384: Add clause preventing `web-origin:` scheme from being used in regular requests
There should be something like 

> the Wallet MUST NOT accept the `web-origin:` client id scheme if the request is not sent via DC API" 

somewhere in the text. Right now, the only requirement is that this scheme MUST NOT be _used_ in the authorization request (not that it MUST NOT be _accepted_).

Sakurann:
resolves #384


bc-pi (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -574,7 +574,7 @@ Body
...
 
-* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). It MUST NOT be used in the Authorization Request.
+* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme, if the request is not sent using the Digital Credentials API defined in (#dc_api).
```
timcappalli, line 577: nit

```suggestion
* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme if the request is not sent via the Digital Credentials API defined in (#dc_api).
```

peppelinux (approved):

awoie (approved):

PR #390, opened 2025-01-21
remove json string

### LINKED ISSUE #295: vp_token encoding is not 100% clear
The encoding of the vp_token parameter is not 100% clear for response_mode fragment or direct_post.

From 7.1 on vp_token:
```
In case Presentation Exchange was used, it is a JSON String or JSON object that MUST contain a single Verifiable Presentation or an array of JSON Strings and JSON objects each of them containing a Verifiable Presentations.
```

So the value can be a JSON String, JSON object or JSON array. If it is a single value and a string, would the thing to encode in the URI be `"the_vp_token_value"` (including quotes, so parsable as JSON string, weird IMHO) or just be `the_vp_token_value` (without quotes, so not parsable as JSON string). I've not seen `"the_verifiable_presentation"` thing but taking the spec by word this should be how it is done.

My current approach, to be prepared for both situations, is to try parsing as JSON string first and if that fails, use the `vp_token` value as is as a string. The problem with that approach is, that any value  is then used in the raw string form, even a value that should be parsable as JSON string but contains an error. This will most likely lead to a less specific error when the vp_token is used later on. Instead a parse error should be reported.

For response_mode direct_post.jwt this is ok, because we are parsing JSON content and thus the plain string option is not valid there.

Sakurann:
resolves #295


openid-4-verifiable-presentations-1_0.md
```diff
@@ -979,7 +979,7 @@ When a VP Token is returned, the respective response includes the following para
...
+ * In case [@!DIF.PresentationExchange] was used, it is asString or JSON object that MUST contain a single Verifiable Presentation or an array of strings and JSON objects each of them containing a Verifiable Presentation. Each Verifiable Presentation MUST be represented as a string (that is a base64url-encoded value) or a JSON object depending on a format as defined in (#format_specific_parameters).  When a single Verifiable Presentation is returned, the array syntax MUST NOT be used.  If (#format_specific_parameters) defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the `vp_token` response parameter. Otherwise, this specification does not require any additional encoding when a Credential format is already represented as a JSON object or a string.
```
Sakurann, line 982:
```suggestion
 * In case [@!DIF.PresentationExchange] was used, it is a string or JSON object that MUST contain a single Verifiable Presentation or an array of strings and JSON objects each of them containing a Verifiable Presentation. Each Verifiable Presentation MUST be represented as a string (that is a base64url-encoded value) or a JSON object depending on a format as defined in (#format_specific_parameters).  When a single Verifiable Presentation is returned, the array syntax MUST NOT be used.  If (#format_specific_parameters) defines a rule for encoding the respective Credential format in the Credential Response, this rules MUST also be followed when encoding Credentials of this format in the `vp_token` response parameter. Otherwise, this specification does not require any additional encoding when a Credential format is already represented as a JSON object or a string.
```

bc-pi (approved):

c2bo (approved):

awoie (approved):

awoie:
I'm wondering why we are replacing JSON string with string but we would keep JSON object? Should we perhaps change that for all the other types as well?

Sakurann:
yeah, the sense in the room was that the use of JSON object was ok..

tplooker (approved):

PR #393, opened 2025-01-23
 DCQL: express desired credential issuers

### LINKED ISSUE #322: Enable RP to convey the desired credential issuers to the Wallet
The ISO specification is considering adding a parameter to identify the supported issuers for the requested document. We should add this as an option to DCQL as well.

The definition for the parameter is mdoc specific so we should probably add it as an mdoc specific parameter to the claims query. It could have the following definition:

issuer_identifiers: this is an array of issuer_identifier. If present, the issuer of the returned mdoc has to be one of issuer_identifier to satisfy the request from the RP.
The format used to identify an issuing authority is the base64url encoded keyIdentifier of the AuthorityKeyIdentifier that’s referenced in one of the certificates in the x509 chain element in the IssuerAuth header of the mdoc.

c2bo:
Closes #322

please re-review as discussed changes were applied:
- drop vical and x5t in this PR

Some next steps
- open the issues to potentially add vical and x5t later
- moving openid federation into a separate PR unless agreement on openid_fed is reached before thursday
- bikeshed the parameter name to the extent it does not impact the merging timelines too much


openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,127 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+#### OpenID Federation
+
+Type:
+: `"oidf"`
+
+Value:
+: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
+an entity in an OpenID Federation System. While this Entity Identifier could be any entity in
+that ecosystem, this entity would usually have the Entity Configuration of a Trust Anchor.
+The trust chain of a matching credential MUST contain the given Entity Identifier.
```
lj-raidiam, line 822:
```suggestion
The trust chain of a matching credential MUST contain the Entity Configuration issued by the Entity identified by the given Entity Identifier.
```

lj-raidiam: There is no direct use case for that but logically we should also consider including option to use intermediary ids. Any thoughts?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,127 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+TODO: Do we need to specify more like Entity Type for matching?
```
lj-raidiam, line 824: If OpenID Federation 4 Wallet Architectures is supported the issuer could be only https://openid.net/specs/openid-federation-wallet-1_0.html#section-6.2 
@peppelinux any thoughts?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,127 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+Each entry in `accepted_issuers` MUST be an object with the following properties:
+
+`type`:
+: REQUIRED. A string uniquely identifying the type of information about the issuer trust framework.
+Possible types are defined below.
+
+`values`:
+: REQUIRED. An array of strings, where each string (value) contains information specific to the
+used Accepted Issuer Query type that allows to identify an issuer or a trust ecosystem that an
```
lj-raidiam, line 733:
```suggestion
used Accepted Issuer Query type that allows to identify an issuer, trust framework, or a federation that an
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,127 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+```
+
+#### OpenID Federation
+
+Type:
+: `"oidf"`
+
+Value:
+: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
+an entity in an OpenID Federation System. While this Entity Identifier could be any entity in
```
lj-raidiam, line 820: We could change OpenID Federation System to a federation or an ecosystem or something along these lines.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,127 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+```
+
+#### ETSI Trusted List
+
+Type:
+: `"etsi_tl"`
+
+Value:
+: The identifier (can also be a URL) of a Trusted List as specified in ETSI TS 119 612.
+TODO: How to express restrictions within that TL? e.g., filtering by service type?
```
c2bo, line 803: My biggest question here is if the expectation is for a specific trusted list (of lists) to only include one type (e.g., all PID issuers), or if we will have bigger lists where we need to filter within that one list for different types.

I started reading ARF annex 2 for PID issuance and currently my understanding would be that for the PID we have a special list of PID Providers (so no filtering necessary). Is that correct and are we fine to start with the assumption that no further filtering is necessary here?

@babisRoutis are you be any chance able to provide a bit more information on this?

openid-4-verifiable-presentations-1_0.md
c2bo: Sounds reasonable with my limited knowledge of OIDF, @selfissued, @peppelinux ?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+of a matching Credential MUST contain at least one X.509 Certificate that matches one of the entires of the
+Trusted List or its cascading Trusted Lists.
+
+Below is a non-normative example of such an entry of type  `etsi_tl`:
+```json
+{
+  "type": "etsi_tl",
+  "values": ["https://lotl.example.com"]
+}
+```
```
c2bo, line 814: I might be missing something but it seems to me that how exactly the PID Issuer trusted list (or list of lists) will look like is currently not 100% clear- In that case I would propose to leave the ETSI TL as is (without more complex matching logic) and open an issue to revisit before going final

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+Below is a non-normative example of such an entry of type `x5t#S256`:
+```json
+{
+  "type": "x5t#S256",
+  "values": ["455943cf819425761d1f950263ebf54755d8d684c25535943976f488bc79d23b"]
+}
+```
+
+#### Authority Key Identifier
```
Sakurann, line 757: ISO feedback. less options the better. AKI alone is probably sufficient. AKI meets ISO requirement.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+{
+  "type": "etsi_tl",
+  "values": ["https://lotl.example.com"]
+}
+```
+
+#### OpenID Federation
+
+Type:
+: `"oidf"`
```
selfissued, line 819:
```suggestion
: `"openid_fed"`
```

At the risk of being a broken record, for branding reasons, we should always be spelling "OpenID" using all of its letters, and not shortening it to "OID", which means Object IDentifier in most identity contexts.  Thanks.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+Value:
+: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
+an entity in an OpenID Federation System. While this Entity Identifier could be any entity in
+that ecosystem, this entity would usually have the Entity Configuration of a Trust Anchor.
+The trust chain of a matching credential MUST contain the given Entity Identifier.
+
+Below is a non-normative example of such an entry of type `oidf`: 
+```json
+{
+  "type": "oidf",
```
selfissued, line 830:
```suggestion
  "type": "openid_fed",
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+Below is a non-normative example of such an entry of type `aki`:
+```json
+{
+  "type": "aki",
+  "values": ["s9tIpPmhxdiuNkHMEWNpYim8S8Y"]
+}
+```
+
+#### Verified Issuer Certificate Authority List (VICAL)
```
Sakurann, line 775: DCP WG discussion. leaning to keep it in OpenID4VP because value is not defined as a URL, but an identifier that does not have to be resovled

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+`values`:
+: REQUIRED. An array of strings, where each string (value) contains information specific to the
+used Accepted Issuer Query type that allows to identify an issuer, trust framework, or a federation that an
+issuer belongs to.
+
+Below are descriptions for the different Type Identifiers (string), the description on how to interpret
+and perform the matching logic for each provided value.
+
+#### X.509 Thumbprint
```
Sakurann, line 739: Reasoning to introduce this was for the CA that do not have AKIs. (How many of certs don't have AKIs?)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+of an X.509 certificate in the trust chain of the Credential.
+
+Below is a non-normative example of such an entry of type `x5t#S256`:
+```json
+{
+  "type": "x5t#S256",
+  "values": ["455943cf819425761d1f950263ebf54755d8d684c25535943976f488bc79d23b"]
+}
+```
+
```
awoie, line 756: I'd suggest removing X.509 thumbprint because it is less useful than `aki` and doesn't have any benefit over `aki`. It has actually a couple of downsides like having issues with re-issued certificates.
```suggestion
```

openid-4-verifiable-presentations-1_0.md
lj-raidiam: Good point. It would be nice to make it more universal.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+Type:
+: `"oidf"`
+
+Value:
+: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
+an entity in an OpenID Federation System. While this Entity Identifier could be any entity in
+that ecosystem, this entity would usually have the Entity Configuration of a Trust Anchor.
+The trust chain of a matching credential MUST contain the given Entity Identifier.
+
+Below is a non-normative example of such an entry of type `oidf`:
```
c2bo, line 827:
```suggestion
Below is a non-normative example of such an entry of type `openid_fed`: 
```

sloops77:
@c2bo what about permitting using DID's? Swiss EID & Ayra are planning to use OID4VC with DIDs and would benefit from this feature

c2bo:
> @c2bo what about permitting using DID's? Swiss EID & Ayra are planning to use OID4VC with DIDs and would benefit from this feature

I thought about that but my assumption was that DIDs would work pretty well with value matching.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: Then we should consider dropping VICAL.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: I agree. The current revision is a good starting point. Implementation experience will show what is needed on top.

tlodderstedt (approved):

openid-4-verifiable-presentations-1_0.md
c2bo: The 2 biggest reasons why I chose to include it in the initial proposal was 
- x5t is somewhat commonly used in the oauth world -> keep the choices consistent
- AuthoriyKeyIdentifier is an extension and I was not sure if every ecosystem that would want to use AKI.

That said, I am happy to remove it if everyone can use / is happy with AKI. Less choices is always a good thing. It would also be easy to re-introduce x5t later on.

sloops77:
>> @c2bo what about permitting using DID's? Swiss EID & Ayra are planning to use OID4VC with DIDs and would benefit from this feature

> I thought about that but my assumption was that DIDs would work pretty well with value matching.

I somehow ignore the first section that described using `iss` matching. I think the name should be looked at again to reduce confusion as I really think this feature is about `accepted_issuer_authorities`

c2bo:
> I somehow ignored the first section that described using `iss` matching. I think perhaps the name should be changed from `accepted_issuers` to `accepted_issuer_authorities` or `accepted_trust_anchors`.
> 
> Alternatively I wonder if you considered doing something wider such as `accepted_trust_frameworks`. This sort of name could support a wider scope, more than simply authority matching, it could require a particular trustmark in OpenId Fed.

Fair point on the naming. We currently have a bit of a mix with more direct matching like AKI and the broader methods like etsi trusted lists and openid federation. That is why I went with `accepted_issuers`, but happy to change if we can agree on a better name.

openid-4-verifiable-presentations-1_0.md
balfanz: With the disclaimer that I'm no expert in OIDF, I believe there is no such thing as "the trust chain" for a given credential. There could be multiple trust roots that a credential could chain up to (especially assuming that roots will cross-certify each other as trusted peers, e.g. in the EU). So I would recommend something like "A valid trust path, terminating at the given Entity Identifier, must be constructible from a matching credential".

c2bo: Agreed that trust chain doesn't seem to be the best wording here. I am a bit hesitant with the word "terminating" as it sounds like we want to allow intermediate entities and terminating could be interpreted as reaching a Trust Anchor?

Would this work for you @lj-raidiam , @balfanz 

```suggestion
A valid trust path, reaching the given Entity Identifier, must be constructible from a matching credential.
```

TimoGlastra (commented):
I think adding `did` as a supported issuer type here makes sense as well, since DIDs are also allowed as `client_id_scheme`, and ETSI trust list is also included (which I'm not sure belongs in the base spec if we follow the reasoning why DIDs was proposed to be removed)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+```
+
+#### ETSI Trusted List
+
+Type:
+: `"etsi_tl"`
+
+Value:
+: The identifier (can also be a URL) of a Trusted List as specified in ETSI TS 119 612 [@ETSI.TL].An ETSI
+Trusted List contains references to other Trusted Lists, creating a list of trustedf lists, or entries
```
TimoGlastra, line 803:
```suggestion
Trusted List contains references to other Trusted Lists, creating a list of trusted lists, or entries
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+for Trust Service Providers with corresponding service description and X.509 Certificates. The trust chain
+of a matching Credential MUST contain at least one X.509 Certificate that matches one of the entires of the
+Trusted List or its cascading Trusted Lists.
+
+Below is a non-normative example of such an entry of type  `etsi_tl`:
```
TimoGlastra, line 808:
```suggestion
Below is a non-normative example of such an entry of type `etsi_tl`:
```

openid-4-verifiable-presentations-1_0.md
TimoGlastra: I'm curious how the processing of this would work. Knowing if you can build a chain including the requested entity id often involves a (complicated) resolving process. As @balfanz pointed out, multiple paths could exist, so you'd either have to:
- store (a subset of) the possible trust chains with the credential, so you can perform this offline 
- resolve the trust chain dynamically to match (this doesn't seem like a good approach, credential matching should be performed locally IMO).

openid-4-verifiable-presentations-1_0.md
```diff
@@ -700,6 +710,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+Note that explicit issuer matching can also be achieved using claim value matching for certain
+Credential formats (e.g., value matching the `iss` claim in an SD-JWT). The methods provided in
+accepted_issuers are more tailored towards matching trust frameworks and those that require more
+complex matching logic (e.g., certificate thumbprints).
+
+Each entry in `accepted_issuers` MUST be an object with the following properties:
+
+`type`:
+: REQUIRED. A string uniquely identifying the type of information about the issuer trust framework.
+Possible types are defined below.
```
TimoGlastra, line 729: "Possible types" suggests this is not extendable. Should this say "defined by this specification" as is done with client_id_scheme?

```suggestion
: REQUIRED. A string uniquely identifying the type of information about the issuer trust framework.
Types defined by this specification are defined below.
```

c2bo:
> I think adding `did` as a supported issuer type here makes sense as well, since DIDs are also allowed as `client_id_scheme`, and ETSI trust list is also included (which I'm not sure belongs in the base spec if we follow the reasoning why DIDs was proposed to be removed)

Do you think adding a section here would be beneficial over value matching on the issuer (e.g., iss = did:xyz:abcdef)?
I tried to keep those kind of simpler/more direct methods that work well with value matching (dids, web-based key resolution, etc.) out of this PR and only cater for trust frameworks and more complex matching.

TimoGlastra:
> Do you think adding a section here would be beneficial over value matching on the issuer (e.g., iss = did:xyz:abcdef)?
> I tried to keep those kind of simpler/more direct methods that work well with value matching (dids, web-based key resolution, etc.) out of this PR and only cater for trust frameworks and more complex matching.

Yeah I guess that would work fine. 

Thought of a few things and only anoncreds seems like it could be an issue:
- mDOC has no issuer field to query, but also doesn't support DIDs
- W3C / SD-JWT support `iss` / `issuer` value querying
- AnonCreds has a custom wrapping defined to make additional metadata values available in the query (such as `schema_id`), though no `issuer_id` value. This might need to be updated then to also include the `issuer_id`.

c2bo:
Questions for the working group:
- Remove/Keep x5t (X.509 thumbprint): Main Question is if we can assume that every X.509 based PKI uses/requires the Authority Key Identifier element
- Remove/Keep VICAL (trust framework for ISO mdoc/mdl): Feedback from the ISO working group seems to be that matching based on AKI would be sufficient.
- OpenID Federation: Should we simplify things and initially only care about express matching of Trust Anchors? We can still extend this later on if necessary

I would also propose to create a small PR in HAIP mandating only support for AKI, which would be in line with the key resolution parts of HAIP etc.

martijnharing:
I'm against including any option that uses a URL as part of the mechanism to determine which issuers are supported when the DC API is used. An explicit design consideration for the use of the DC API is that there is no necessity for any of the components (browser, os, wallet etc) to make an external call (at the very least before user consent) and we made significant efforts in the design to ensure that this is the case.
Having a mechanism that goes against those principles (as it is a privacy leak) should not be an option unless we have a very good reason for it to exist. 
Since there are mechanisms specified that do not require any call-outs, I don't see why there is such a necessity and think it's harmful if such an option exists in the specification.

c2bo:
> I'm against including any option that uses a URL as part of the mechanism to determine which issuers are supported when the DC API is used. An explicit design consideration for the use of the DC API is that there is no necessity for any of the components (browser, os, wallet etc) to make an external call (at the very least before user consent) and we made significant efforts in the design to ensure that this is the case. Having a mechanism that goes against those principles (as it is a privacy leak) should not be an option unless we have a very good reason for it to exist. Since there are mechanisms specified that do not require any call-outs, I don't see why there is such a necessity and think it's harmful if such an option exists in the specification.

I understand the concern, but the current wording only requires an ID (which could also be a URL - I can remove that remark). We have existing systems like the EU Trusted Lists that are usually referenced as the URL of the root list (https://ec.europa.eu/tools/lotl/eu-lotl.xml). All systems I've seen do not actively poll/resolve that URL, but in theory it is still provided via those signed XML files with their root at that URL and the URL seems to be a decent identifier.

openid-4-verifiable-presentations-1_0.md
TimoGlastra: Reaching would still suggest an end, maybe including works?

```suggestion
A valid trust path, including the given Entity Identifier, must be constructible from a matching credential.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -693,6 +693,16 @@ this object are defined per Credential Format. Examples of those are in (#sd_jwt
...
+`accepted_issuers`:
+: OPTIONAL. A non-empty array of objects as defined in (#dcql_accepted_issuers) that
+specifies expected issuers or trust frameworks that an issuer belongs to, that the
+Verifier will accept. Every Credential returned by the Wallet SHOULD match at least
+one of the conditions present in the corresponding `accepted_issuers` array if present.
+
+Note that Relying Parties must verify that the issuer of a received presentation is
+trusted on their own and this feature mainly aims to help data minimization by not
+revealing information that would likely be rejected.
+
```
peppelinux, line 705:
```suggestion
`accepted_issuers`:
: OPTIONAL. A non-empty array of objects as defined in (#dcql_accepted_issuers) that
specifies expected Credential Issuers that the
Verifier accepts. Every Credential returned by the Wallet SHOULD match at least
one of the conditions present in the corresponding `accepted_issuers` array if present.
Relying Parties using this parameter MUST ensure that the Credential Issuers specified are both active and trusted at the time of the query.
```

I would not mix trust frameworks identifiers with credential identifiers in the same parameter

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,128 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+: REQUIRED. A string uniquely identifying the type of information about the issuer trust framework.
+Types defined by this specification are listed below.
+
+`values`:
+: REQUIRED. An array of strings, where each string (value) contains information specific to the
+used Accepted Issuer Query type that allows to identify an issuer, trust framework, or a federation that an
+issuer belongs to.
+
+Below are descriptions for the different Type Identifiers (string), the description on how to interpret
+and perform the matching logic for each provided value.
```
peppelinux, line 742:
```suggestion
An Accepted Issuer Query is an object representing information that identify one or more Credential Issuers. 
A Credential matches an Accepted Issuer Query if it aligns at least its Credential Issuer and the Credential Issuer configures the Credential as supported (eg: `iss` and `vct` for SD-JWT). Different Credential Types may use different matching approaches.

Each `accepted_issuers` object MUST have the following members:

`credential_issuer`:
: REQUIRED. A string uniquely identifying the Credential Issuer.

`trust_evaluations_supported`:
: REQUIRED. An array of strings, uniquely identifying the type of information about the trust evaluation mechanism the Credential Issuer MUST support.

`trust_frameworks_supported`:
: OPTIONAL. An array of strings, where each string (value) contains information specific to the
supported trust framework the Credential Issuer MUST belong to, to be considered a valid Credential Issuer.

Below are descriptions for the different Type Identifiers (string), the description on how to interpret
and perform the matching logic for each provided value.
```

peppelinux (changes requested):
Before continuing the revision I ask to further clarify the scope of this parameter and this PR, since it seems about the accepted issuers but contains information that mixes trust evaluation mechanisms and trust frameworks (that are separate thing, even if they should be complementary. Consider that a trust evaluation mechanism can live without a properly defined trust framework).

My comments aims to facilitate this change, even if in an non exhaustive way.

selfissued:
I support @lj-raidiam's and @peppelinux's change suggestions.

openid-4-verifiable-presentations-1_0.md
c2bo: I don't think your proposal would work with the current scope of the PR:
- Part of the problem initially discussed  in the issue was that not every issuer is easily identified by a string (see the Authority Key Identifier discussion)
- The discussion in the issue basically spanned 2 problems:
  - Direct issuer matching -> This will be done by value matching unless impossible (AKI)
  - Expressing requirements about the trust chain/path -> Instead of directly identifying the issuer, expressing some expected properties/entities in the trust chain

This PR mainly covers the second part - describing some requirements that the trust chain/path of a credential has to fulfil. I think some of the confusion stems from the rather unfitting name "accepted_issuers"?

martijnharing:
> > I'm against including any option that uses a URL as part of the mechanism to determine which issuers are supported when the DC API is used. An explicit design consideration for the use of the DC API is that there is no necessity for any of the components (browser, os, wallet etc) to make an external call (at the very least before user consent) and we made significant efforts in the design to ensure that this is the case. Having a mechanism that goes against those principles (as it is a privacy leak) should not be an option unless we have a very good reason for it to exist. Since there are mechanisms specified that do not require any call-outs, I don't see why there is such a necessity and think it's harmful if such an option exists in the specification.
> 
> I understand the concern, but the current wording only requires an ID (which could also be a URL - I can remove that remark). We have existing systems like the EU Trusted Lists that are usually referenced as the URL of the root list (https://ec.europa.eu/tools/lotl/eu-lotl.xml). All systems I've seen do not actively poll/resolve that URL, but in theory it is still provided via those signed XML files with their root at that URL and the URL seems to be a decent identifier.

If the URL is only used as an identifier and it's prohibited to actually use it as a URL I agree there aren't any direct privacy concerns. But from a design perspective it's a bit risky to rely on the fact that implementers have to realize that they are prohibited from using the URL as URL, especially if they don't have the contents of what the URL is referring to but still they are prohibited from using it.
Other issues with URLs is that it can be a bit annoying for versioning, and doesn't lend itself cleanly to deviate slightly from the list identified by the URL.
Ultimately, the RP is aware of exactly which issuers can be used when it is making the request (in the example of using the list identified by the URL, the RP will be in possession of that list). So why not just let the RP provide all the elements on that list directly.

c2bo:
My take on WG discussion:
- remove x5t, VICAL for the time being - can be reintroduced later
- Need to find a better name than `accepted_issuers`
- What do we do with the OpenID Federation part? Should we move that into a separate PR and discuss or can we find a consensus that works for the time being?

Sakurann:
WG discussion:
- drop vical and x5t in this PR and open the issues to potentially add them later
- moving openid federation into a separate PR unless agreement on openid_fed is reached before thursday

openid-4-verifiable-presentations-1_0.md
lj-raidiam: > I'm curious how the processing of this would work. Knowing if you can build a chain including the requested entity id often involves a (complicated) resolving process. As @balfanz pointed out, multiple paths could exist, so you'd either have to:
> 
> * store (a subset of) the possible trust chains with the credential, so you can perform this offline
> * resolve the trust chain dynamically to match (this doesn't seem like a good approach, credential matching should be performed locally IMO).

Yes, wallets will have to store it for all credentials to do the filtering. Perhaps in more complex case (like two paths) this filtering will not be ideal. Verifiers will have a harder work to do as they will have to resolve it dynamically in most cases.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+```
+
+#### OpenID Federation
+
+Type:
+: `"openid_fed"`
+
+Value:
+: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
+an entity in an OpenID Federation System. While this Entity Identifier could be any entity in
```
c2bo, line 789:
```suggestion
: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
an entity in a federation. While this Entity Identifier could be any entity in
```

c2bo:
I switched from `accepted_issuers` to `trusted_authorities` - that was the best option I came up with. Felt quite a bit better to me - what do people think?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
 
 Note that multiple Credential Queries in a request MAY request a presentation of the same Credential.
 
+### Trusted Authorities Query {#dcql_trusted_authorities}
+
+A Trusted Authorities Query is an object representing information that helps to identify an issuer
+or the trust framework that an issuer is belonging to. A Credential is identified as a match
```
deshmukhrajvardhan, line 721: nit:belongs to

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+to a Trusted Authorities Query if it to matches with one of the provided values in one of the provided
```
deshmukhrajvardhan, line 722: nit: if it matches

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+types. How exactly the matching works is defined for the different types below
+
+Note that explicit issuer matching can also be achieved using claim value matching for certain
+Credential formats (e.g., value matching the `iss` claim in an SD-JWT). The methods provided in
+`trusted_authorities` are tailored towards matching trust frameworks and more direct issuer matching
+variants that require more complex matching logic (e.g., certificate thumbprints).
```
TimoGlastra, line 728: Certificate thumbprints are not part of this PR anymore. Would aki suffice as an example here? Or wouldn't that be called direct issuer matching?

openid-4-verifiable-presentations-1_0.md
c2bo: thanks for catching that - fixed with last commit

openid-4-verifiable-presentations-1_0.md
c2bo: Changed the example to AKI

openid-4-verifiable-presentations-1_0.md
```diff
@@ -693,6 +693,16 @@ this object are defined per Credential Format. Examples of those are in (#sd_jwt
 no specific constraints are placed on the metadata or validity of the requested
 Credential.
 
+`trusted_authorities`:
+: OPTIONAL. A non-empty array of objects as defined in (#dcql_trusted_authorities) that
+specifies expected issuers or trust frameworks that an issuer belongs to, that the
```
sloops77, line 698:
```suggestion
specifies expected authorities or trust frameworks that certify issuers, that the
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
 
 Note that multiple Credential Queries in a request MAY request a presentation of the same Credential.
 
+### Trusted Authorities Query {#dcql_trusted_authorities}
+
+A Trusted Authorities Query is an object representing information that helps to identify an issuer
+or the trust framework that an issuer belongs to. A Credential is identified as a match
```
sloops77, line 721:
```suggestion
A Trusted Authorities Query is an object representing information that helps to identify an authority
or the trust framework that certifies issuers. A Credential is identified as a match
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+to a Trusted Authorities Query if it matches with one of the provided values in one of the provided
+types. How exactly the matching works is defined for the different types below
+
+Note that explicit issuer matching can also be achieved using claim value matching for certain
```
sloops77, line 725:
```suggestion
Note, as opposed to authority or trust framework matching, explicit issuer matching should use claim value matching for certain
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+Credential formats (e.g., value matching the `iss` claim in an SD-JWT). The methods provided in
+`trusted_authorities` are tailored towards matching trust frameworks and more direct issuer matching
+variants that require more complex matching logic (e.g., Authority Key Identifier).
```
sloops77, line 728:
```suggestion
Note direct issuer matching should use claim value matching for certain Credential formats (e.g., value matching the `iss` claim in an SD-JWT).  The methods provided in `trusted_authorities` are for more indirect methods using trust frameworks or variants that require authority matching.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+A Trusted Authorities Query is an object representing information that helps to identify an authority
+or the trust framework that certifies issuers. A Credential is identified as a match
+to a Trusted Authorities Query if it matches with one of the provided values in one of the provided
+types. How exactly the matching works is defined for the different types below
+
+Note that direct issuer matching should not use `trusted_authorities`. Direct issuer matching
+should use claim value matching if possible (e.g., value matching the `iss` claim in an SD-JWT).
+The methods provided in `trusted_authorities` are for more indirect matching logic required by trust
+frameworks or X.509 PKI.
```
martijnharing, line 728: I don't understand this recommendation. 
For example, issuer matching based on the AKI of a certificate owned by a particular issuer can be very direct. In that example I think using AKI is the more reliable, more privacy and security preserving than a mechanism based on claim matching.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+
+#### Authority Key Identifier
+
+Type:
+: `"aki"`
+
+Value:
+: The base64url encoded Authority Key Identifier element as defined in Section 4.2.1.1 of [@!RFC5280].
+The raw byte representation of this element MUST match with the AuthorityKeyIdentifier element of an X.509
+certificate in the trust chain of a matching Credential.
```
martijnharing, line 752: Can we be more specific than 'trust chain'? For example 'the AKI of one of the certificates carried in the chain carried present in the header parameters of the credential. Note that chain can consist of a single certificate.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+}
+```
+
+#### ETSI Trusted List
+
+Type:
+: `"etsi_tl"`
+
+Value:
+: The identifier (can also be a URL) of a Trusted List as specified in ETSI TS 119 612 [@ETSI.TL].An ETSI
```
martijnharing, line 768: Since a URL is mentioned, can we add language that prohibits using the URL from retrieving the trust chain because of the privacy implications of such a fetch?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -705,6 +715,89 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+: The `Entity Identifier` as defined in Section 1 of [@!OpenID.Federation] that is bound to
+an entity in a federation. While this Entity Identifier could be any entity in
+that ecosystem, this entity would usually have the Entity Configuration of a Trust Anchor.
+A valid trust path, including the given Entity Identifier, must be constructible from a matching credential.
+
+Below is a non-normative example of such an entry of type `openid_fed`: 
+```json
+{
+  "type": "openid_fed",
+  "values": ["https://trustanchor.example.com"]
```
martijnharing, line 797: Since a URL is mentioned, can we add language that prohibits using the URL from retrieving the trust chain because of the privacy implications of such a fetch?

martijnharing (changes requested):

openid-4-verifiable-presentations-1_0.md
c2bo: This recommendation is not intended to say that some mechanisms are better or worse, but explain that if ecosystems want to do direct issuer matching based on claims, then `trusted_authorities` is not the right place and value-based matching should be used instead (e.g, matching `iss` of an SD-JWT to the value of an URL or a DID).

selfissued (approved):
I'm good with the current `openid_fed` language in the PR.

openid-4-verifiable-presentations-1_0.md
c2bo: Would this work for you @martijnharing ?
```suggestion
certificate in the certificate chain present in the credential (e.g., in the header of an mdoc or SD-JWT).
Note that the chain can consist of a single certificate.
```

openid-4-verifiable-presentations-1_0.md
c2bo: I can remove URL here if that helps? 
```suggestion
: The identifier of a Trusted List as specified in ETSI TS 119 612 [@ETSI.TL].An ETSI
```

openid-4-verifiable-presentations-1_0.md
c2bo: I understand your concern, but I don't think that would make sense in the context of openid federation?

c2bo: I don't think OpenID4VP would be the right place to do so - profiles like HAIP would be the right place to make recommendations in that regard imho

sloops77 (approved):
lgtm

openid-4-verifiable-presentations-1_0.md
c2bo: Hmm maybe it would make sense to add a remark here or in the privacy considerations section that the different mechanisms can have different privacy properties and especially in the case of online lookup, might leak some information?

Building on top of that, HAIP could make choices. Would that work for you?

openid-4-verifiable-presentations-1_0.md
selfissued: On the 20-Feb-25 working group call, there was support for discussing the privacy implications of online trust validation processes in Privacy Considerations text, and not blocking this PR on this.  I believe that @c2bo is going to create the PR.

c2bo:
As discussed on last WG call: added a small note that references a new subsection in the privacy considerations that describes possible privacy implications depending on used mechanism (especially online lookup).

@martijnharing does this work for you?

martijnharing:
> As discussed on last WG call: added a small note that references a new subsection in the privacy considerations that describes possible privacy implications depending on used mechanism (especially online lookup).
> 
> @martijnharing does this work for you?

The text added mostly talks about benefits of different trust mechanisms in general, which is related to the issue we are discussing but not necessarily fully applicable in the context of this PR. The goal of the issuer selection mechanism is only to tell the wallet which credential issuers it accepts, that's a signal coming from the RP. There isn't really any security impact related to it, it's just there for privacy and ux reasons.

So while there may be trust systems that require online checks to be done in certain scenarios, for the specific scenario of telling the wallet which RPs it will accept, the RP can tell the wallet which credential issuers it trusts. Even if that lists lives online, the RP has to be in possession in it and it could therefore provide the (simplified version) of that list to the wallet. 

Looking what this means for the proposed text, it says that:

The proposed text says "Generally speaking, a distinction can be made between self-contained mechanisms, where all information necessary to validate if a credential matches the request is already present in the Wallet and Verifier, and those mechanisms that require some form of online resolution."

My comment on this text would be that there is a difference for privacy and security purposes between the RP needing to call out for certain checks versus the wallet needing to call out for those checks (especially before user authentication). Ideally when possible we minimize the needs for the wallet to call out externally, more so than the need for the RP to do so.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1676,6 +1774,14 @@ Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount
 
 In the event that another component is invoked instead of the Wallet, the End-User MUST be informed and give consent before the invoked component returns the `wallet_unavailable` Authorization Error Response to the Verifier.
 
+## Privacy implications of mechanisms to establish trust in issuers {#privacy_trusted_authorities}
+
+This specification introduces an extension point that allows for a Verifer to express expected issuers or trust frameworks that cerify issuers.
```
selfissued, line 1779:
```suggestion
This specification introduces an extension point that allows for a Verifier to express expected issuers or trust frameworks that certify issuers.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1676,6 +1774,14 @@ Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount
...
+It is important to understand the implications that different mechanism to establish trust in issuers have on the privacy of the overall system.
+Generally speaking, a distinction can be made between self-contained mechanisms, where all information necessary to validate if a credential matches the request is already present in the Wallet and Verifier, and those mechanisms that require some form of online resolution.
+Mechanisms that require online resolution can leak information that could be used to profile the usage of credentials and the overall ecosystem.
+Ecosystems that plan to leverage the trusted authorities mechanisms SHOULD make sure that the privacy properties of the mechanisms they choose to support, matches with the desired privcacy properties of the overall ecosystem.
```
selfissued, line 1783:
```suggestion
Ecosystems that plan to leverage the trusted authorities mechanisms SHOULD make sure that the privacy properties of the mechanisms they choose to support matches with the desired privacy properties of the overall ecosystem.
```

TimoGlastra:
While prototyping this, one consideration we found of keeping some credential issuer restrictions as claim queries and some using this new syntax is that it can become hard to combine the two. It won't be possible to request a credential that is either issued by this did, or by this x509 certificate chain (aki). It will be possible however to request a credential issued by an openid federation OR an x509 certificate chain (aki). This puts a disadvantage on DIDs and I'd prefer it to be queryable using the customized syntax as well.

Sakurann:
@TimoGlastra 
>  It won't be possible to request a credential that is either issued by this did, or by this x509 certificate chain (aki). It will be possible however to request a credential issued by an openid federation OR an x509 certificate chain (aki). 

I think that issue is easy to solve by simply defining trusted_authorities of type `did` - either in OpenID4VP directly or its profile. Please feel free to open an issue on this - it would be v1.1 - and open a PR

openid-4-verifiable-presentations-1_0.md
```diff
@@ -698,6 +698,16 @@ this object are defined per Credential Format. Examples of those are in (#sd_jwt
 no specific constraints are placed on the metadata or validity of the requested
 Credential.
 
+`trusted_authorities`:
+: OPTIONAL. A non-empty array of objects as defined in (#dcql_trusted_authorities) that
```
peppelinux, line 702: In allegato this pagamento agraph, issuer -> Issuer since It Is intended to be credential issuers

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1691,6 +1789,17 @@ Requests from the Wallet to the Verifier SHOULD be sent with the minimal amount
 
 In the event that another component is invoked instead of the Wallet, the End-User MUST be informed and give consent before the invoked component returns the `wallet_unavailable` Authorization Error Response to the Verifier.
 
+## Privacy implications of mechanisms to establish trust in issuers {#privacy_trusted_authorities}
+
+This specification introduces an extension point that allows for a Verifier to express expected issuers or trust frameworks that certify issuers.
```
peppelinux, line 1794: issuers -> Issuers

peppelinux (approved):

openid-4-verifiable-presentations-1_0.md
c2bo: Fixed in latest commit

openid-4-verifiable-presentations-1_0.md
c2bo: Fixed in latest commit

openid-4-verifiable-presentations-1_0.md
martijnharing: Updated text looks good to me.

martijnharing (approved):

lj-raidiam:
> * for ETSI trust list, third party is hosting the URL, so issuers and RPs would not get a signal. 

In most cases it will also be true for the federation.

@martijnharing 
 
> Next step: Christian to add one more paragraph in privacy considerations about "the wallet should fetch the URL only when the wallet should have a way to verify that the url in the request is legitimate (not just blindly trust the value given by the RP). when trusted third party is hosting that URL, issuers and RPs would not get a signal, so the privacy concern is mitigated.

Makes sense.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -710,6 +720,93 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+#### Authority Key Identifier
+
+Type:
+: `"aki"`
+
+Value:
+: Contains the KeyIdentifier of the AuthorityKeyIdentifier as defined in Section 4.2.1.1 of [@!RFC5280],
+encoded as base64url. The raw byte representation of this element MUST match with the AuthorityKeyIdentifier
+element of an X.509 certificate in the certificate chain present in the credential (e.g., in the header of
+an mdoc or SD-JWT). Note that the chain can consist of a single certificate.
```
awoie, line 761:
```suggestion
: Contains the KeyIdentifier of the AuthorityKeyIdentifier as defined in Section 4.2.1.1 of [@!RFC5280],
encoded as base64url. The raw byte representation of this element MUST match with the AuthorityKeyIdentifier
element of an X.509 certificate in the certificate chain of the credential signing certificate. Note that the chain can consist of a single certificate, and the credential can include the entire X.509 chain or parts of it.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -710,6 +720,93 @@ The rules for selecting claims to send are defined in (#selecting_claims).
...
+}
+```
+
+#### ETSI Trusted List
+
+Type:
+: `"etsi_tl"`
+
+Value:
+: The identifier of a Trusted List as specified in ETSI TS 119 612 [@ETSI.TL].An ETSI
```
awoie, line 777:
```suggestion
: The identifier of a Trusted List as specified in ETSI TS 119 612 [@ETSI.TL]. An ETSI
```

openid-4-verifiable-presentations-1_0.md
awoie: I mean the whole chain. For mdocs you have x5chain = doc signer, and the IACA is transmitted out of band. The AKI could also match the IACA, not only the IACA if this makes sense.

Sakurann (approved):

Issue #397, opened 2025-01-28
same credential fulfilling multiple credential queries


Sakurann:
(might be a duplicate)
high-level requirement for a special (?) case: define the wallet's behavior/processing rules when the same credential can fulfill multiple credential queries in the same RP request.

example 1 - RP asks for mdoc for age_over_18 (query_id `abc`); and mdoc for address (query_id `xyz`):

option A:
vp_token: {
`abc`: mdoc with age_over_18
`xyz`: mdoc with address
}

Option B:
vp_token: {
`abc`: mdoc with both age_over_18 and address
`xyz`: mdoc with both age_over_18 and address (same as above)
}

- concern with option A is that it can trigger multiple user authentication? and for ZKP systems might mean that ZKP has to be computed twice? 

- does this result in the limitation of returning one mdoc per deviceResponse?

charsleysa:
Option A would be the most straightforward option to implement and would be deterministic with regards to which claims would be present in the presented credential.

If two credential queries were used instead of one then there may be a processing requirement where they need a presented credential with only the specified claims so with option B providing more claims than asked for could violate some PII policy.

Sakurann:
WG discussion:
1. don't care, don't do anything (option A in the example above)
2. duplicate the response (option B in the example above)
2.5 map in a map.
4. change the response structure (comma separate key in the VP Token map to allow multiple query_ids to match with multiple credentials?) 

- Concerns that options 3 is too complex

c2bo:
Example for the 2 map options:
```json
{
  "response": {
    "1": "ey...",
    "2": "<some other encoded credential>"
  } 
  "dcql": {
    "abc": "1",
    "xyz": "1",
    "uvw": "2"
  }
}
```
- one for id to credential
- one for DCQL Credential Query to id

That seems like the cleanest option to me, but the whole thing becomes quite a bit more complex which we should probably try to avoid if possible.

awoie:
> Example for the 2 map options:
> 
> {
>   "response": {
>     "1": "ey...",
>     "2": "<some other encoded credential>"
>   } 
>   "dcql": {
>     "abc": "1",
>     "xyz": "1",
>     "uvw": "2"
>   }
> }
> * one for id to credential
> * one for DCQL Credential Query to id
> 
> That seems like the cleanest option to me, but the whole thing becomes quite a bit more complex which we should probably try to avoid if possible.

You could do also the following that uses just one map if I understood DCQL correctly:

```
{
  "response": {
[...]

PR #398, opened 2025-01-28
support returning multiple presentations for a single dcql credential query when requested using `multiple`

### LINKED ISSUE #298: Support for selecting multiple credentials via a single dcql
In Velocity Network, a career and education ecosystem, wallets are using PE to select multiple credentials based on a particular claim.

dcql (#266) does not seem to currently support a common use case we have is to return all credentials matching `“vct”: “https://credentialrepository.com/education_degree”`

charsleysa:
Closes #298

Signed-off-by: Stefan Charsley <charsleysa@gmail.com>


openid-4-verifiable-presentations-1_0.md
```diff
@@ -978,7 +981,7 @@ When a VP Token is returned, the respective response includes the following para
 
 `vp_token`:
 : REQUIRED. The structure of this parameter depends on the query language used to request the presentations in the Authorization Request:
- * If DCQL was used, this is a JSON-encoded object; the keys are the `id` values used for the Credential Queries in the DCQL query, and the values are the Verifiable Presentations that match the respective Credential Query. The Verifiable Presentations are represented as strings or objects depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
+ * If DCQL was used, this is a JSON-encoded object; the keys are the `id` values used for the Credential Queries in the DCQL query, and the values are either the Verifiable Presentations that match the respective Credential Query or, if `multiple` in the Credential Query is set to `true`, an array of one or more Verifiable Presentations that match the respective Credential Query. The Verifiable Presentations are represented as strings or objects depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
```
babisRoutis, line 984: Dear @charsleysa 
I find this description, about the value corresponding for each key (of the `vp_token`) some how confusing. 

I believe if we change the plural, then we can simplifying to something like

If DCQL was used, this is a non-empty JSON-encoded object which contains entries having as
- key the `id` value used for the Credential Query, and 
- value an array of one or more Verifiable Presentations that match the respective credential Query, depending on the `multiple` option. The array will contain either strings or JSON objects depending on the format ...

charsleysa: The intention of the wording is to have the value be a single Verifiable Presentation (not an array) if `multiple` is false, and an array of one or more Verifiable Presentations if the `multiple` is true.

Your suggested wording infers that the value will always be an array but will only contain more than one Verifiable Presentation when `multiple` is true. Is this what you intended?

sloops77: @charsleysa There is a preference not to have polymorphic types to make implementors jobs easier

charsleysa: Noted, I'll update the PR to use only arrays and remove non-array single Verifiable Presentation as a possible value. This will introduce a breaking change for existing implementations.

charsleysa: PR updated

sloops77 (approved):
lgtm

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2797,6 +2810,7 @@ The technology described in this specification was made available from contribut
    * require `typ` value in request object to be `oauth-authz-req+jwt`
    * add `SessionTranscript` requirements 
    * use claims path pointer for mdoc based credentials
+   * support returning multiple presentations for a single dcql credential query when requested using `multiple`
```
TimoGlastra, line 2813: Shouldn't this be added to 25? as 24 has already been published? 
https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html

charsleysa: If that's the case then it would probably be best if one of the maintainers should increment the version numbers rather than including it in this PR.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -669,8 +669,8 @@ unknown properties.
 
 ## Credential Query {#credential_query}
 
-A Credential Query is an object representing a request for a presentation of one
-Credential.
+A Credential Query is an object representing a request for a presentation of one or more matching
+Credentials.
```
Sakurann, line 673: I understand the motivation behind this change, and yet, I am not sure about the working... I think we should make it very explicit that one credential query is intended to describe exactly one credential. but, there is also a way to get copies of that one credential

openid-4-verifiable-presentations-1_0.md
```diff
@@ -978,7 +981,7 @@ When a VP Token is returned, the respective response includes the following para
...
- * If DCQL was used, this is a JSON-encoded object; the keys are the `id` values used for the Credential Queries in the DCQL query, and the values are the Verifiable Presentations that match the respective Credential Query. The Verifiable Presentations are represented as strings or objects depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
+ * If DCQL was used, this is a JSON-encoded object which contains entries having: the key as the `id` value used for a Credential Query in the DCQL query; and the value as an array of one or more Verifiable Presentations that match the respective Credential Query. When `multiple` is not `true`, the array MUST contain only one Verifiable Presentation. There MUST NOT be any entry in the JSON-encoded object for optional Credential Queries when there are no matching Credentials for the respective Credential Query. The Verifiable Presentations are represented as strings or objects depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
```
Sakurann, line 984:
```suggestion
 * If DCQL was used, this is a JSON-encoded object containing entries where: the key is the `id` value used for a Credential Query in the DCQL query; and the value is an array of one or more Verifiable Presentations that match the respective Credential Query. When `multiple` is  omitted, or set to `false`, the array MUST contain only one Verifiable Presentation. There MUST NOT be any entry in the JSON-encoded object for optional Credential Queries when there are no matching Credentials for the respective Credential Query. Each Verifiable Presentation is represented as a string or object, depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
```

Sakurann, line 984: what are you trying to say with `There MUST NOT be any entry in the JSON-encoded object for optional Credential Queries when there are no matching Credentials for the respective Credential Query.`? if "do not return what was not requested", wasn't this somewhere in the processing rules?

openid-4-verifiable-presentations-1_0.md
charsleysa: My understanding is that this is not for getting multiple copies of one credential, but instead getting multiple credentials that satisfy the credential query.

The example provided is in #298 is that the credential query asks for credentials of vct type `https://credentialrepository.com/education_degree` and the wallet can respond with multiple credentials each representing a different degree.

openid-4-verifiable-presentations-1_0.md
charsleysa: The intention of this wording was to clarify that if optional credentials are requested, but there are no matching credentials, do not include an empty entry in the JSON-encoded object (eg. `"my_optional_credential": []`).

If this is already covered in another section I can remove this wording.

danielfett (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -978,7 +981,7 @@ When a VP Token is returned, the respective response includes the following para
...
- * If DCQL was used, this is a JSON-encoded object; the keys are the `id` values used for the Credential Queries in the DCQL query, and the values are the Verifiable Presentations that match the respective Credential Query. The Verifiable Presentations are represented as strings or objects depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
+ * If DCQL was used, this is a JSON-encoded object containing entries where: the key is the `id` value used for a Credential Query in the DCQL query; and the value is an array of one or more Verifiable Presentations that match the respective Credential Query. When `multiple` is  omitted, or set to `false`, the array MUST contain only one Verifiable Presentation. There MUST NOT be any entry in the JSON-encoded object for optional Credential Queries when there are no matching Credentials for the respective Credential Query. Each Verifiable Presentation is represented as a string or object, depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Verifiable Presentations.
```
tplooker, line 984: I think we need to take a step back and consider how we reference a returned verifiable presentation against not only a credential query but also a credential set otherwise with complex queries it may become increasingly difficult for the verifier to understand how the query was satisfied when a credential set was used. If we accept this proposal as is, without discussing that I think we might make that issue worse.

charsleysa: @tplooker My understanding is these changes shouldn't add complexity to credential set functionality.

The only change is that the RP can now receive more than one presentation if they have specified so using `multiple` in the credential query, so instead of performing a single check on a single presentation the RP would perform the same check on each presentation in the array. Essentially the checks are now always performed in a loop.

This complexity is added to the check at the credential query level rather than the credential set level.

Though I do see your point that there is more complexity being added overall, moreso when factoring in other desired changes such as #397.

Sakurann: what are the implications? reconsidering a change introduced in this PR? adding a note in this PR? opening an issue? no action?

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
Sakurann: ok, yes.

babisRoutis (approved):

openid-4-verifiable-presentations-1_0.md
sloops77: @tplooker i think this is an edge case but also that the spec makes sense for the use cases I'm dealing with. To make foo and bar concrete an RP could ask for either my transcript OR multiple course badges

c2bo (approved):

paulbastian (changes requested):
I  have a hard time imagining how to do this from a Wallet UX perspective. Do we have any implementation experience on this? As it seems easy to add this feature lateron, I would be hesitant to complicate the DCQL syntax prior this experience.

babisRoutis:
> I have a hard time imagining how to do this from a Wallet UX perspective. Do we have any implementation experience on this? As it seems easy to add this feature lateron, I would be hesitant to complicate the DCQL syntax prior this experience.

For what is worth, I have implemented such a use-case (see https://github.com/openid/OpenID4VP/issues/298#issuecomment-2606682393).
Wallet was use-case specific, though.

I believe that there would be additional use-cases that could take advantage of this feature. Certainly, yhe way that this is handled by Wallet UX is important, but perhaps this is another concern.

charsleysa:
> I have a hard time imagining how to do this from a Wallet UX perspective. Do we have any implementation experience on this? As it seems easy to add this feature lateron, I would be hesitant to complicate the DCQL syntax prior this experience.

We have implemented this as well. The UI is pretty simple with checkbox selection. For credential queries where only one can be selected, the other credentials become disabled. When multiple can be selected, they all remain enabled.

Screenshot_20250221_215129_Edge

paulbastian (approved):
Ok thanks for the feedback. As there is no possbility to take back requested change, I will approve to move this forward

openid-4-verifiable-presentations-1_0.md
Sakurann: i created this issue https://github.com/openid/OpenID4VP/issues/439 to continue the discussion if needed.

PR #401, opened 2025-01-29
make consistent the use of prefixes in the client_id prefixing

### LINKED ISSUE #376: Client Identifier Schemes violate RFC 3986
Among the client identifier schemes defined in "[Section 5.10.4. Defined Client Identifier Schemes](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-5.10.4)" of [OpenID4VP](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html), `redirect_uri`, `verifier_attestation`, `x509_san_dns`, and `x509_san_uri` violate the scheme definition in "[Section 3.1. Scheme](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.1)" of "[RFC 3986 Uniform Resource Identifier (URI): Generic Syntax](https://www.rfc-editor.org/rfc/rfc3986.html)". Simply put, scheme strings must not include underscores. Is the DCP WG aware of this?

The name "Client Identifier Scheme" is misleading enough to make people believe it is valid as a URI scheme. Unless there is a strong reason to prefer underscores over hyphens, I believe it would be better to change them to `redirect-uri`, `verifier-attestation`, `x509-san-dns`, and `x509-san-uri`. This would also make it easier to parse client identifiers as URIs in many programming languages, simplifying the process of extracting client identifier schemes. For reference, I’ve attached experimental codes in Java and Ruby, and their result.

```java
import java.net.URI;

public class ClientIdentifierSchemeTest
{
    public static void main(String[] args)
    {
        // Client identifier examples from the OpenID4VP spec.
        String[] ids = new String[] {
            "redirect_uri:https://client.example.org/cb",
            "https://federation-verifier.example.com",
            "did:example:123#1",
            "verifier_attestation:verifier.example",
            "x509_san_dns:client.example.org",
            "x509_san_uri:https://client.example.org/cb",
            "web-origin:https://verifier.example.com"
        };

        for (String id : ids)
        {
            System.out.format("%-45s %s URI%n", id, validity(id));
        }

        System.out.print("\n..... Replacing underscores with hyphens .....\n\n");

        for (String id : ids)
        {
            // Replace underscores with hyphens.
            id = id.replaceAll("_", "-");

            // Parse as a URI.
            URI uri = URI.create(id);

            // Print the identifier and the scheme part.
            System.out.format("%-45s scheme=%s%n", id, uri.getScheme());
        }
    }

    private static String validity(String string)
    {
        try
        {
            // Parse the string as a URI.
            new URI(string);

            // Valid URI.
            return "Valid";
        }
        catch (Exception cause)
        {
            // Invalid URI.
            return "Invalid";
        }
    }
}
```

```ruby
#!/usr/bin/env ruby

require 'uri'

def main(args)
  # Client identifier examples from the OpenID4VP spec.
  ids = [
    'redirect_uri:https://client.example.org/cb',
    'https://federation-verifier.example.com',
    'did:example:123#1',
    'verifier_attestation:verifier.example',
    'x509_san_dns:client.example.org',
    'x509_san_uri:https://client.example.org/cb',
    'web-origin:https://verifier.example.com'
  ]

  ids.each do |id|
    printf("%-45s %s URI\n", id, validity(id))
  end

  printf "\n..... Replacing underscores with hyphens .....\n\n"

  ids.each do |id|
    # Replace underscores with hyphens.
    id = id.tr("_", "-")

    # Parse as a URI.
    uri = URI.parse(id)

    # Print the identifier and the scheme part.
    printf("%-45s scheme=%s\n", id, uri.scheme)
  end
end

def validity(string)
  begin
    # Parse the string as a URI.
    URI.parse(string)

    # Valid URI.
    return "Valid"
  rescue
    # Invalid URI.
    return "Invalid"
  end
end

main(ARGV)
```

**Result** (Both programs output the same result):

```
redirect_uri:https://client.example.org/cb    Invalid URI
https://federation-verifier.example.com       Valid URI
did:example:123#1                             Valid URI
verifier_attestation:verifier.example         Invalid URI
x509_san_dns:client.example.org               Invalid URI
x509_san_uri:https://client.example.org/cb    Invalid URI
web-origin:https://verifier.example.com       Valid URI

..... Replacing underscores with hyphens .....

redirect-uri:https://client.example.org/cb    scheme=redirect-uri
https://federation-verifier.example.com       scheme=https
did:example:123#1                             scheme=did
verifier-attestation:verifier.example         scheme=verifier-attestation
x509-san-dns:client.example.org               scheme=x509-san-dns
x509-san-uri:https://client.example.org/cb    scheme=x509-san-uri
web-origin:https://verifier.example.com       scheme=web-origin
```

bc-pi:
~~with a default allowance~~ 

~~to help with~~ closes #376


openid-4-verifiable-presentations-1_0.md
```diff
@@ -522,11 +522,15 @@ Depending on the Client Identifier Scheme, the Verifier can communicate a JSON o
...
 
-For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interprete the Client Identifier as referring to a pre-registered client.
+For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
-From this definition, it follows that pre-registered clients MUST NOT contain a `:` character in their Client Identifier.
+If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Scheme value, the Wallet MAY treat the Client Identifier as having a default Client Identifier Scheme. 
+
+For example, an Authorization Request containing a `client_id` value of `=https://federation-verifier.example.com` could be interpreted by the Wallet as referring to an OpenID Federation Entity Identifier, with the default Client Identifier Scheme being `openid_federation`.
```
timcappalli, line 531: should this be: 

```suggestion
For example, an Authorization Request containing a `client_id` value of `https://federation-verifier.example.com` could be interpreted by the Wallet as referring to an OpenID Federation Entity Identifier, with the default Client Identifier Scheme being `openid_federation`.
```

Sakurann (changes requested):
please separate this PR into two PRs: one that includes what helps with #376 and editorial changes; and the other with what is a normative change we have not agreed in the WG (openid-federation and decentralzied-identifier). otherwise, will close.

bc-pi:
> please separate this PR into two PRs: one that includes what helps with #376 and editorial changes; and the other with what is a normative change we have not agreed in the WG (openid-federation and decentralzied-identifier). otherwise, will close.

I understand that you don't see this as important and are not happy to see it appear given the many competing priorities and constrained timelines from many external factors. But it is no less deserving of the WGs attention than many other items that have been brought to the front of the queue. I will respectfully suggest that "otherwise, will close" is an understandable but wholly inappropriate response. Please reconsider.

bc-pi:
and even though it shouldn't work in this direction, whatever happens here in OpenID4VP will likely have ramifications in regular old OAuth. And it's not okay to push special treatment of DIDs and OpenID Federation onto that layer.

aaronpk:
Sorry for being so blunt, but if the DCP WG wants to do things the OpenID4VP way while ignoring feedback from the underlying OAuth perspective, that's fine with me, but then I expect that those decisions will not be brought back to the OAuth WG at the IETF. Making a decision here that is knowingly not compatible with work happening at the IETF is a decision the DCP WG will have to live with as things develop independently at the IETF.

If the changes in this PR are not made, then I expect to hear no complaints when this topic is brought back up at the IETF.

selfissued (changes requested):
This PR proposes breaking changes to the hard-won consensus position on Client ID prefixes developed in PR #263 .  It would break existing client_id values using OpenID Federation and existing ones that are DIDs.  That seems ill-advised.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -554,11 +558,11 @@ Location: https://wallet.example.org/universal-link?
...
+* `openid_federation`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Since the Entity Identifier is already defined to start with `https:`, this Client Identifier Scheme MUST NOT be prefixed additionally. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `openid_federation:https://federation-verifier.example.com`.
```
aaronpk, line 561:
```suggestion
* `openid_federation`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `openid_federation:https://federation-verifier.example.com`.
```

this sentence isn't necessary anymore since this PR adds an explicit prefix for OpenID Federation

bc-pi: Thank you. I actually thought I'd removed that sentence but apparently I missed it.

aaronpk:
I would like this to be a DCP working group decision that it's acceptable to diverge from what the OAuth WG may do with client ID schemes in the IETF. Can we get this decision onto the agenda for the next call?

Sakurann (commented):
I think `decentralized-identifier:did:` sounds very weird, and it's only DIDs that use `did:` anyway, unlike `https:` that is used not just by openid federation, so why `did:` is a preferential treatment? I support `openid_federation:` but am less sure about `decentralized-identifier:`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -554,11 +558,11 @@ Location: https://wallet.example.org/universal-link?
...
+* `decentralized-identifier`: This value indicates that the Client Identifier is a DID defined in [@!DID-Core]. The request MUST be signed with a private key associated with the DID. A public key to verify the signature MUST be obtained from the `verificationMethod` property of a DID Document. Since DID Document may include multiple public keys, a particular public key used to sign the request in question MUST be identified by the `kid` in the JOSE Header. To obtain the DID Document, the Wallet MUST use DID Resolution defined by the DID method used by the Verifier. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter as defined in (#new_parameters). Example Client Identifier: `decentralized_identifier:did:example:123`.
```
bc-pi, line 563:
```suggestion
* `decentralized_identifier`: This value indicates that the Client Identifier is a DID defined in [@!DID-Core]. The request MUST be signed with a private key associated with the DID. A public key to verify the signature MUST be obtained from the `verificationMethod` property of a DID Document. Since DID Document may include multiple public keys, a particular public key used to sign the request in question MUST be identified by the `kid` in the JOSE Header. To obtain the DID Document, the Wallet MUST use DID Resolution defined by the DID method used by the Verifier. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter as defined in (#new_parameters). Example Client Identifier: `decentralized_identifier:did:example:123`.
```

bc-pi:
> I think `decentralized-identifier:did:` sounds very weird, and it's only DIDs that use `did:` anyway, unlike `https:` that is used not just by openid federation, so why `did:` is a preferential treatment? I support `openid_federation:` but am less sure about `decentralized-dentifier:`

The idea here is to have all client id schemes use a prefix. That's the consistency part. The text prior to this PR has a prefix for the client id schemes except for Federation and DIDs, which used the URI scheme as both the client id scheme and part of the client id value. This PR adjusts things so that all client id schemes consistently have a prefix part that is distinct from the value part. I chose `decentralized_identifier:` for Decentralized Identifiers (DIDs) v1.0 because I assumed someone would say `did:did` sounded weird or was confusing. It doesn't have to be `decentralized_identifier:` though.  But it does need to be something. Would `w3c_did:` or `w3c_decentralized_identifiers_v1_0:`or just`did:` (but that would then have double "did:did") or `decentralized_id:` or `w3c_did_core_architecture_data_model_and_representations:` or `did_core:` or `did_core_2021:` or something along those lines be less weird to your eye?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -554,11 +558,11 @@ Location: https://wallet.example.org/universal-link?
...
+* `decentralized_identifier`: This value indicates that the Client Identifier is a DID defined in [@!DID-Core]. The request MUST be signed with a private key associated with the DID. A public key to verify the signature MUST be obtained from the `verificationMethod` property of a DID Document. Since DID Document may include multiple public keys, a particular public key used to sign the request in question MUST be identified by the `kid` in the JOSE Header. To obtain the DID Document, the Wallet MUST use DID Resolution defined by the DID method used by the Verifier. All Verifier metadata other than the public key MUST be obtained from the `client_metadata` parameter as defined in (#new_parameters). Example Client Identifier: `decentralized_identifier:did:example:123`.
 
-The following is a non-normative example of a header and a body of a signed Request Object when Client Identifier scheme is a `did`:
+The following is a non-normative example of a header and a body of a signed Request Object when the Client Identifier scheme is a `decentralized_identifier`:
```
bc-pi, line 565:
```suggestion
The following is a non-normative example of a header and a body of a signed Request Object when the Client Identifier scheme is `decentralized_identifier`:
```

bc-pi:
> One of the core principles of the OpenID Foundation is that we collaboratively support each other's work.

That's a lovely sounding platitude on the surface of things but is actually absurd and untenable.

selfissued:
@aaronpk, the security vulnerability was closed by PR #263 unambiguously defining that Client IDs beginning with `https:` are OpenID Federation Entity Identifiers.  Relaxing that to leave the interpretation up to implementations would re-open the security vulnerability.

It's not a special case.  It's a clear definition to remove ambiguity.

aaronpk:
That's pretty effectively demonstrating the exact problem I mentioned before.

Defining `https` schemes as OpenID Federation identifiers is an OpenID-centric view of the world. Doing that is fine in an OpenID context of course. But someone not using OpenID Federation should not be _prevented_ from also using `https` the scheme for client IDs. So when a draft is brought to the IETF that establishes `https` scheme client IDs, the response cannot be that this scheme is already defined by OpenID Federation so can't be used elsewhere.

selfissued:
For context, what IETF draft are you referring to, Aaron? OpenID4VP is an OpenID draft.

aaronpk:
I presented [Client ID Metadata Document](https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-metadata-document/) at IETF 121, and there were comments during the session about the potential conflict with the OpenID Federation  use of `https` URLs. That led to a discussion between myself, Dr Fett, and Joseph resulting in presenting [OAuth 2.0 Client ID Scheme](https://datatracker.ietf.org/doc/draft-parecki-oauth-client-id-scheme/) at an interim meeting in January. In both cases, I mentioned that there are a number of deployments of systems that currently use `https` URLs as client IDs that do not use (and likely never will use) OpenID Federation.

aaronpk:
Right, which is why I said I'm fine with either outcome for OpenID4VP. The problem is that whatever gets decided here will ultimately trickle back up to OAuth IETF specs. That's why I asked for a DCP working group decision that it is acceptable to make a decision here that will _not_ be brought back to IETF.

selfissued:
Ok, that makes sense.  Of course, both officially and practically, we all participate as individuals...

selfissued:
@bc-pi , in response to my comment:

> One of the core principles of the OpenID Foundation is that we collaboratively support each other's work.

you wrote:

> That's a lovely sounding platitude on the surface of things but is actually absurd and untenable.

At a personal level, I sure hope we can all commit to working collaboratively and supporting each other's work, because our lives (and specs) will all be better for it.  This isn't abstract to me, nor a platitude.  I've experienced the non-collaborative alternative in practice and so have you Brian - it's the way the VCWG works.  I care too much about the OpenID Foundation and our work together to see it head in that direction.  For what it's worth...

peppelinux:
It makes no sense to define `decentralized_identifier` in addition to `did`. In this I agree with Kristina. `did` alone is quite telling.

I appreciate `https` and not because this necessarily has to correspond exclusively to openid federation. I like being able to implement an entity identification mechanism that involves an https URL, indipendently of the discovery mechanisms supported. I am aware that via https there may be different ways of making a discovery, considering different trust frameworks. I am and I imagine everyone is aware that this client scheme is about the client's unique identifier and not about the trust framework that involved several other aspects about discovery and trust evaluation.

with the proposal consolidated to date in the openid4vp draft I still see interesting, creative degrees of freedom. For example, an x509_san_uri can support hybrid discovery mechanisms, combining certificates and metadata resolution on well-known endpoints (and this does not actually exclude openid federation: please don't get excited to reduce these freedoms).

I would like mechanisms that scale with different trust frameworks and that allow a single implementation work in different contexts without requiring any particular modifications or specializations. 

I regret destroying any expected technical arbitrariness aimed at simplifying our world, unfortunately or fortunately, the future is hybrid, the technical rules must not aim to solve the problems of trust frameworks but accommodate their implementations.

I like how Brian writes and I like some of this PR. I do not support the specialization of client schemes, for what my opinion is worth I ask for its removal. I would keep only the editorials in this PR, reverting all about client id schemes parameters.

aaronpk:
@peppelinux you're completely missing the original issue that led to this proposal, described in #376.

Currently, the client ID values are mixed between being prefixed with a value that is not in the client ID value.

```
            "redirect_uri:https://client.example.org/cb",
            "https://federation-verifier.example.com",
            "did:example:123#1",
            "verifier_attestation:verifier.example",
            "x509_san_dns:client.example.org",
            "x509_san_uri:https://client.example.org/cb",
            "web-origin:https://verifier.example.com"
```

Giving these all to a URI parser results in errors because underscores are not allowed in URI schemes. The point brought up in prior discussions is that these aren't actually URIs, rather they are their own construct of `prefix:value`. You would have to write a bespoke parser for these, which is done easily by looking for the first colon and treating everything before the colon as the client ID scheme and everything after as the value. That results in this table:

image

Now you can see the two outliers here are `did` and `https`, where the value is missing some characters. So these then have to be special-cased to put back together the correct client ID value.

This PR makes these consistent with the others by adding a prefix to everything:

image

peppelinux:
@aaronpk thank you for having put me in the right direction.

I would not add `decentralized_identifier` to `did` because `did` means decentralized identifier and ity represents a repetition. The did URIs seem to work properly without a bespoke parser.

I would not add `x509_dns_uri` or `openid_federation`, because `https` alone is quite telling. Hinting the use of federation or X.509 can be achieved using `trust_chain` or `x5c` within the JWS headers. I would say the same about `web-origins` 

but I don't want to bother you, in general I'm a little worried about the over-specification of every detail. 

This way of doing things leaves no room for an implementation to use a client_id with more than one trust framework or trust evaluation mechanism. In general it doesn't scale... But this is an old discussion. 

To carry out this PR in a positive way, it is only necessary to resolve the URIs that cause exceptions in the parsers without touching those that do not cause exceptions, also in consideration of the current implementations and the desire not to introduce breaking changes where not strictly necessary

tplooker:
Im in favour of this change from @bc-pi. I raised concerns around the in-consistent application of prefixes to client IDs when the original proposal was made and I think both @bc-pi and @aaronpk have further distilled arguments for why it would be a bad idea to let that continue.

lj-raidiam (changes requested):
https://github.com/openid/OpenID4VP/pull/401#issuecomment-2654195261

Sakurann (approved):
after the WG discussion, I think this makes sense.

examples/response/vp_token_raw_ldp_vp.json
```diff
@@ -40,7 +40,7 @@
     "type": "Ed25519Signature2018",
     "created": "2021-03-19T15:30:15Z",
     "challenge": "n-0S6_WzA2Mj",
-    "domain": "https://client.example.org/cb",
+    "domain": "x509_san_uri:https://client.example.org/cb",
```
jogu, line 43: I don't think I have the necessary knowledge, can someone confirm that this change is correct please? It looks at best odd to combine x509_san_uri with did holder & did verification method.

charsleysa (approved):

examples/response/vp_token_raw_ldp_vp.json
bc-pi: I don't honestly remember why that seemed like the right/best thing to change there and [vp_token_multiple_vps.json](https://github.com/openid/OpenID4VP/pull/401/files#diff-ef8c16339812a418241f69e9dcdafb60a1fc89dc227b1bbb26ff362d76b1d15fR44) above which appear in https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#section-8.1.2 but it being openid federation (b/c there's no prefix prior to this PR and that's the special case) didn't seem right either. Maybe `redirect_uri:` would make more sense?

jogu: That seems fair. And I since discovered that the `domain` claim is actually explicitly defined to be the client_id ( https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-B.1.2.3 ) so I agree this change is at least not introducing a new problem.

bc-pi:
> @peppelinux you're completely missing the original issue that led to this proposal, described in #376.
> 
> Currently, the client ID values are mixed between being prefixed with a value that is not in the client ID value.
> 
> ```
>             "redirect_uri:https://client.example.org/cb",
>             "https://federation-verifier.example.com",
>             "did:example:123#1",
>             "verifier_attestation:verifier.example",
>             "x509_san_dns:client.example.org",
>             "x509_san_uri:https://client.example.org/cb",
>             "web-origin:https://verifier.example.com"
> ```
> 
> Giving these all to a URI parser results in errors because underscores are not allowed in URI schemes. The point brought up in prior discussions is that these aren't actually URIs, rather they are their own construct of `prefix:value`. You would have to write a bespoke parser for these, which is done easily by looking for the first colon and treating everything before the colon as the client ID scheme and everything after as the value. That results in this table:
> 
> image
> 
> Now you can see the two outliers here are `did` and `https`, where the value is missing some characters. So these then have to be special-cased to put back together the correct client ID value.
> 
> This PR makes these consistent with the others by adding a prefix to everything:
> 
> image

Just a note @aaronpk per one of the little items we chatted about last night (on the rooftop bar in Bangkok!) is that https://github.com/openid/OpenID4VP/pull/430 was merged recently, which introduced `x509_hash` and removed `x509_san_uri`.

examples/response/ldp_vp.json
```diff
@@ -46,7 +46,7 @@
     "type": "Ed25519Signature2018",
```
OR13, line 46: This example is old, and should be updated to support https://www.w3.org/TR/vc-data-integrity/#dataintegrityproof

openid-4-verifiable-presentations-1_0.md
```diff
@@ -554,11 +558,11 @@ Location: https://wallet.example.org/universal-link?
...
+* `openid_federation`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `openid_federation:https://federation-verifier.example.com`.
```
OR13, line 561: I support this change.

OR13 (approved):

examples/response/ldp_vp.json
bc-pi: that's kinda outside the scope of this PR though

peppelinux:
The challenge lies in the fact that this consistency renders trust evaluation mechanisms mutually exclusive, making the use of one mechanism incompatible with others. Implementations capable of supporting multiple trust evaluation mechanisms may find limited accommodation within the current technical specifications.

Ultimately, implementers who wish to support multiple trust evaluation mechanisms will be compelled to deploy separate instances of the same solution with different configurations. This approach does not align with the scalable model I envision. I would prefer more flexible schemes, similar to those found in the `https` scheme.

OR13:
@peppelinux sounds like you are maybe wishing for conneg, that would be basically a 2 param system, where you have resource identifier, and resource representation.

jogu:
Hi @lj-raidiam 

> Potential conflict with the Client ID Metadata draft is an issue, however the bigger issue with this PR is that when merged it would quite severely impact other OpenID specification that is in a quite advanced state. Why would we even do that?

Because it turns out the current OID4VP specification conflicts with other ecosystems and we need to look at the big picture if we want to avoid introducing long term inconsistencies & confusion. This was information the DCP WG did not take into account when previously discussing this topic (as far as I know, no one raised it, unfortunately), and it is absolutely right to revisit the previous decision in light of the new information given none of the specifications are final yet.

Note that the wording in this pull request explicitly allows for backwards compatibility with existing deployments (though it may well be that no one currently has wallets that support federation deployed into production with end users, so it may be that exception isn't needed). Wallets updating to final would need to support the new syntax (in addition to the fallback path), but this is far from the only change that wallets need to make in order to support the latest version of VP.

> Perhaps adding an optional metadata type parameter only for `https` identifiers, which when present would convey information about use of a scheme other than the default one (openid federation) could be a solution?

Are you talking about additional wallet metadata? Unfortunately (in the general case) the verifiers have no way to fetch wallet metadata before making a request so I can't see how this would help.

peppelinux:
> @peppelinux sounds like you are maybe wishing for conneg, that would be basically a 2 param system, where you have resource identifier, and resource representation.

@OR13 according to the content of this PR and the awareness to have this claimed consistency, probably the best solution would be make the client_id an array of multiple values, accomodating the coexistence, within a single implementation, of multiple schemes.

an alternative to the client_id[ARRAY] could be the use of an optional parameter called `alternatives_client_ids`[ARRAY].

I would not advace for having the previous approach using client_id and client_id_scheme as different parameters, since it is now exceeded.

tlodderstedt (approved):

peppelinux:
> Note that the wording in this pull request explicitly allows for backwards compatibility with existing deployments (though it may well be that no one currently has wallets that support federation deployed into production with end users, so it may be that exception isn't needed). Wallets updating to final would need to support the new syntax (in addition to the fallback path), but this is far from the only change that wallets need to make in order to support the latest version of VP.

@jogu the italian wallet impl profiles uses openid federation [1] and has more than 4 millions of active Users [0]

AFAIK, many organization participating at the SPRIN-D competitions expresses their support for OpenID Federation. Some of them started important initiatives, like Animo at the OWF [2], other simply supports Federations (wwWallet, Lissi) ...

Probably like you, I don't have time to collect data and study trends, I just get some news from newspaper and ppl reaching me with information.

[1] https://italia.github.io/eid-wallet-it-docs/versione-corrente/en/trust.html
[2] https://openwallet.foundation/2025/02/25/openid4vc-dcql-and-openid-federation-three-new-fundamental-typescript-projects-incubated-at-openwallet-foundation/?utm_campaign=OpenWallet%20Foundation&utm_content=325814393&utm_medium=social&utm_source=linkedin&hss_channel=lcp-90554582

sakimura:
This is a big thread and I do not claim to fully appreciate the depth and breadth of the issue. 
However, I do believe that if one WG wants to make a decision that impacts a product of another WG, then coordination should be done before making such a change. 

Thanks, 

Nat

sakimura (changes requested):
Requesting coordination with the OpenID Connect working group, per my comment https://github.com/openid/OpenID4VP/pull/401#issuecomment-2746798968

aaronpk:
I just submitted a PR that modifies the text in this PR: #468 

The intent of this PR is to keep the `openid_federation` prefix as the canonical version for processing, but unambiguously define the processing of plain `https` URLs to mean the `openid_federation` prefix. This way the plain `https` URLs are still treated as a fallback/legacy, while existing deployments of OpenID Federation can continue to use them without the prefix since they are defined unambiguously.

I've also updated the IETF draft to remove references to all the schemes defined in OpenID4VP, defining only the framework for defining prefixes, and leaving the possibility of specific profiles or ecosystems to define their default behavior for un-prefixed Client Identifiers. You can see the draft text of that [here](https://drafts.aaronpk.com/oauth-client-id-prefix/draft-parecki-oauth-client-id-prefix.html#section-3.2)

The goal is to keep the IETF draft generic, allowing more specific profiles or ecosystems to define their own default behavior for un-prefixed Client IDs. I personally don't have an issue with OpenID4VP defining the default behavior of `https` URLs as long as it's defined only within the context of OpenID4VP and doesn't spill out to unrelated communities and ecosystems.

Sakurann:
chair hat off.

I really do not want to reserve `https:` only for openid federation spec in the DCP WG. I am ok to say that in the ecosystems where only one https based client id scheme is used, https: can be used without being prefixed, so in the ecosystems where only openid federation is used, `https:` can be used, but when these are entering ecosystems where multiple trust mechanisms co exist, there needs to be a way to differentiate (disambiguate even?) openid federation from other schemes in a more deterministic way. 

I understand `openid-federation:` scheme is a breaking change for openid federation spec, and if that is not ok with openid federation authors or connect wg, i would rather remove openid-federation client id scheme from openid4vp 1.0 than reserve something as "significant" as https for one specific mechanism,

Sakurann:
chair hat on. @sakimura, since the client identifier scheme is defined in OpenID4VP, DCP WG chairs would expect DCP WG, not the Connect WG, to define its meaning and set the rules. Multiple people have initiated number of discussions with the authors of OpenID Federation and IETF OAuth WG - this work is definitely happening under coordination.

examples/response/ldp_vp.json
bc-pi: #487

sakimura:
> chair hat on. @sakimura, since the client identifier scheme is defined in OpenID4VP, DCP WG chairs would expect DCP WG, not the Connect WG, to define its meaning and set the rules. Multiple people have initiated number of discussions with the authors of OpenID Federation and IETF OAuth WG - this work is definitely happening under coordination.

My point is that if the change requires normative changes to the other spec that is sometimes used in conjunction, then there should be coordination, for example, if it requires entity identifier syntax to be changed or a new parameter for the identifier type in the Federation spec. 

These aspects were discussed in the AB/C call today (2025-04-03) and the suggestion at the end of the call was to push the processing rule to the VP client (relying party) and the wallets. (The call abruptly dropped there so no consensus, however.) 
If that is the case, it sounds like it is purely falling into DCP WG. (I would be worried a bit with the backward compatibility/migration issues for the already deployed ecosystem like Italy, but that also is a discussion to be had in this group.)

Sakurann:
WG discussion
- proposal: openid4vp mandates that client_ids are always prefixed for the consistency (including openid federation) (what is mostly in this PR though minor changes might be needed); while OpenIDFederation defines a case where no prefixing is needed.
    - **this does not impact interoperability and is not a breaking change for openid federation specification,** because when the client/verifier uses openid federation within the context of openid4vp, it formulates the request according to the rules defined in openid4vp. and likewise, when the wallet processes openid4vp request from the verifier who used openid federation, the wallet needs to do so within the context of openid4vp. 
    - context: openid4vp is at the end of the dependency chain. only vp depends on vp :D

examples/response/ldp_vp.json
Sakurann: Could you please fix it in this PR, tho, Brian....? given the time pressure, etc...? don't think it is that big of a scope creep...

bc-pi: I also don't know how to fix it or if it actually needs fixing. Being old doesn't make it wrong.

bc-pi: I don't have the knowledge of any of the W3C VC WG/CG stuff to know if or what fixing is needed or appropriate and I'm definitely not going to acquire that knowledge anytime soon.

danielfett: Doing this here in this PR will not help anyone or speed up anything.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -522,11 +526,12 @@ Location: https://wallet.example.org/universal-link?
     7D%7D%7D
 ```
 
-* `https`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Since the Entity Identifier is already defined to start with `https:`, this Client Identifier Prefix MUST NOT be prefixed additionally. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier Prefix is used. Example Client Identifier: `https://federation-verifier.example.com`.
+<<<<<<< HEAD
```
bc-pi, line 529:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -468,7 +468,7 @@ Note: "https://self-issued.me/v2" is a symbolic string and can be used as an `au
...
 
-The Client Identifier Prefix is a string that MAY be communicated by the Verifier in a prefix within the `client_id` parameter in the Authorization Request. A fallback to pre-registered Clients as in [@!RFC6749] remains in place as a default mechanism in case no Client Identifier Prefix was provided. A certain Client Identifier Prefix may require the Verifier to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
+The Client Identifier Prefix is a string that MAY be communicated by the Verifier in a prefix within the `client_id` parameter in the Authorization Request. A fallback to pre-registered Clients as in [@!RFC6749] or a default Client Identifier Prefix is in place as a default mechanism in case no Client Identifier Prefix was provided. A certain Client Identifier Prefix may require the Verifier to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
```
bc-pi, line 471:
```suggestion
The Client Identifier Prefix is a string that MAY be communicated by the Verifier in a prefix within the `client_id` parameter in the Authorization Request. A fallback to pre-registered Clients as in [@!RFC6749] remains in place as a default mechanism in case no Client Identifier Prefix was provided. A certain Client Identifier Prefix may require the Verifier to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -490,11 +490,15 @@ Depending on the Client Identifier Prefix, the Verifier can communicate a JSON o
 
 ### Fallback
 
-If a `:` character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client. This is equivalent to the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
+If a `:` character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client. This is equivalent to the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
```
bc-pi, line 493:
```suggestion
If a `:` character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client. This is equivalent to the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -490,11 +490,15 @@ Depending on the Client Identifier Prefix, the Verifier can communicate a JSON o
...
+If a `:` character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client. This is equivalent to the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms. 
 
-For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interprete the Client Identifier as referring to a pre-registered client.
+For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
-From this definition, it follows that pre-registered clients MUST NOT contain a `:` character in their Client Identifier.
+If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Prefix value, the Wallet MAY treat the Client Identifier as having a default Client Identifier Prefix.
```
bc-pi, line 497:
```suggestion
If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Prefix value, the Wallet can treat the Client Identifier as referring to a pre-registered client or it may refuse the request.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -490,11 +490,15 @@ Depending on the Client Identifier Prefix, the Verifier can communicate a JSON o
...
 
-For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interprete the Client Identifier as referring to a pre-registered client.
+For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
-From this definition, it follows that pre-registered clients MUST NOT contain a `:` character in their Client Identifier.
+If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Prefix value, the Wallet MAY treat the Client Identifier as having a default Client Identifier Prefix. 
+
+For example, an Authorization Request containing a `client_id` value of `https://federation-verifier.example.com` could be interpreted by the Wallet as referring to an OpenID Federation Entity Identifier, with the default Client Identifier Prefix being `openid_federation`.
```
bc-pi, line 499:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -536,15 +540,15 @@ Body
...
 
-To use the Client Identifier Prefixes `https`, `did`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be confidential clients. This might require changes to the technical design of native apps as such apps are typically public clients.
+To use the Client Identifier Prefix `https`, `decentralized_identifier`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be confidential clients. This might require changes to the technical design of native apps as such apps are typically public clients.
```
bc-pi, line 551:
```suggestion
To use the Client Identifier Prefixes `https`, `decentralized_identifier`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be capable of securely storing private key material. This might require changes to the technical design of native apps as such apps are typically public clients.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1384,7 +1388,10 @@ The following is a non-normative example of a `vp_formats_supported` parameter:
 ```
 
 `client_id_prefixes_supported`:
-: OPTIONAL. Array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `https`, `verifier_attestation`, `did`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles of this specification.
+: OPTIONAL. Array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `https`, `verifier_attestation`, `decentralized_identifier`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles of this specification.
+
+`client_id_prefix_default`:
+: OPTIONAL. A string containing the value of the default Client Identifier Prefix that the Wallet will use when appropriate.
```
bc-pi, line 1394:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2089,9 +2096,9 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
 
 The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
 
-Parameters defined by a specific client identifier scheme (such as the `trust_chain` parameter for the OpenID Federation client id scheme) are also supported over the W3C Digital Credentials API.
+Parameters defined by a specific client identifier prefix (such as the `trust_chain` parameter for the OpenID Federation client id prefix) are also supported over the W3C Digital Credentials API.
```
bc-pi, line 2099:
```suggestion
Parameters defined by a specific Client Identifier Prefix (such as the `trust_chain` parameter for the OpenID Federation Client Identifier Prefix) are also supported over the W3C Digital Credentials API.
```

bc-pi (commented):
review with proposed changes aimed at https://github.com/openid/OpenID4VP/pull/401#issuecomment-2776912290

Sakurann:
@sakimura @peppelinux @selfissued @lj-raidiam please re-review this PR in light of new understanding in DCP WG that is documented in https://github.com/openid/OpenID4VP/pull/401#issuecomment-2776912290

examples/response/ldp_vp.json
Sakurann: one less issue and one less PR in my view. but sure, I will bug you all to do and/or review that editorial PR when the time comes

openid-4-verifiable-presentations-1_0.md
```diff
@@ -536,15 +539,15 @@ Body
...
 
-To use the Client Identifier Prefixes `https`, `did`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be confidential clients. This might require changes to the technical design of native apps as such apps are typically public clients.
+To use the Client Identifier Prefixes `https`, `decentralized_identifier`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be capable of securely storing private key material. This might require changes to the technical design of native apps as such apps are typically public clients.
```
jogu, line 550: A "https" got missed:

```suggestion
To use the Client Identifier Prefixes `openid_federation`, `decentralized_identifier`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be capable of securely storing private key material. This might require changes to the technical design of native apps as such apps are typically public clients.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1384,7 +1387,7 @@ The following is a non-normative example of a `vp_formats_supported` parameter:
 ```
 
 `client_id_prefixes_supported`:
-: OPTIONAL. Array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `https`, `verifier_attestation`, `did`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles of this specification.
+: OPTIONAL. Array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `https`, `verifier_attestation`, `decentralized_identifier`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles of this specification.
```
jogu, line 1390:
```suggestion
: OPTIONAL. Array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `openid_federation`, `verifier_attestation`, `decentralized_identifier`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles of this specification.
```

openid-4-verifiable-presentations-1_0.md
bc-pi: Thank you. I think this was the result a conflict resolution merge error on my part.

selfissued (changes requested):
As requested, here's my re-review of the PR in light of https://github.com/openid/OpenID4VP/pull/401#issuecomment-2776912290.

It remains that this PR contradicts the clear normative requirement in OpenID Federation that "the RP employs its Entity Identifier as the Client ID".  This would cause unnecessary developer confusion, in that we'd be asking them to use one Client ID value in some contexts and a different one in others.

We already created a good solution to the issue in PR #263 and that's what's been in OpenID4VP ever since.  It unambiguously defines both the semantics of https Client IDs when using OpenID4VP and clearly defines how to use OpenID Federation with OpenID4VP.

I could also live with Aaron's version of this PR in PR #468, because it also unambiguously defines https Client IDs to refer to OpenID Federation when using OpenID4VP.

I note that yesterday the provision to be able to have a default Client ID prefix was removed from this PR.  That makes this PR worse than it was before, by making OpenID4VP incompatible with not only OpenID Federation, but also with any other systems using https Client IDs.

In conclusion, this PR as it stands is strictly worse from an interop and developer confusion point of view than what's already in in the spec.  I believe that we should therefore either close it with no action or, if people prefer, merge it after applying Aaron's updates to it that unambiguously define what https Client IDs mean when using OpenID4VP.  Thank you.

aaronpk:
> It remains that this PR contradicts the clear normative requirement in OpenID Federation that "the RP employs its Entity Identifier as the Client ID".

...and OpenID Federation contradicts the [clear normative requirement of RFC6749](https://www.rfc-editor.org/rfc/rfc6749.html#section-2.2) that client identifiers are assigned by the authorization server. I point this out because it's fine for an extension spec to change these kinds of requirements of the spec it's extending.

> This would cause unnecessary developer confusion, in that we'd be asking them to use one Client ID value in some contexts and a different one in others.

I would like to hear from some actual developers about whether this would cause confusion.

My instinct is that developers will know that they are in the VP context, so would appreciate the consistency of having only one way to use an OpenID Federation Entity Identifier in the flow in that context. 

In other words, this:

* The verifier adds the `openid_federation` prefix to the Entity Identifier when talking to a wallet, and uses the plain Entity Identifier when performing a web-based redirect flow.

is simpler to understand than:

* The verifier adds the `openid_federation` prefix to the Entity Identifier when talking to a wallet that requires the prefix but might not add the prefix if it's in an ecosystem where the wallet has a default handling for unprefixed Entity Identifiers, and uses the plain Entity Identifier when performing a web-based redirect flow.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -504,7 +507,7 @@ Confusing Verifiers using a Client Identifier Prefix with those using none can l
...
+* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). An example Client Identifier value is `redirect_uri:https://client.example.org/cb`. Requests using the `redirect_uri` Client Identifier Prefix cannot be signed because there is no method for the Wallet to obtain a trusted key for verification. Therefore, implementations requiring signed requests cannot use the `redirect_uri` Client ID Prefix.
```
bc-pi, line 510:
```suggestion
* `redirect_uri`: This value indicates that the Client Identifier is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). An example Client Identifier value is `redirect_uri:https://client.example.org/cb`. Requests using the `redirect_uri` Client Identifier Prefix cannot be signed because there is no method for the Wallet to obtain a trusted key for verification. Therefore, implementations requiring signed requests cannot use the `redirect_uri` Client ID Prefix.
```

danielfett (approved):

timcappalli (approved):

aaronpk (approved):

sakimura:
> In other words, this:
> 
> * The verifier adds the `openid_federation` prefix to the Entity Identifier when talking to a wallet, and uses the plain Entity Identifier when performing a web-based redirect flow.
> 
> is simpler to understand than:
> 
> * The verifier adds the `openid_federation` prefix to the Entity Identifier when talking to a wallet that requires the prefix but might not add the prefix if it's in an ecosystem where the wallet has a default handling for unprefixed Entity Identifiers, and uses the plain Entity Identifier when performing a web-based redirect flow.

I do agree on this. Of course, the client has to implement change in the client logic. 

Though I have to say this, it is an echo of https://github.com/openid/OpenID4VP/issues/505. I do hate the use of ":" as the delimiter. It introduces problems with legacy client_ids that are using scheme-based identifiers, as well as urn-based ones. Like @bc-pi points out, delimiters like "|" are much better.

bc-pi:
Honestly, if we're worried about "unnecessary developer confusion", we should remove OpenID Federation entirely from the 4VC specs. DIDs too.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -492,9 +492,12 @@ Depending on the Client Identifier Prefix, the Verifier can communicate a JSON o
 
 If a `:` character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client. This is equivalent to the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interprete the Client Identifier as referring to a pre-registered client.
+For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
-From this definition, it follows that pre-registered clients MUST NOT contain a `:` character in their Client Identifier.
+If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Prefix value, the Wallet can treat the Client Identifier as referring to a pre-registered client or it may refuse the request.
```
Sakurann, line 497:
```suggestion
If a : character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Prefix value, the Wallet can treat the Client Identifier as referring to a pre-registered client, or a legacy client that does not support Client Identifier Prefix. Unless the Wallet can establish enough security context with it, it may refuse the request.
```

Sakurann:
WG discussion:
- agreement to adding `decentralized_identity:` and `openid-federation:`
- remaining discussion is about adding behavior of non-prefixed client identifiers that start with `https://`
  - another option would be to remove any text about https:// now and re-add it during 1.0 wglc

sakimura (approved):

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -522,11 +525,11 @@ Location: https://wallet.example.org/universal-link?
...
+* `openid_federation`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier Prefix is used. Example Client Identifier: `openid_federation:https://federation-verifier.example.com`.
```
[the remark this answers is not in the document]
sakimura: I have raised https://github.com/openid/OpenID4VP/issues/512 for this.

openid-4-verifiable-presentations-1_0.md
bc-pi: I believe this was a suggestion from Nat during the call that was found to be not acceptable soon after to the rough census of the WG.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -492,9 +492,12 @@ Depending on the Client Identifier Prefix, the Verifier can communicate a JSON o
 
 If a `:` character is not present in the Client Identifier, the Wallet MUST treat the Client Identifier as referencing a pre-registered client. This is equivalent to the [@!RFC6749] default behavior, i.e., the Client Identifier needs to be known to the Wallet in advance of the Authorization Request. The Verifier metadata is obtained using [@!RFC7591] or through out-of-band mechanisms.
 
-For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interprete the Client Identifier as referring to a pre-registered client.
+For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
-From this definition, it follows that pre-registered clients MUST NOT contain a `:` character in their Client Identifier.
+If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Prefix value, the Wallet can treat the Client Identifier as referring to a pre-registered client or it may refuse the request.
+
```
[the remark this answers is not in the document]
bc-pi: I believe this suggestion was made during the call and the discussion therein was that it was not acceptable soon after to the rough census of the WG.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
bc-pi: Thanks @sakimura

lj-raidiam:
> * agreement to adding `decentralized_identity:` 
I respect the decision; however, we should ask ourselves whether ultimate parsing convenience should  prevail over keeping it logical and not prefixing "did" with de facto another "did".

>   * option 2: define a behavior for not-prefixed client identifiers that start with `https://`
What does stop us from going with this option and merging this PR? Would making federation that default mechanism in addition to prefixes introduced by the vp itself or oauth in the future be harmful in any way for the vp spec? If yes, how?

In addition to that looking at the notes from the WG meeting I wanted to draw attention of the WG to one detail. Correct me if I am wrong, but the client_id_scheme can't be directly compared to the prefix mechanism. When the client_id_scheme param was present, the entity identified with a certain clientid could span across multiple namespaces (client_id _schemes). A single client using https based identifier, could use various schemes, but it was still a single client. This is not the case now when the prefix defining the namespace is included in the client id and the above mentioned rule is not in the spec.

As long as this is not an issue for non-global identifiers it has an impact on https. Given the above, why shouldn't we be worried that prefixing https, which is designed to uniquely identify resources globally, would lead to either clients identified with the same URLs yet supporting multiple metadata discovery or trust establishment mechanisms having different identities? ([also commented here](https://github.com/openid/OpenID4VP/pull/401#issuecomment-2645376059))

peppelinux (approved):

peppelinux:
Trust Frameworks might decide to describe how to use a single, general purpose, client_id scheme identifier, with multiple trust evaluation methods. For instance, `https` or `x509_san_uri` might be reintroduced by implementation profiles supported by trust frameworks.

I would have preferred client identifiers that are not strictly tied to trust evaluation technology, as defining a client's unique identity serves a different purpose than assessing its trust. The current PR results in client ID schemes that involve distinct, mutually exclusive trust evaluation mechanisms.

It also doesn't give any opportunity for a client to use more than a single client_id scheme or client_id, forcing implementers supporting more than a single trust evaluation method to create different deploys of an equivalent RP configuration. This doesn't match the vision I personally have about scalability of the solutions within multiple contexts.

While this choice has been extensively debated and only partially considered, it introduces a breaking change for systems currently in production. Nevertheless, it's important to remain flexible and seek solutions that simplify our work.

I truly appreciate the efforts of this working group and express my support by facilitating your work. This is the reason for my approval.

danielfett:
@lj-raidiam @peppelinux A single client ID that supports different schemes (or trust evaluation technologies) may be considered a feature in some settings, but it can be, in fact, a security vulnerability. This was discussed [here](https://github.com/openid/OpenID4VP/issues/124) and the working group decided that a single, clearly identified mechanism is the way forward. That also explains why a generic "https" prefix would have to be tied to a single mechanism (federation) and cannot stand for a range of mechanisms (federation, metadata from somewher else, x509-something, etc.). So "https" would be a hard-coded exception for federation that is not of much value for anyone not using federation.

openid-4-verifiable-presentations-1_0.md
lj-raidiam, line 528: The meaning of Client Identifier is not perfectly consistent in this spec and even in the scope of this bullet point. Usually it refers to the entire client identifier (also referred to as "full Client Identifier"), while sometimes it is orig_client_id (aka "usual Client Identifier"). Could we make it more specific which Client Identifier identifier it is. 

In the redirect_uri we got that _(without the prefix `redirect_uri:`)_ that didn't look great but did the job. Now when it got removed over there too this section got ambiguous. This problem does not apply only to this bullet.

lj-raidiam:
> @lj-raidiam @peppelinux A single client ID that supports different schemes (or trust evaluation technologies) may be considered a feature in some settings, but it can be, in fact, a security vulnerability. This was discussed [here](https://github.com/openid/OpenID4VP/issues/124) and the working group decided that a single, clearly identified mechanism is the way forward. 

@danielfett +1. I remember that. We got that PR that introduced changes that addressed that problem. Nevertheless, after merging we got https and dids without a prefix. Https URLs were pure play identifiers without info on mechanism being part of the identifier. It complied with the Federation because it (and not only it) uses URLs to identify entities and locate metadata. 

> That also explains why a generic "https" prefix would have to be tied to a single mechanism (federation) and cannot stand for a range of mechanisms (federation, metadata from somewher else, x509-something, etc.). So "https" would be a hard-coded exception for federation that is not of much value for anyone not using federation.

It was a compromise that worked yet was federation-centric. I recognize that there may be many mechanisms that leverage https and ideally they should use pure play https URLs as identifiers, but they'll end up being prefixed in VP irrespective whether there's a default option or not. 

Having a default option to maintain the current behaviour maybe wouldn't provide value to others (whoever they are or will be) but I don't see how it could be harmful. Additionally it would be important for harmonization with and increase of chances of adoption of other openid spec with signifiant potential. Merging this PR without federation as the default option on the other hand will cause incompatibility between these specs. That's why I would encourage everyone to rethink why we would not want to do it.

openid-4-verifiable-presentations-1_0.md
lj-raidiam: > Automatic Registration as defined in [@!OpenID.Federation] MUST be used.
With this new wording, what implementers should do to satisfy both the requirements of both VP and OpenID Federation automated registration, given that the latter requires the client_id to be the entity ID, while the former requires adding the openid_federation prefix to it?

bc-pi: Is there really ambiguity or confusion about the prefix being a prefix and the other part being the part used as the value to do stuff with? I would like to have the text be clear and do acknowledge that these kinds of things can be difficult to describe precisely in spec text. Can you suggest something that would make the text more clear? Something that works in the context of this larger change that has prefixing applied equally to all.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
bc-pi: it seems perfectly fine as is

openid-4-verifiable-presentations-1_0.md
```diff
@@ -480,7 +480,7 @@ In the `client_id` Authorization Request parameter and other places where the Cl
 
 Here, `<client_id_prefix>` is the Client Identifier Prefix and `<orig_client_id>` is an identifier for the Client within the namespace of that prefix. See (#client_identifier_prefixes) for Client Identifier Prefixes defined by this specification. 
 
-Wallets MUST use the presence of a `:` (colon) character to determine whether a Client Identifier Prefix is used. If a `:` character is present, the Wallet MUST interpret the Client Identifier according to the Client Identifier Prefix, here defined as the string before the (first) `:` character. If the Wallet does not support the Client Identifier Prefix, the Wallet MUST refuse the request.
+Wallets MUST use the presence of a `:` (colon) character and the content preceding it to determine whether a Client Identifier Prefix is used. If a `:` character is present and the content preceding it is a recognized and supported Client Identifier Prefix value, the Wallet MUST interpret the Client Identifier according to the given Client Identifier Prefix. The Client Identifier Prefix is defined as the string before the (first) `:` character.
```
bc-pi, line 483:
```suggestion
Wallets MUST use the presence of a `:` (colon) character and the content preceding it to determine whether a Client Identifier Prefix is used. If a `:` character is present and the content preceding it is a recognized and supported Client Identifier Prefix value, the Wallet MUST interpret the Client Identifier according to the given Client Identifier Prefix. The Client Identifier Prefix is defined as the string before the (first) `:` character. Note that implementations should not assume that the presence of a `:` character implies that the entire value can be processed as a valid URI. Instead, the specific processing rules defined for the specified Client Identifier Prefix (see (#client_identifier_prefixes)) should be used to parse the `client_id` value.
```

openid-4-verifiable-presentations-1_0.md
lj-raidiam: @bc-pi 
I am referring to the difference in e.g. these two statements.
_"This value indicates that the Client Identifier is an Entity Identifier"_
_Example Client Identifier: `openid_federation:https://federation-verifier.example.com`._

I don't have specific wording that I could suggest but I think that the definition of this prefix should be something like: This prefix indicates that OpenID Federation is in use and the value following it is an Entity Identifier ...

bc-pi: 02290542420880867ce77e02f0ceb56249f38d12 attempts to bring some clarity to that treatment for all the prefixes.

jogu (approved):

jogu:
Just for the record, the chairs and other working group members have had several discussions with Lukasz and we appreciate his feedback - we're happy to discuss further once he's back from holiday.

@Razumain We didn't reply to your comment before merging - apologies for that - but please rest assured that it had been read and considered - to respond a little:

> The base problem is that the client identifier has overloaded semantics, which I argue is problematic for several reasons.

This isn't a new problem in this PR. And we don't disagree that it's not great, and the initial proposal was not to overload client id - see the client_id_scheme parameter in this older draft: https://openid.net/specs/openid-4-verifiable-presentations-1_0-18.html#section-5

That proposal unfortunately had a security issue ( https://github.com/openid/OpenID4VP/issues/124 ) and the only sensible way to solve it was to move the scheme into the client_id. We are expecting the OAuth WG to take a similar approach.

> One problem I can see right away with this solution is how we plan to deal with trust in our national eIDAS wallet implementation in Sweden. Here we must support both Trusted List and OpenID federation. It is very possible to allow services to be available through multiple trust infrastructures. But we do not plan to do the client registration, or to enforce it. This is up to each party to decide what functions of the infrastructure they want to use (Trusted List, OpenID federation resolvers, or OpenID federation client registration with chain validation).
>
> If the format of the client identifier limits our freedom to build and offer trust infrastructures, we have big problems.

Note that even before this PR, it would not have been compliant with the specification to use a single client identifier to mean multiple trust schemes. https:// client ids could only mean OpenID Federation.

I think the problem you mention can be solved in at least two different ways:

1. You could define a new Client Id Prefix with the properties you want of supporting both Trusted List and OpenID Federation.
2. In the DC API the verifier can sign the request using multiple methods, allowing the wallet to select which one it wants to use.
3.  In the advanced post mode for request_uri, the wallet can indicate to the verifier which methods it supports.

> Isn't there a better alternative than prefixing?

We have not been able to come up with one.

PR #403, opened 2025-01-30
clarified that redirect_uri scheme must not be used with rfc9101

### LINKED ISSUE #392: Conflicting requirement on request object signature
Hi

The standard says in section [5.10.4](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-defined-client-identifier-s):
> * redirect_uri: This value indicates that the Client Identifier (without the prefix redirect_uri:) is the Verifier's Redirect URI (or Response URI when Response Mode direct_post is used). The Authorization Request MUST NOT be signed.

The requirement on the absence of a signature seems to be in conflict with the requirement from section [5.11.1](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-request-uri-response) if combined with the `request_uri` parameter:
> The Request URI response MUST be an HTTP response with the content type "application/oauth-authz-req+jwt" and the body being a signed, optionally encrypted, request object as defined in [[RFC9101](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#RFC9101)].

So, is there a signature required on the request object returned from the `request_uri` resource if the `client_id_scheme` is `redirect_uri`?

Sakurann:
resolves #392

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -7,7 +7,7 @@ keyword = ["security", "openid", "ssi"]
 
 [seriesInfo]
 name = "Internet-Draft"
-value = "openid-4-verifiable-presentations-1_0-24"
+value = "openid-4-verifiable-presentations-1_0-25"
 status = "standard"
 
 [[author]]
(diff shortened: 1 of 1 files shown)


jogu (changes requested):
The justification on the issue for this change is:

> rfc9101 does not allow alg = none


However JAR does allow alg none, it's only disallowed if `requred_signed_request_object` is set to true in server metadata: https://www.rfc-editor.org/rfc/rfc9101.html#require_signed_request_object

Sakurann:
@jogu
so you are suggesting when using JAR with client id scheme redirect_uri, alg = none must be used?
isn't prohibiting using JAR with client id scheme redirect_uri safer and more in spirit of sentiment around alg=none ?

jogu:
My thinking didn't go as far as a suggestion, it was more at the "if we're making a breaking change to the spec we should make sure we have a clear correct reason for doing so" level.

I hate alg=none as much as everyone else. The conformance suite does support alg:none here (it calls it "request_uri_unsigned" in the dropdown choice) and I can see over the past few months a few people have used it with redirect_uri client scheme, though whether they did so 100% intentionally I'm not sure, at least a few of the tests did pass so people support it.

I guess if we do want to discourage alg=none, we should do that consistently across all client id schemes?

awoie (changes requested):
I agree with @jogu. I believe this is what some people are using and we never disallowed that.

Sakurann:
so do you want me to change this to "please use alg=none when you want to use redirect_uri with signed_request"?

jogu:
I'm fine with some clarification along those lines

jogu:
closing this as we seem to have consensus not to go in this direction - let's discuss on issue.

PR #404, opened 2025-01-30
Add language on nonce/client ID guidance for W3C VCs and ISO mdocs

### LINKED ISSUE #171: Define how OID4VP parameters are carried over to VPs normatively
There is an implicit understanding that parameters such as `nonce`, `client_id` and `client_id_scheme` are used in VPs but we don't have explicit language how this is done for specific format identifiers.

I propose to add language that define that profiles must describe how. For profiles defined by OID4VP, this language should be added to the credential format sections.

awoie:
This PR may fix #171 :
- Add language for nonce/client ID binding for W3C VCs and ISO mdocs

I am not able to add similar language to Anoncreds. I suggest to create a new issue for this if required. I'm not an Anoncreds expert.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2797,6 +2797,9 @@ The technology described in this specification was made available from contribut
 
    [[ To be removed from the final specification ]]
 
+   -25
+   * add language on client ID and nonce binding for ISO mdocs and W3C VCs
+
    -24
 
    * add mdoc specific `intent_to_retain` mechanism, using the definition from 18013-5
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2102,6 +2102,11 @@ This `presentation_definition` parameter contains a single `input_descriptor` el
 
 #### Presentation Response
 
+The following requirements apply to the `nonce` and `aud` claims of the Verifiable Presentation:
+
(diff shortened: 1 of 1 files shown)


bc-pi (approved):

c2bo:
The AnonCreds one seems to be a bit more tricky. I agree a new issue/PR for that one sounds reasonable

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2791,6 +2797,9 @@ The technology described in this specification was made available from contribut
 
    [[ To be removed from the final specification ]]
 
+   -25
+   * add language on client ID and nonce binding for ISO mdocs and W3C VCs
```
c2bo, line 2801:
```suggestion
   -25
   
   * add language on client ID and nonce binding for ISO mdocs and W3C VCs
```

c2bo (approved):

Sakurann:
for anoncreds, could we just add one line saying that resulting credential needs to be bound to client_id and nonce (without normatively defining how), if we do not have that language yet?

Sakurann (approved):

c2bo:
> for anoncreds, could we just add one line saying that resulting credential needs to be bound to client_id and nonce (without normatively defining how), if we do not have that language yet?

I can create a PR for that and ask some people more familiar with AnonCreds to check/approve.

paulbastian (approved):

Sakurann:
please open a separate issue for anoncreds :)

PR #421, opened 2025-02-11
Make transaction data credential format specific and define details for sd-jwt and mdoc

### LINKED ISSUE #418: Clarification on `transaction_data_hashes`
Currently, on d23/d24 the specification contains the following description:

> transaction_data_hashes: Array of hashes, where each hash is calculated using a hash function over the strings received in the transaction_data request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format, but it has to be included in the mechanism used for the proof of possession of the Credential that is signed using the user-controlled key

In addition, in the SD-JWT-VC section there is the following (I guess non-normative) example of a KB-JWT

```json
{
  "nonce": "n-0S6_WzA2Mj",
  "aud": "x509_san_dns:client.example.org",
  "iat": 1709838604,
  "sd_hash": "Dy-RYwZfaaoC3inJbLslgPvMp09bH-clYP_3qbRqtW4",
  "transaction_data_hashes": [ "fOBUSQvo46yQO-wRwXBcGqvnbKIueISEL961_Sjd4do" ]
}
```

I guess that each element of `transaction_data_hashes` should be calculated as described in the above text, but it also needs to be encoded using Base64 URL encoded (no padding).

I don't know if it is implied, yet I miss a reference on the base64 encoding `transaction_data_hashes` to the KB-JWT.

### LINKED ISSUE #423: Broaden transaction data requirements
**Context:** This enables the QES creation use case where transaction data contains parameters for advanced e-signature creation. In these cases, the transaction data do not get hashed in their JSON representation format using a single hash algorithm. Instead, the X.509 credential format specifies specific ways of processing, that still meet the broadened requirements.

**Proposed changes:**

- When `transaction_data_hashes_alg` is omitted, do not require a default value of `sha-256`. This enables a credential format to specify hash algorithms.
- Broaden `transaction_data_hashes` specification in VP token. This enables advanced e-signature (AdES) processing rather than straight hashing of a serialised JSON array.

**Draft PR:** https://github.com/openid/OpenID4VP/pull/421

sander:
This enables the QES creation use case where transaction data contains parameters for advanced e-signature creation. In these cases, the transaction data do not get hashed in their JSON representation format using a single hash algorithm. Instead, the X.509 credential format specifies specific ways of processing, that still meet the broadened requirements.

Highlight of changes:

- Leave it up to credential format specs to define transaction data support and requirements.
- Clarify that W3C VC and AnonCreds do not have specified transaction data support.
- Clarify how to support transaction data with mdoc.
- Move `transaction_data_hashes` and `transaction_data_hashes_alg` into SD-JWT VC format spec.
    - Keep `sha-256` a mandatory default.
    - Remove `sha-256` interoperability promotion remark.
    - Clarify what to hash.
    - Clarify how to represent the digest.

Closes #423 

Closes #418

Relates to #259

Does not address #442, #443


sander:
Done, thanks @babisRoutis.

babisRoutis:
> Done, thanks @babisRoutis.

Pointed description to issue 423.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -314,7 +314,7 @@ This enables the Wallet to assess the Verifier's capabilities, allowing it to tr
...
+    * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` SHOULD be used. To promote interoperability, implementations SHOULD support the sha-256 hash algorithm.
```
c2bo, line 317: We should be more explicit if we remove the default behaviour. Something along the lines of "default value of sha-256 MUST be used unless otherwise conveyed in the credential" if I understood your motivation correctly?

I'd keep the "implementations MUST support the sha-256 hash algorithm." - we want everyone to at least support sha-256 for interop imho.

sander: > We should be more explicit if we remove the default behaviour. Something along the lines of "default value of sha-256 MUST be used unless otherwise conveyed in the credential" if I understood your motivation correctly?

Agreed. Is “MUST unless” a proper formulation? Alternatively, specify here that the credential format SHOULD specify default behaviour, and that if the credential does not specify default behaviour, the default behaviour MUST be to use SHA-256.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1196,7 +1196,7 @@ The Wallet that received the `transaction_data` parameter in the request MUST in
...
 
-* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format, but it has to be included in the mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
+* `transaction_data_hashes`: Representation of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format, but it has to be included in the mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
```
matzf, line 1199: My interpretation of "a hash function over the strings" was to compute a hash over the base64-encoded data in the transaction_data, that is for sha-256: `transaction_data_hashes[i] = base64_urlsafe_nopad_encode(sha256(transaction_data[i]))`.

Now with the phrasing "the data in the strings" this must mean that the hash is to be computed over the (base64-)decoded transaction data, that is `transaction_data_hashes[i] = base64_urlsafe_nopad_encode(sha256(base64_urlsafe_decode(transaction_data[i])))`.

Is this an intended change in behaviour or was my previous interpretation just incorrect?

openid-4-verifiable-presentations-1_0.md
c2bo: I guess the cleanest way for the spec would be to describe priorities - something like this?
1. If `transaction_data_hashes_alg` is provided, the hash function provided in that parameter MUST be used
2. If a hash function is provided in the credential and `transaction_data_hashes_alg` is omitted, the hash function contained in the credential MUST be used; the way this hash function is provided in the different Credential Formats is defined in Appendix B. 
3. `sha-256` SHOULD be used otherwise

tlodderstedt:
@c2bo @sander After re-reading the spec and playing with some examples, I would argue transaction_data_hashes_alg and transaction_data_hashes are designed and might work well for sd-jwt. I'm not so sure for mdoc and I'm pretty sure it does not work well for x.509 (QES). For example, the transaction data type for QES (see below) already contains identifiers for algorithms (hashAlgorithmOID) based on pre-existing standards (ETSi and CSC) and QES/CMS standards define what needs to be incorporated into the signature, some of this data not even passed in the request but obtained from other sources, like time stamping services (in this cases the "T" in "Ades-B-T"). 

```json
{
    "type": "https://cloudsignatureconsortium.org/2025/qes",
    "credential_ids": [
        "certificate_by_policy"
    ],
    "signatureQualifier": "eu_eidas_qes",
    "documentDigests": [
        {
            "signature_format": "P",
            "conformance_level": "Ades-B-T",
            "signed_envelope_property": "Certification",
            "label": "Example Contract",
            "href": "https://protected.rp.example/contract-01.pdf?token=HS9naJKWwp901hBcK348IUHiuH8374",
            "checksum": "sha256-sTOgwOm+474gFj0q0x1iSNspKqbcse4IeiqlDg/HWuI=",
            "access": {
                "type": "OTP",
                "oneTimePassword": "51623"
            }
        }
    ],
    "hashAlgorithmOID": "2.16.840.1.101.3.4.2.1"
}
```

Based on those thoughts, I would suggest to make the way transaction data is included or referenced truly a credential format specific thing. That would mean to move transaction_data_hashes_alg and transaction_data_hashes to appendix B.4 and change the second and third paragraph in section 8.4 to something like

"The Wallet that received the transaction_data parameter in the request MUST include a representation or reference to the data in the in the respective credential presentation. How this is done is specific to each credential format and is defined by each Credential Format, such as those in Appendix B. If the wallet does not support transaction_data parameter, it MUST return an error."

sander:
@tlodderstedt agreed. My original PR was to keep the change as small as possible, but your proposal is the better change.

For mdoc queries, the transaction data is expected to be returned as `DeviceSignedItems`. That means that the document type specification should include attribute definitions matching the transaction data requirements. This is something we cannot specify in detail in OpenID4VCI. Instead CSC should ensure that rulebooks get published/extended with device-signed attribute definitions confirming QES creation in the QTSP-centric model. This could be an extension of the PID rulebook in the case of one-time signatures for example. For the Wallet-centric model, we’re good already.

Since OID4VCI also specifies a W3C VC credential format, what would be the best way to include transaction data there?

openid-4-verifiable-presentations-1_0.md
andprian: I suggest we do not leave room for interpretation and specifically give this information, whether the hash is computed over encoded data or decoded data, etc. because not everyone will interpret the same way or understand all subtleties when reading the text.

andprian, line 1199: @sander could you please indicate the reason why you replaced "array" with "representation"?

sander: See https://github.com/openid/OpenID4VP/pull/421#issuecomment-2708975577: so far, only in the SD-JWT credential format we will know for sure it is an array. For the other formats, we still need to specify.

openid-4-verifiable-presentations-1_0.md
sander: In the current proposal, instead `sha-256` is a mandatory default for SD-JWT, and we leave it up to other credential format specs what they take as default behaviour. What do you think?

openid-4-verifiable-presentations-1_0.md
sander: In the current version, I’ve tried to clarify this in the SD-JWT VC section. But upon writing, I started to prefer the second interpretation by @matzf. It seems to be somewhat more robust since base64url-encoded data seems harder to corrupt than an UTF-8 string representation. So I’ve included it like this in the proposal, what do you think?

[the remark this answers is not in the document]
sander: Clarified in the current version, what do you think?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1192,12 +1190,7 @@ The following is a non-normative example of the payload of the JWT used in the e
...
-
-* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format, but it has to be included in the mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
-* `transaction_data_hashes_alg`: REQUIRED when this parameter was present in the `transaction_data` request parameter. String representing the hash algorithm identifier used to calculate hashes in `transaction_data_hashes` response parameter.
+The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the in the respective credential presentation. How this is done is part of the specification of any Credential Format that supports it, such as some of those in (#format_specific_parameters). If the Wallet does not support `transaction_data` parameter, it MUST return an error.
```
tlodderstedt, line 1193:
```suggestion
The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective credential presentation. How this is done is part of the specification of any Credential Format that supports it, such as some of those in (#format_specific_parameters). If the Wallet does not support `transaction_data` parameter, it MUST return an error.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
  * Contents:
 `["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
 
+### Transaction Data
+
+The SD JWT VC format supports `transaction_data` as specified in (#transaction_data). If used, it is recommended that the transaction data type specification includes the following parameter, in addition to `type` and `credential_ids` from (#new_parameters):
```
tlodderstedt, line 2394: Is it just recommended or is this how the data must be represented in SD-JWTs?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
...
+
+* `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification.
```
tlodderstedt, line 2396: wouldn't it be better to have an array of objects containing the hash and the corresponding hash alg?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
...
+
+If the Wallet received the `transaction_data` parameter in the request, it MUST include in the Key Binding JWT as a top level claim a `transaction_data_hashes` parameter defined below.
+
+* `transaction_data_hashes`: Array of base64url-encoded hashes, where each hash is calculated using a hash function over the data in the base64url-decoded strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
+* `transaction_data_hashes_alg`: REQUIRED when this parameter was present in the `transaction_data` request parameter. String representing the hash algorithm identifier used to calculate hashes in `transaction_data_hashes` response parameter.
```
tlodderstedt, line 2401: why does the text have the transaction_data_hashes_alg twice?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
...
+
+This means that transaction data mechanism cannot be used with SD-JWT VCs without cryptographic key binding and, therefore, do not use KB JWT. If the wallet does not support `transaction_data` parameter, it MUST return an error.
```
tlodderstedt, line 2403:
```suggestion
Note: This means that transaction data mechanism requires use of a SD-JWT VCs with cryptographic key binding.
```

tlodderstedt (changes requested):

openid-4-verifiable-presentations-1_0.md
andprian, line 1193: > If the Wallet does not support `transaction_data` parameter, it MUST return an error.

What would that error be?  We have already defined `invalid_transaction_data` in clause 8.5 but this applies for cases where the wallet supports transaction_data but there is something wrong inside the structure. Should we define a new error code for when `transaction_data` is not supported?

openid-4-verifiable-presentations-1_0.md
sander: If we can still include new mandatory requirements at this stage, let’s indeed make it mandatory.

openid-4-verifiable-presentations-1_0.md
sander: Possibly, yes: https://github.com/openid/OpenID4VP/issues/442. I did not yet look into this. Seems like it needs a separate PR.

openid-4-verifiable-presentations-1_0.md
sander: Which two instances do you mean?

openid-4-verifiable-presentations-1_0.md
andprian: OK, I will create a separate issue for this .

openid-4-verifiable-presentations-1_0.md
matzf: The phrasing "calculated using a hash function over the data in the base64url-decoded strings" seems clear now. The original phrasing "calculated using a hash function over the strings" also seemed clear enough. Only the intermediate state on which I commented seemed quite ambiguous.

Regarding which semantics to adopt, I don't see a compelling  reason to prefer either variant, but I'd avoid changing it now.

sander: Opened a separate issue: https://github.com/openid/OpenID4VP/issues/457

openid-4-verifiable-presentations-1_0.md
sander: I would however suggest to have that detail in a separate issue/PR, to keep the current one’s scope limited to enabling the use of transaction data in other credential formats.

openid-4-verifiable-presentations-1_0.md
sander: Ah I see now: it’s once for inclusion in the request (specifying options), and once consequently in the response (specifying the selected option).

I’d like to have something simpler, but don’t think the current PR should change these semantics.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
  * Contents:
 `["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
 
+### Transaction Data
+
+The SD JWT VC format supports `transaction_data` as specified in (#transaction_data). If used, it is recommended that the transaction data type specification includes the following parameter, in addition to `type` and `credential_ids` from (#new_parameters):
+
+* `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification.
+
+If the Wallet received the `transaction_data` parameter in the request, it MUST include in the Key Binding JWT as a top level claim a `transaction_data_hashes` parameter defined below.
```
c2bo, line 2398:
```suggestion
If the Wallet received the `transaction_data` parameter in the request, it MUST be included in the Key Binding JWT as a top level claim a `transaction_data_hashes` parameter defined below.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
...
+
+* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
```
c2bo, line 2400:
```suggestion
* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter and encoded with base64url encoding. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2251,6 +2252,10 @@ ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4] Ann
 The `SessionTranscript` and `Handover` CBOR structure when the invocation does not use the DC API. Also see (#non-dc-api-invocation).
 * Additional restrictions on Authorization Request and Authorization Response parameters to ensure compliance with ISO/IEC TS 18013-7 [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4]. For instance, to comply with ISO/IEC TS 18013-7 [@ISO.18013-7], only the same-device flow is supported, the `request_uri` Authorization Request parameter is required, and the Authorization Response has to be encrypted.
 
+### Transaction Data
+
+Some document types support some transaction data ((#transaction_data)) to be protected using mdoc authentication, as part of the `DeviceSigned` data structure [@ISO.18013-5]. In those cases, the specifications of these document types include which transaction data types are supported and how the transaction data parameters map to namespaces and device-signed items. If a Wallet receives a request with a `transaction_data` type that is not specified for the document type, the Wallet MUST reject the request due to an unsupported transaction data type.
+
```
c2bo, line 2258: Should we create soemthing similar to `transaction_data_hashes_alg` for mdoc as well, referencing the registered hash functions in COSE Algorithms?

I guess this part should be aligned with #330?

openid-4-verifiable-presentations-1_0.md
Sakurann: i think we should do it in the same PR. I don't think this one is complete without both hash and hash_alg

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1192,12 +1190,7 @@ The following is a non-normative example of the payload of the JWT used in the e
 
 The transaction data mechanism enables a binding between the user's identification/authentication and the user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
 
-The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `transaction_data_hashes` parameter defined below. If the wallet does not support `transaction_data` parameter, it MUST return an error.
-
-Where to include the`transaction_data_hashes` parameter in the response is specific to each credential format and is defined by each Credential Format, such as those in (#format_specific_parameters).
```
Sakurann, line 1197: i am not sure where, but I would really want to keep in the main text 

> Where to include the `transaction_data_hashes` parameter in the request and the response is specific to each credential format and is defined by each Credential Format, such as those in (#format_specific_parameters).

openid-4-verifiable-presentations-1_0.md
sander: That would be incorrect, how about 47f6370?

openid-4-verifiable-presentations-1_0.md
sander: Let’s make this change in a separate issue: https://github.com/openid/OpenID4VP/issues/457. Different developers have had different interpretations, and this further specification deserves a dedicated entry in the OID4VP changelog.

openid-4-verifiable-presentations-1_0.md
sander: Addressed in a505ea3.

openid-4-verifiable-presentations-1_0.md
sander: The main section on transaction data currently contains:
https://github.com/openid/OpenID4VP/blob/a505ea33dc6e732343f814bae976d1fbc5ece332/openid-4-verifiable-presentations-1_0.md?plain=1#L1193

Instead of mentioning `transaction_data_hashes`, which is an SD-JWT specific implementation, it refers to the Credential Format specifications. Do you consider this sufficient?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2174,6 +2167,10 @@ The following is a non-normative example of the Verifiable Presentation in the `
 
 <{{examples/response/ldp_vp.json}}
 
+### Transaction Data
+
+The W3C Verifiable Credentials format does not support transaction data as specified in (#transaction_data).
```
tlodderstedt, line 2172:
```suggestion
The W3C Verifiable Credentials format profile as defined in this specification does not support transaction data as defined in (#transaction_data).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2234,6 +2231,10 @@ The following is the content of the `vp_token` parameter:
 
 <{{examples/response/ac_vp_sd.json}}
 
+### Transaction Data
+
+The AnonCreds format does not support transaction data as specified in (#transaction_data).
```
tlodderstedt, line 2236:
```suggestion
The AnonCreds format profile as defined in this specification does not support transaction data as defined in (#transaction_data).
```

tlodderstedt (approved):
LGTM. Just added small editorial suggestion.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2251,6 +2252,10 @@ ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4] Ann
...
 * Additional restrictions on Authorization Request and Authorization Response parameters to ensure compliance with ISO/IEC TS 18013-7 [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4]. For instance, to comply with ISO/IEC TS 18013-7 [@ISO.18013-7], only the same-device flow is supported, the `request_uri` Authorization Request parameter is required, and the Authorization Response has to be encrypted.
 
+### Transaction Data
+
+Some document types support some transaction data ((#transaction_data)) to be protected using mdoc authentication, as part of the `DeviceSigned` data structure [@ISO.18013-5]. In those cases, the specifications of these document types include which transaction data types are supported and how the transaction data parameters map to namespaces and device-signed items. This mapping includes a specification of any data processing requirements, potentially including any hash function to be applied. If a Wallet receives a request with a `transaction_data` type that is not specified for the document type, the Wallet MUST reject the request due to an unsupported transaction data type.
```
Sakurann, line 2257: I don't understand why it is doctype specific for mdocs? could we define mdocs in #330 ?

```suggestion
```

openid-4-verifiable-presentations-1_0.md
Sakurann, line 2394:
```suggestion
The SD-JWT VC format supports `transaction_data` as specified in (#transaction_data). If used, it is RECOMMENDED that the transaction data type specification includes the following parameter, in addition to `type` and `credential_ids` from (#new_parameters):
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2389,19 @@ __Claim `birthdate`__:
...
+
+If the Wallet received the `transaction_data` parameter in the request, the Wallet MUST include a top level claim parameter `transaction_data_hashes` in the Key Binding JWT as defined below.
+
+* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
+* `transaction_data_hashes_alg`: REQUIRED when this parameter was present in the `transaction_data` request parameter. String representing the hash algorithm identifier used to calculate hashes in `transaction_data_hashes` response parameter.
+
+Note: This means that transaction data mechanism requires use of a SD-JWT VCs with cryptographic key binding.
```
Sakurann, line 2403:
```suggestion
Note: This means that transaction data mechanism requires use of an SD-JWT VC with cryptographic key binding.
```

openid-4-verifiable-presentations-1_0.md
Sakurann, line 2396: please re-add a text that defines behavior when this parameter is not present

```suggestion
* `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1192,12 +1190,7 @@ The following is a non-normative example of the payload of the JWT used in the e
...
-
-* `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. Where in the response this parameter is included is defined by each Credential Format, but it has to be included in the mechanism used for the proof of possession of the Credential that is signed using the user-controlled key.
-* `transaction_data_hashes_alg`: REQUIRED when this parameter was present in the `transaction_data` request parameter. String representing the hash algorithm identifier used to calculate hashes in `transaction_data_hashes` response parameter.
+The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective credential presentation. How this is done is part of the specification of any Credential Format that supports it, such as some of those in (#format_specific_parameters). If the Wallet does not support `transaction_data` parameter, it MUST return an error.
```
Sakurann, line 1193:
```suggestion
The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective credential presentation. How this is done is Credential Format specific and is, therefore defined by each Credential Format, such as those in (#format_specific_parameters).

If the Wallet does not support `transaction_data` parameter, it MUST return an error upon receiving a request that includes it.
```

Sakurann (changes requested):
mostly editorial comments. but I don't understand why mdoc part is doctype specific and i would like to see `If this parameter is not present, a default value of `sha-256` MUST be used.` readded

openid-4-verifiable-presentations-1_0.md
sander: As I understand mdoc, the doc type determines which items the doc can include, and whether these items must be issuer-signed or device-signed. I expect that:

- some doc types will not support device-signed items at all, such as PID for now
- some doc types specify particular device-signed items for particular transaction types
- some doc types will include the catch-all item as suggested in #330

The current text enables all three options, what do you think? If we would instead mandate option 1 in OID4VP, it would require ISO 18013-5 and other mdoc rulebooks to be updated to remain OID4VP-compliant. That seems a too big breaking change to introduce in the current PR.

If we still disagree that all three options are valid, indeed we can instead leave transaction data unspecified for now in the current PR.

openid-4-verifiable-presentations-1_0.md
sander: The default behaviour is now specified elsewhere, at the place where it is actually used:

https://github.com/openid/OpenID4VP/blob/20ec20745be709da24db7e3aa7a17b3373d68c8c/openid-4-verifiable-presentations-1_0.md?plain=1#L2402

I think this is clearer, what do you think?

Sakurann: i think under transaction_data_hashes is more appropriate. 

and I believe `To promote interoperability, implementations MUST support the sha-256 hash algorithm.` is important too

Sakurann: or we can keep both places for clarity. but for optional parameters we usually put default behavior under that specific parameter

openid-4-verifiable-presentations-1_0.md
GarethCOliver: My preference would be that this is done as follows: 

The author of a particular 'transaction_type' specifies the particular element that it should be returned in (and the information to be hashed)
The author (and issuer) of a particular doctype can decide whether or not to include a transaction type's data element in their doctype 

Potentially, separately, we could have a generic one that a doctype author could include, but seems less desirable in practice.

sander: @GarethCOliver Should it always be a single element, or could a single transaction data object be confirmed in a set of elements?

In general, would the current PR text work for you?

GarethCOliver: That's an interesting question. I don't quite understand the usecase for multiple elements for a single transaction type (functionally, the only reason to have different elements is to be able to selective disclose things. If you want more structure within the information returned then the single element can be a cbor object). 

While I don't strongly object, I think that the following isn't desirable:

'...the specifications of these document types include which transaction data types are supported and how the **transaction data parameters map to namespaces and device-signed items.**'

My assumption is that each transaction_data type is going to be specified independently to docTypes. As such, I think it makes more sense for the specifier of transaction_data type provide the mapping to device signed, which should amount to: 

  1) Providing the namespace/field of the element to be used to return the processed transaction_data
  2) The processing rules to populate the data, and the structure of that data. 

This means that if multiple documents use the same transaction_data it'll be more consistent (making an RPs life a little easier, improving practical interop).

Example text: 
It is (recommended?) that each transaction_type defines a DataElement to be used to return the processed transaction data. Additionally it is (recommended?) that it specifies the processing rules, potentially including any hash function to be applied, and the expected resulting structure.

Some document types support some transaction data ((#transaction_data)) to be protected using mdoc authentication, as part of the `DeviceSigned` data structure [@ISO.18013-5]. In those cases, the specifications of these document types include which transaction data types are supported, and issuer includes the relevant data elements in the KeyAuthorization. If a Wallet receives a request with a `transaction_data` type whose DataElement is authorized, the Wallet MUST reject the request due to an unsupported transaction data type.

(Example for transaction_type spec:)
Namespace: com.example.namespace, Field: ExampleTransactionDataType, type = ExampleTransactionDataObject

ExampleTransactionDataObject = {
  "hash" = bstr // Hash of the transaction_data
  "alg" = tstr // String of the algorithm used to hash
}

(We could separately support universally allowing all transaction_data types to be authorized at once, but I'm a little skeptical of that. We could also do that with namespace authorization).

GarethCOliver (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1192,12 +1190,9 @@ The following is a non-normative example of the payload of the JWT used in the e
 
 The transaction data mechanism enables a binding between the user's identification/authentication and the user’s authorization, for example to complete a payment transaction, or to sign specific document(s) using QES (Qualified Electronic Signatures). This is achieved by signing the transaction data used for user authorization with the user-controlled key used for proof of possession of the Credential being presented as a means for user identification/authentication.
 
-The Wallet that received the `transaction_data` parameter in the request MUST include in the response a `transaction_data_hashes` parameter defined below. If the wallet does not support `transaction_data` parameter, it MUST return an error.
+The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective credential presentation. How this is done is Credential Format specific and is, therefore defined by each Credential Format, such as those in (#format_specific_parameters).
```
tlodderstedt, line 1193:
```suggestion
The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective credential presentation. How this is done is Transaction Data Type specific. Credential Formats can give recommendations of how to handle transaction data, such as those in (#format_specific_parameters).
```

Sakurann:
WG discussion:
- keep text with terminology specific to mdoc/sd-jwt vc, but add a generic text that transaction data type rules the content of transaction data and each credential format defines how to incorporate it.
    - how The Wallet that received the `transaction_data` parameter in the request includes **a representation or reference to the transaction data** in the respective credential presentation is **transaction data type specific, not credential format specific.**
- current text for mdoc makes sense: each transaction data type defines namespace etc and each doctype defines whether it authorizes that transaction data type
- agreed to provide the same level of flexibility to sd-jwt vc (keeping `transaction_data_hashes_alg` and `transaction_data_hashes` parameters as recommendation/jfyi, or potentially dropping them even)

leecam (approved):

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2174,6 +2169,10 @@ The following is a non-normative example of the Verifiable Presentation in the `
 
 <{{examples/response/ldp_vp.json}}
 
+### Transaction Data
+
+The W3C Verifiable Credentials format profile as defined in this specification does not support transaction data as defined in (#transaction_data).
```
tlodderstedt, line 2174: I suggest to drop this section as credential formats can give recommendations but do not need to. Implementers can freely decide which credential formats they use in conjunction with their transaction data types.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2174,6 +2169,10 @@ The following is a non-normative example of the Verifiable Presentation in the `
 
 <{{examples/response/ldp_vp.json}}
 
+### Transaction Data
```
tlodderstedt, line 2172:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2174,6 +2169,10 @@ The following is a non-normative example of the Verifiable Presentation in the `
...
+
```
tlodderstedt, line 2173:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
tlodderstedt, line 2174:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2234,6 +2233,10 @@ The following is the content of the `vp_token` parameter:
 
 <{{examples/response/ac_vp_sd.json}}
 
+### Transaction Data
```
tlodderstedt, line 2236:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2234,6 +2233,10 @@ The following is the content of the `vp_token` parameter:
...
+
```
tlodderstedt, line 2237:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2234,6 +2233,10 @@ The following is the content of the `vp_token` parameter:
...
+The AnonCreds format profile as defined in this specification does not support transaction data as defined in (#transaction_data).
```
tlodderstedt, line 2238:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2393,22 @@ __Claim `birthdate`__:
  * Contents:
 `["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
 
+### Transaction Data
+
+It is RECOMMENDED that each transaction data type defines a top level claim parameter to be used in the Key Binding JWT to return the processed transaction data. Additionally it is RECOMMENDED that it specifies the processing rules, potentially including any hash function to be applied, and the expected resulting structure.
+
+Note: When following this recommendation, the transaction data mechanism requires use of an SD-JWT VC with cryptographic key binding.
+
+#### Example Specification
```
tlodderstedt, line 2402: not sure whether we should make this a recommendation, too. Example sounds a bit to weak to me.

openid-4-verifiable-presentations-1_0.md
sander: Done in eaa0426

openid-4-verifiable-presentations-1_0.md
sander: How about Common Mechanism? Like in [RFC 9396 § 2.2](https://www.rfc-editor.org/rfc/rfc9396#name-common-data-fields).

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2251,6 +2248,12 @@ ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4] Ann
...
 
+### Transaction Data
+
+It is RECOMMENDED that each transaction data type defines a data element (`NameSpace`, `DataElementIdentifier`, `DataElementValue`) to be used to return the processed transaction data. Additionally it is RECOMMENDED that it specifies the processing rules, potentially including any hash function to be applied, and the expected resulting structure.
+
+Some document types support some transaction data ((#transaction_data)) to be protected using mdoc authentication, as part of the `DeviceSigned` data structure [@ISO.18013-5]. In those cases, the specifications of these document types include which transaction data types are supported, and the issuer includes the relevant data elements in the `KeyAuthorizations`. If a Wallet receives a request with a `transaction_data` type whose data element is unauthorized, the Wallet MUST reject the request due to an unsupported transaction data type.
```
Sakurann, line 2255:
```suggestion
Some mdoc document types support some transaction data ((#transaction_data)) to be protected using mdoc authentication, as part of the `DeviceSigned` data structure [@ISO.18013-5]. In those cases, the specifications of these document types include which transaction data types are supported, and the issuer includes the relevant data elements in the `KeyAuthorizations`. If a Wallet receives a request with a `transaction_data` type whose data element is unauthorized, the Wallet MUST reject the request due to an unsupported transaction data type.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2387,22 @@ __Claim `birthdate`__:
  * Contents:
 `["6Ij7tM-a5iVPGboS5tmvVA", "birthdate", "1940-01-01"]`
 
+### Transaction Data
+
+It is RECOMMENDED that each transaction data type defines a top level claim parameter to be used in the Key Binding JWT to return the processed transaction data. Additionally it is RECOMMENDED that it specifies the processing rules, potentially including any hash function to be applied, and the expected resulting structure.
+
+Note: When following this recommendation, the transaction data mechanism requires use of an SD-JWT VC with cryptographic key binding.
```
Sakurann, line 2394:
```suggestion
Note: When following this recommendation, the transaction data mechanism requires use of an SD-JWT VC with Cryptographic Holder Binding.
```
`Cryptographic Holder Binding` is a defined term we use

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2387,22 @@ __Claim `birthdate`__:
...
+
+#### Example Specification
+
+The following is a non-normative example that can be included in a transaction data type specification:
+
+* The transaction data includes the following parameter, in addition to `type` and `credential_ids` from (#new_parameters):
```
Sakurann, line 2400:
```suggestion
* The `transaction_data` request parameter includes the following parameter, in addition to `type` and `credential_ids` from (#new_parameters):
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2387,22 @@ __Claim `birthdate`__:
...
+  * `transaction_data_hashes_alg`: OPTIONAL. Array of strings each representing a hash algorithm identifier, one of which MUST be used to calculate hashes in `transaction_data_hashes` response parameter. The value of the identifier MUST be a hash algorithm value from the "Hash Name String" column in the IANA "Named Information Hash Algorithm" registry [@IANA.Hash.Algorithms] or a value defined in another specification and/or profile of this specification. If this parameter is not present, a default value of `sha-256` MUST be used. To promote interoperability, implementations MUST support the sha-256 hash algorithm.
+* The Key Binding JWT includes the following top level claim parameters:
```
Sakurann, line 2402:
```suggestion
* The Key Binding JWT in the response includes the following top level claim parameters:
```

openid-4-verifiable-presentations-1_0.md
Sakurann: how common we want these to be? we say parameters are transaction data type specific and then we define parameters without binding them to any transaction data type?

Sakurann: Thinking about this more, this section on static configuration values is probably the closest to what we are looking for here (https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-profiles-that-define-static)

```suggestion
#### A Profile of Transaction Data in SD-JWT VC
```

Another option would be to move this section to implementation considerations, tho that might make it harder to read..

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2384,6 +2387,22 @@ __Claim `birthdate`__:
...
 
+### Transaction Data
+
+It is RECOMMENDED that each transaction data type defines a top level claim parameter to be used in the Key Binding JWT to return the processed transaction data. Additionally it is RECOMMENDED that it specifies the processing rules, potentially including any hash function to be applied, and the expected resulting structure.
+
+Note: When following this recommendation, the transaction data mechanism requires use of an SD-JWT VC with cryptographic key binding.
+
+#### Example Specification
+
+The following is a non-normative example that can be included in a transaction data type specification:
```
tlodderstedt, line 2398: non-normative example does not match the new heading in my opinion. Suggest to frame it as "example profile" or "one profile"

sander: abcddf1

tlodderstedt (approved):

Sakurann (approved):

PR #422, opened 2025-02-11
dcql_query & presentation_definition are not strings

### LINKED ISSUE #378: Should `dcql_query` and `presentation_definition` be a string when using JAR?
When using JAR, the authorization request is encoded as JSON in a JWT. What I've seen most implementations do with `presentation_definition` field is to keep it as JSON when using JAR, and not encode the JSON object as a string even though the spec mentions "A string containing a Presentation Definition JSON object". We addressed this in our implementation by allowing both string and json.

For DCQL we ran into the same issue again. We encode it as string now, but when doing interop testing with @markuskreusch for DCQL it was encoded as JSON object. We hadn't applied the same fix yet to also allow JSON (we currently always expect string for DCQL).

Encoding the JSON as string is required when using query parameters, but not when using JAR, and it seems to lead to ambiquity. In my opinion it makes sense to allow JSON for JAR (as it's more natural to keep JSON as JSON when putting it in a JSON object). I think most important is to maybe add something in the spec about what to do when using JAR:
- keep requirements same, but mention that when using JAR it should also still be a JSON string, not the JSON object
- make an exception for JAR, and allow/require it to be encoded as JSON instead of string.

jogu:
When putting JSON into JSON for inclusion in request objects, you should definitely not encode it as a string first.

Also switch one of the examples from PE to DCQL as currently the only example showing dcql_query parameter is in DC API appendix.

closes #378


examples/request/request.txt
```diff
@@ -2,6 +2,6 @@ GET /authorize?
   response_type=vp_token
   &client_id=x509_san_dns%3Aclient.example.org
   &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
-  &presentation_definition=...
+  &dcql_query={ ... }
```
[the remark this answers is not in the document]
jogu: I did wonder about that. I think there's a few places where we have `{}`s that probably should be url encoded. We should maybe do them all...

openid-4-verifiable-presentations-1_0.md
```diff
@@ -280,16 +280,18 @@ One exception to this rule is `transaction_data` parameter, and the wallets that
 This specification defines the following new request parameters:
 
 `presentation_definition`:
-: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details.
+: A JSON object containing a Presentation Definition. See (#request_presentation_definition) for more details.
```
danielfett, line 283: I don't think this change is helpful. This should clarify that the encoding varies between the query string (where it is a JSON-encoded string) and the request object and DC API call (where it is an object and it is not additionally JSON-encoded).

openid-4-verifiable-presentations-1_0.md
```diff
@@ -280,16 +280,18 @@ One exception to this rule is `transaction_data` parameter, and the wallets that
...
 `presentation_definition_uri`:
 : A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. See (#request_presentation_definition_uri) for more details.
 
 `dcql_query`:
-: A string containing a JSON-encoded DCQL query as defined in (#dcql_query).
+: A JSON object containing a DCQL query as defined in (#dcql_query).
 
 Exactly one of the following parameters MUST be present in the Authorization Request: `dcql_query`, `presentation_definition`, `presentation_definition_uri`, or a `scope` value representing a Presentation Definition.
 
+In the context of an authorization request according to [RFC6749], parameters containing JSON objects are encoded using the application/x-www-form-urlencoded format of the serialized JSON.
```
danielfett, line 293:
```suggestion
In the context of an authorization request according to [@RFC6749], parameters containing JSON objects are encoded using the application/x-www-form-urlencoded format of the serialized JSON.
```

openid-4-verifiable-presentations-1_0.md
tplooker: I sort of understand this perspective @danielfett, however I'm struggling to see what other interpretation a developer might have about how to convey a JSON object in a query parameter value?

jogu: @danielfett Are we at least agreed the current text is bad? It has resulted in people doing this in request objects:

`"dcql_query": "{\"id\":\"ffc717a3-abaf-4ec3-9c55-a9b8e998874c\",\"name\":\"A`

The way I've done it here (combined with the text below explaining how to put JSON objects into the url query) seems consistent with how it's done in RAR ( https://datatracker.ietf.org/doc/html/rfc9396#section-2 ), and we have the example to try and help people struggling to interpret it too.

charsleysa (approved):

openid-4-verifiable-presentations-1_0.md
danielfett: When this is a DC API call, this is an object (not JSON-encoded).
```suggestion
: An object containing a Presentation Definition. See (#request_presentation_definition) for more details.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -280,16 +280,18 @@ One exception to this rule is `transaction_data` parameter, and the wallets that
...
 `presentation_definition`:
-: A string containing a Presentation Definition JSON object. See (#request_presentation_definition) for more details.
+: A JSON object containing a Presentation Definition. See (#request_presentation_definition) for more details.
 
 `presentation_definition_uri`:
 : A string containing an HTTPS URL pointing to a resource where a Presentation Definition JSON object can be retrieved. See (#request_presentation_definition_uri) for more details.
 
 `dcql_query`:
-: A string containing a JSON-encoded DCQL query as defined in (#dcql_query).
+: A JSON object containing a DCQL query as defined in (#dcql_query).
```
danielfett, line 289: As above
```suggestion
: An object containing a DCQL query as defined in (#dcql_query).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -280,16 +280,18 @@ One exception to this rule is `transaction_data` parameter, and the wallets that
...
 
 Exactly one of the following parameters MUST be present in the Authorization Request: `dcql_query`, `presentation_definition`, `presentation_definition_uri`, or a `scope` value representing a Presentation Definition.
 
+In the context of an authorization request according to [@RFC6749], parameters containing JSON objects are encoded using the application/x-www-form-urlencoded format of the serialized JSON.
```
danielfett, line 293:
```suggestion
In the context of an authorization request according to [@RFC6749], parameters containing objects are transferred as JSON-serialized strings (using the application/x-www-form-urlencoded format as usual for request parameters).
```

openid-4-verifiable-presentations-1_0.md
jogu: If we're going to make that change for that reason I suggest doing it consistently everywhere we use the term 'JSON object' in a separate PR.

tplooker (approved):

jogu:
Confirmed directly with Daniel that he's happy for this to be merged, I transferred his comment about JSON object into a new issue: #446

4 approvals & open a few weeks - merging!

Issue #395, opened 2025-01-23
Clarify that signed requests for DC API should be rejected if the signature can't be verfied


GarethCOliver:
It'd be good to call out that, if the wallet can not verify the signature for the signed version (A.3.2), then they must reject the request. As we were implementing this, initially we assumed that falling back to how we handle the unsigned case would make sense. However, because the origin is not included for signed cases, it would be vulnerable to replay attacks in ways that unsigned is not. 

At least to me, this wasn't clear from reading A.3. 

It'd be nice (IMO) if an unsigned request, and a signed request that can't be verified by a wallet had the same security properties and would be a motivation for including the origin in all cases.

c2bo:
If that kind of downgrade is wanted, couldn't you in theory just change the client_id to web-origin:Origin in the response? That way you would still include the platform-provided origin in the response and keep the same security assumptions as a standard unsigned flow.

That would also clearly signal to the RP that the wallet was not able to verify the trust framework it provided and downgraded to the unsigned flow.

jogu:
Two probably very related clarifying questions:

@GarethCOliver "However, because the origin is not included for signed cases" - do you mean we should state that in the response to a signed request (e.g. the `aud` value in SD-JWT) the `web-origin:https://www.example.com/` value should be used in this case? Or did you mean something else?

@c2bo Included where? In an mdoc response I don't think we return the `client_id` to the verifier other than encoded in the session transcript?

jogu:
It was raised in today's ISO meeting that (if we are going to allow the wallet to accept requests where it can't validate the signature) then we do need to be clear in OID4VP how the wallet should then behave (e.g. if it should use the client_id in the signed request, or as suggested above use one derived from the origin instead), so that the verifier then knows what to expect.

This is particularly important in the case of mdoc/mdl where the verifier will need to know what client_id the wallet decided to use in order to calculate the SessionTranscript correctly.

GarethCOliver:
So I was initially assuming we should just reject if the request signature can't be verified (as expected_origins can't be trusted in that case). @c2bo's suggest is more elegant: 

Essentially treat the platform-provided origin as one of many different potential client_ids. Whichever one the RP trusts, should be included in the aud (and then which one was trusted needs to be included in the response somewhere to prevent needing to attempt verification multiple times). 

[...]

PR #432, opened 2025-02-20
Swap URN for dash delimited protocol identifer

timcappalli:
After all the discussion on the protocol identifier, unfortunately a URN is not going to work. [W3C TAG has guidelines for enum strings, where only dashes can be used for separation](https://github.com/WICG/digital-credentials/issues/191#issuecomment-2661685099).

After discussion on the 2025-02-19 DC API call, this PR changes the URN to a dash delimited identifier (and removes URN specific things like "urn", org, and "protocol".

/cc @bc-pi

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2000,14 +2000,14 @@ And lastly, as part of the request, the Wallet is provided with information abou
 
 ## Protocol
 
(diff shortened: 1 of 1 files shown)


bc-pi (approved):
For the record, I've already been angry with @timcappalli about this and moved past it. Approve.

c2bo:
Depending on the outcome of #308 and #395, it might make sense to also remove the -signed and -unsigned (as one request might be able to combine both variants). Would it make sense to hold this PR until that is resolved?

jogu:
> Depending on the outcome of #308 and #395, it might make sense to also remove the -signed and -unsigned (as one request might be able to combine both variants). Would it make sense to hold this PR until that is resolved?

I'd suggest we go ahead with this as-is given this change makes sense by itself - the current PR for 308 doesn't currently touch the URN values.

jogu (approved):

c2bo (approved):

PR #435, opened 2025-02-24
Improve text on effective Client Identifiers

### LINKED ISSUE #399: text about effective client id in DC API can probably be improved
https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#dc_api_request says:

Image

It's not completely clear that the two highlighted sentences apply only in the case of unsigned requests. I think we should consider moving this whole paragraph to the unsigned request section.

https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#appendix-B.3.4.1 says:

Image

I am not sure it is good to use "effective Client Identifier" here. Effective Client Identifier is only defined in the context of unsigned requests, so written like this is it possible for people to get confused as to what client id the wallet should use in signed requests. Possibly just removing 'effective' would solve it.

Mentioned by Martijn during the ISO mDL meeting today.

My understanding of the working group position is that the client id for signed requests is always the `client_id` parameter found in the request object. (With a possible exception for cases where the signature can't be validated depending on what happens with #395)

c2bo:
Closes #399 

I increased the scope of the PR slightly to also change the text about the web-origin client id scheme which doesn't seem to be correct right now (see comment in #399). Happy to remove if we don't agree on that part.


openid-4-verifiable-presentations-1_0.md
```diff
@@ -574,7 +574,7 @@ Body
...
 
-* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme if the request is not sent via the Digital Credentials API defined in (#dc_api).
+* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request) and MUST only be used if the request is sent as an unsigned request via the Digital Credentials API defined in (#dc_api). The Wallet MUST NOT accept this Client Identifier Scheme in a Request, but the Client Identifier using this Client Identifier Scheme is constructed by the Wallet based on the Origin provided by the platform as defined in (#unsigned_request).
```
Sakurann, line 577:
```suggestion
* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request) and MUST only be used if the request is sent as an unsigned request via the Digital Credentials API defined in (#dc_api). The Wallet MUST construct the Client Identifier based on the Origin provided by the platform as defined in (#unsigned_request).
```
I think "The Wallet MUST NOT accept this Client Identifier Scheme in a Request" is misleading because client_id must not be present in the unsigned dc api requests

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2055,8 +2055,9 @@ Any OpenID4VP request compliant to this section of this specification can be use
 
 ### Unsigned Request {#unsigned_request}
 
-The Verifier MAY send all the OpenID4VP request parameters as members in the request member passed to the API. In this case, the Wallet will use the Verifier's Origin to determine the Verifier's effective Client Identifier.
-
+The Verifier MAY send all the OpenID4VP request parameters as members in the request member passed to the API. In this case, the Wallet will use the plattform-provided Origin of the Verifier to determine the Verifier's authenticated Client Identifier.
+This Client Identifier is composed of a synthetic Client Identifier Scheme of `web-origin` and the Origin itself. For example, an Origin of `https://verifier.example.com` would result in a Client Identifier of `web-origin:https://verifier.example.com`.
```
Sakurann, line 2059:
```suggestion
```
I think this belongs in the definition of web-origin client identifier scheme....

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2447,10 +2448,8 @@ A non-normative example of the Authorization Response would look the same as in
 
 The following requirements apply to the `nonce` and `aud` claims in the Key Binding JWT:
 
-- the `nonce` claim MUST be the value of `nonce` from the Authorization Request;
-- the `aud` claim MUST be the value of the Client Identifier; 
-
-Note that for an unsigned Authorization Request over the DC API, the `client_id` parameter is not used. Instead, the effective Client Identifier is derived from the Origin, as described in (#dc_api_request).
```
Sakurann, line 2453: I would keep below clarification.  it feels important to repeat that client_id is omitted in unsigned dc api requests
> Note that for an unsigned Authorization Request over the DC API, the `client_id` parameter is not used.

c2bo: I was hoping that the addition of 2059 would make that unnecessary. I would like to make it clear that for DC API unsigned, the client_id is constructed by the wallet and not have to repeat that for every credential format

c2bo:
I guess with the resolution to remove client id for dc api, we can close this?

Sakurann:
superceded by #448

PR #448, opened 2025-03-06
DC API: Always use Origin for binding response

### LINKED ISSUE #351: rename client_id_scheme web-origin?
Do we need to call out these can be web origins or native app origins?

_Originally posted by @leecam in https://github.com/openid/OpenID4VP/pull/337#discussion_r1850770883_
            
current agreement seems to be no need to differentiate but opening an issue just in case.

### LINKED ISSUE #395: Clarify that signed requests for DC API should be rejected if the signature can't be verfied

jogu:
Implements option 5 from https://github.com/openid/OpenID4VP/issues/395#issuecomment-2698869922

The means that, for example, the aud value in the SD-JWT key binding is based on the Origin obtained for the platform, even in the case of signed requests.

The Client Identified Scheme prefix on the value is retained in these cases, to avoid any potential mixup attack between these DC API responses and non-DC API responses to clients using OpenID Federation.

The Client Identifier Scheme name is changed from 'web-origin' to just 'origin' as it may contain a native app identifier, as per #351.

The "client_id" is removed from the mdoc OpenID4VPDCAPIHandover as it now serves no clear purpose.

It is left unsaid whether the Wallet is required to validate the signature on signed requests in all cases, this is a policy decision for Wallets. This change allows Wallets to skip the validation but still return a response that the Verifier can successfully process, as there is no longer any uncertaintly over what would be used for `aud` in SD-JWT KB JWT or for `client_id` in the ISO mdoc OpenID4VPDCAPIHandover structure.

closes #351
closes #395


openid-4-verifiable-presentations-1_0.md
```diff
@@ -574,7 +574,7 @@ Body
...
 
-* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme if the request is not sent via the Digital Credentials API defined in (#dc_api).
+* `origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme in requests - it is only used in responses.
```
c2bo, line 577: Shouldn't we remove this whole section?
```suggestion
```

c2bo, line 577: Shouldn't we remove this whole section? This is not about a client_id anymore - we are just changing what to put into aud for DC API?
```suggestion
```

jogu: @c2bo I did wonder about that, but it's defined that what goes into `aud` is the client_id, and we're back to the problem of putting something into client_id that I think we need to be clear if it was intended for a client that authenticated using federation (+ custom url schemes) or one authenticated using just web PKI (+ DC API), for the same reasons we originally put the client id scheme into the `aud` - there's no other way for clients to tell the difference between those two types of responses.

Or to put another way, we are defining a meaning for 'origin' client id scheme in the response, so it's important that people don't define an `origin` client id scheme that means something different, hence retaining the definition in the client id scheme section as it would eventually need to be registered in any future IANA registry of client id schemes.

If Federation was always prefixed (with something other than "https://") then I think using just the url would make sense.

tlodderstedt: The alternative to say that in OID4VC over DC API, the audience of the credential presentation is the origin of the request and not the client identifier. That would require us to clearly split the spec in the "traditional" and the "DC" part including all security considerations.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2091,10 +2090,12 @@ This is an example of the payload of a signed OpenID4VP request used with the DC
 
 The signed request allows the Wallet to authenticate the Verifier using a trust framework other than the Web PKI utilized by the browser. An example of such a trust framework is the Verifier (RP) management infrastructure set up in the context of the eIDAS regulation in the European Union, in which case, the Wallet can no longer rely only on the Origin of the Verifier. This Origin MAY still be used to further strengthen the security of the flow. The external trust framework could, for example, map the Client Identifier to registered Origins.
 
-## Response
+## Response {#dc_api_response}
 
 Every OpenID4VP Authorization Request results in a response being provided through the Digital Credentials API (DC API). The response is an instance of the `DigitalCredential` interface, as defined in [@!W3C.Digital_Credentials_API], and the OpenID4VP Authorization Response parameters as defined for the Response Type are represented as an object within the `data` attribute.
 
+In places where the Client Identifier would be used in the response (for example, the `aud` value in a Key Binding JWT), the Origin MUST be used instead, prefixed with `origin:`.
```
tlodderstedt, line 2097: This should go into a general description of the DC API profile. Here I would mention that client ids with authentication credentials can go into the request, but only for authentication/authorization purposes. The audience of the credential presentation is not affected by any of those client ids.

openid-4-verifiable-presentations-1_0.md
c2bo: That feels like the better/cleaner approach to me, but a pretty big change - especially given the current timelines

jogu: @tlodderstedt 
> The alternative to say that in OID4VC over DC API, the audience of the credential presentation is the origin of the request and not the client identifier. That would require us to clearly split the spec in the "traditional" and the "DC" part including all security considerations.

I don't think we  really have a defined concept of the "audience of the credential presentation" in our credential format profiles though - we'd have to either introduce that concept, or have separate text for "traditional" versus "DC" in the credential format profiles too.

Sakurann: > I don't think we really have a defined concept of the "audience of the credential presentation"

why? we do use a term `audience (the Client Identifier of the Verifier)` in the spec already.

> That feels like the better/cleaner approach to me, but a pretty big change - especially given the current timelines

as an option, we could merge this PR and do this as part of a PR addressing #453 ? (cc @awoie )

Sakurann: so basically, we want the aud value to be prefixed with `origin:`, right? (to basically prevent the attacks that made us turn client id scheme parameter into a prefix in the first place). in which case, i agree it makes sense to keep this, but I would like to see more explanation that just `The Wallet MUST NOT accept this Client Identifier Scheme in requests - it is only used in responses.` something like 

> The Wallet MUST NOT accept this Client Identifier Scheme in requests - it is only used in responses.  `aud` claims of the Verifiable Presentation MUST be the origin value prefixed with `origin:`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2052,7 +2052,7 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
 * `transaction_data`
 * `dcql_query`
 
-The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the effective Client Identifier from the Origin. The effective Client Identifier is composed of a synthetic Client Identifier Scheme of `web-origin` and the Origin itself. For example, an Origin of `https://verifier.example.com` would result in an effective Client Identifier of `web-origin:https://verifier.example.com`. The transport of the request and Origin to the Wallet is platform-specific and is out of scope of OpenID4VP over the W3C Digital Credentials API. The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
+The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request. The security properties that are normally provided by the Client Identifier are achieved by binding the response to the Origin it was received from. The transport of the request and Origin to the Wallet is platform-specific and is out of scope of OpenID4VP over the W3C Digital Credentials API.
```
Sakurann, line 2055: Didn't we agree that client_id is omitted from signed requests too? so `client_id` needs to be removed from line 2045 too

```suggestion
The Wallet MUST ignore any `client_id` parameter that is present in either signed and unsigned requests. The security properties that are normally provided by the Client Identifier are achieved by binding the response to the Origin it was received from. The transport of the request and Origin to the Wallet is platform-specific and is out of scope of OpenID4VP over the W3C Digital Credentials API.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: agreed. made suggestion above

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2836,6 +2831,7 @@ The technology described in this specification was made available from contribut
    
    * clarify DCQL case of `claims` and `claim_sets` being absent
    * add language on client ID and nonce binding for ISO mdocs and W3C VCs
+   * for DC API, always use Origin for binding the response (e.g. in Key Binding JWT `aud`)
```
Sakurann, line 2834:
```suggestion
   * for DC API, omit client_id from the request and always use Origin for binding the response (e.g. in Key Binding JWT `aud` and sessionTranscript in mdoc)
```

Sakurann (changes requested):

openid-4-verifiable-presentations-1_0.md
c2bo: client_id is omitted for the response, not the request for signed requests

openid-4-verifiable-presentations-1_0.md
jogu: As noted somewhere else, `client_id` is still in the request - committing the other part of your suggestion though:

```suggestion
   * for DC API, always use Origin for binding the response (e.g. in Key Binding JWT `aud` and sessionTranscript in mdoc)
```

openid-4-verifiable-presentations-1_0.md
jogu: Agree with Christian, we didn't agree to omit from signed requests, and client_id in signed requests is required to know how to verify the signature (it tells you how to find the key etc) so we can't omit it.

tlodderstedt: exactly. We don't want to change the signed request payload (as it governs the handling of the client id scheme specific credential). We drop the idea to use the client id as audience in the credential presentation and use the origin instead.

openid-4-verifiable-presentations-1_0.md
jogu: I've added a sentence of explanation like @Sakurann suggested.

openid-4-verifiable-presentations-1_0.md
jogu: I've written this text and added an explicit sentence about explaining the audience isn't affected by the client id in the request.

openid-4-verifiable-presentations-1_0.md
Sakurann: ah ok, got it - thanks for clarifying

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2091,10 +2092,14 @@ This is an example of the payload of a signed OpenID4VP request used with the DC
...
 
-## Response
+## Response {#dc_api_response}
 
 Every OpenID4VP Authorization Request results in a response being provided through the Digital Credentials API (DC API). The response is an instance of the `DigitalCredential` interface, as defined in [@!W3C.Digital_Credentials_API], and the OpenID4VP Authorization Response parameters as defined for the Response Type are represented as an object within the `data` attribute.
 
+The security properties that are normally provided by the Client Identifier are achieved by binding the response to the Origin it was received from.
+
+The audience for the response (for example, the `aud` value in a Key Binding JWT) MUST be the Origin, prefixed with `origin:`. This is the case even for signed requests - when using OpenID4VP over the DC API the Client Identifier is never used as the audience for the response.
```
Sakurann, line 2101:
```suggestion
The audience for the response (for example, the `aud` value in a Key Binding JWT) MUST be the Origin, prefixed with `origin:`, for example `origin:https://verifier.example.com/`. This is the case even for signed requests. Therefore, when using OpenID4VP over the DC API, the Client Identifier is not used as the audience for the response.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -574,7 +574,7 @@ Body
...
 
-* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme if the request is not sent via the Digital Credentials API defined in (#dc_api).
+* `origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme in requests. In OpenID4VP over the Digital Credentials API the audience of the Credential Presentation is the origin value prefixed by `origin:`, for example `origin:https://verifier.example.com/`.
```
Sakurann, line 577:
```suggestion
* `origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme in requests. In OpenID4VP over the Digital Credentials API, the audience of the Credential Presentation is always the origin value prefixed by `origin:`, for example `origin:https://verifier.example.com/`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2136,7 +2141,7 @@ This `presentation_definition` parameter contains a single `input_descriptor` el
 The following requirements apply to the `nonce` and `aud` claims of the Verifiable Presentation:
 
 - the `nonce` claim MUST be the value of `nonce` from the Authorization Request;
-- the `aud` claim MUST be the value of the Client Identifier; or effective Client Identifier for an unsigned Authorization Request over the DC API, as described in (#dc_api_request).
+- the `aud` claim MUST be the value of the Client Identifier; or for a request over the DC API MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```
Sakurann, line 2144:
```suggestion
- the `aud` claim MUST be the value of the Client Identifier. For a request over the DC API, it MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2181,7 +2186,7 @@ This `presentation_definition` parameter contains a single `input_descriptor` el
 The following requirements apply to the `challenge` and `domain` claims within the `proof` object in the Verifiable Presentation:
 
 - the `challenge` claim MUST be the value of `nonce` from the Authorization Request;
-- the `domain` claim MUST be the value of the Client Identifier; or effective Client Identifier for an unsigned Authorization Request over the DC API, as described in (#dc_api_request).
+- the `domain` claim MUST be the value of the Client Identifier; or for a request over the DC API MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```
Sakurann, line 2189:
```suggestion
- the `domain` claim MUST be the value of the Client Identifier. For a request over the DC API, it MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2463,9 +2464,7 @@ A non-normative example of the Authorization Response would look the same as in
 The following requirements apply to the `nonce` and `aud` claims in the Key Binding JWT:
 
 - the `nonce` claim MUST be the value of `nonce` from the Authorization Request;
-- the `aud` claim MUST be the value of the Client Identifier; 
-
-Note that for an unsigned Authorization Request over the DC API, the `client_id` parameter is not used. Instead, the effective Client Identifier is derived from the Origin, as described in (#dc_api_request).
+- the `aud` claim MUST be the value of the Client Identifier; or for a request over the DC API MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```
Sakurann, line 2467:
```suggestion
- the `aud` claim MUST be the value of the Client Identifier; For a request over the DC API, it MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2052,14 +2052,16 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
 * `transaction_data`
 * `dcql_query`
 
-The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the effective Client Identifier from the Origin. The effective Client Identifier is composed of a synthetic Client Identifier Scheme of `web-origin` and the Origin itself. For example, an Origin of `https://verifier.example.com` would result in an effective Client Identifier of `web-origin:https://verifier.example.com`. The transport of the request and Origin to the Wallet is platform-specific and is out of scope of OpenID4VP over the W3C Digital Credentials API. The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
+The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
```
Sakurann, line 2055:
```suggestion
The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.

In signed requests, `client_id` parameter MUST be present, as it governs Client Identifier Scheme specific handling of the request.
```
 this one sentence would prevent confusion i had...

jogu: Edited slightly for consistency with the sentence about unsigned and to try and be clearer on the reason:

```suggestion
The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.

The `client_id` parameter MUST be present in signed requests defined in (#signed_request), as it communicates to the wallet which Client Identifier Scheme and Client Identifier to use when verifying the request signature or retrieving client metadata.
```

openid-4-verifiable-presentations-1_0.md
jogu: I think we do need the "or" or some other way to make clearer that the first sentence doesn't always apply:

```suggestion
- the `aud` claim MUST be the value of the Client Identifier, except for requests over the DC API where it MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```

openid-4-verifiable-presentations-1_0.md
jogu:
```suggestion
- the `aud` claim MUST be the value of the Client Identifier, except for requests over the DC API where it MUST be the Origin prefixed with `origin:`, as described in (#dc_api_response).
```

jogu:
@Sakurann 

> looks good - editorial suggestions. assuming majority of them would be incorporated

Thank you! I've committed all your suggestions, a few I tweaked a little first.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -574,7 +574,7 @@ Body
...
 
-* `web-origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme if the request is not sent via the Digital Credentials API defined in (#dc_api).
+* `origin`: This Client Identifier Scheme is defined in (#dc_api_request). The Wallet MUST NOT accept this Client Identifier Scheme in requests. In OpenID4VP over the Digital Credentials API, the audience of the Credential Presentation is always the origin value prefixed by `origin:`, for example `origin:https://verifier.example.com/`.
```
tlodderstedt, line 577: Do we still need this scheme?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2052,14 +2052,18 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
 * `transaction_data`
 * `dcql_query`
 
-The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet determines the effective Client Identifier from the Origin. The effective Client Identifier is composed of a synthetic Client Identifier Scheme of `web-origin` and the Origin itself. For example, an Origin of `https://verifier.example.com` would result in an effective Client Identifier of `web-origin:https://verifier.example.com`. The transport of the request and Origin to the Wallet is platform-specific and is out of scope of OpenID4VP over the W3C Digital Credentials API. The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
+The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
+
+The `client_id` parameter MUST be present in signed requests defined in (#signed_request), as it communicates to the wallet which Client Identifier Scheme and Client Identifier to use when verifying the request signature or retrieving client metadata.
```
tlodderstedt, line 2057:
```suggestion
The `client_id` parameter MUST be present in signed requests defined in (#signed_request), as it communicates to the wallet which Client Identifier Scheme and Client Identifier to use when authenticating the client through verification of the request signature or retrieving client metadata.
```

tlodderstedt (approved):

openid-4-verifiable-presentations-1_0.md
c2bo: I brought up the same question before (https://github.com/openid/OpenID4VP/pull/448#discussion_r1984999561) and this seems to be more of a placeholder so nobody use that as a scheme. It is technically not a real client id scheme (cannot be used in the request), but blocked for the response.

I don't really like it but I guess this is fine?

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
jogu: Thanks Torsten - applied that change!

openid-4-verifiable-presentations-1_0.md
jogu: Yes, what Christian said - client id schemes are basically defining prefixes to be used in two places:

1. The client id parameter
2. The `aud` value in responses

The `origin:` prefix is definitely used in "2" so I think to avoid any potential for future confusion it should be a "registered" client id scheme.

hlozi (approved):

GarethCOliver (approved):

Issue #420, opened 2025-02-10
Value matching clarifications


davidz25:
I'm implementing DCQL right now (draft 24) and it's not clear to me from the spec how to implement value matching. Some observations

- The standard says about `values` that it's "An array of strings, integers or boolean values that specifies the expected values of the claim." ... however it goes together with a path so the value to compare against might indeed be an object (e.g. `"path": ["address"]`) where `address` is an object with `street_address`, `postal_code` etc elements. Why don't we allow `values` to also include objects?

- For ISO mdoc, it's not at all clear how to match. Yes, trivial cases where the data element is a `tstr` or boolean (e.g. major type 7) or integers (major type 0 and 1)... but what about things like `full-date` (which according to RFC 8943 is a string with tag 1004), for example, should we automatically drop all tags? In that case you can use e.g. `values = ["2025-02-11"]` to match a `full-date`. I think we should have concise language for the implementer here and not have them guess what the intent is.

c2bo:
The whole topic of value matching is still something that will need a bit more work imho and where there is ongoing discussion in the WG. The initial proposal by @danielfett also included more advanced options like applying transformations to claims (similar to https://openid.bitbucket.io/ekyc/openid-connect-advanced-syntax-for-claims.html#name-defining-transformed-claims).

I agree that we probably need to be more explicit on data types and especially how matching works for formats not based on JSON serialization.

On-going discussion on value matching: #307

charsleysa:
In regards to the address example, is it not valid to compare against one of the object properties?

e.g. the "address" claim as a whole is selectively discloseable, but not the individual properties
```
{
    "address": {
        "street_address": "123 street",
        "postal_code": "1234"
    }
}
```
```
"claims": [
    { "path": ["address","street_address"], "value": "123 street" }
]
```
this would successfully match and return the entire "address" claim because it can only be disclosed as a whole

Sakurann:
PR #538 merged

PR #452, opened 2025-03-10
Value Matching

### LINKED ISSUE #307: Usecases for value matching in DCQL
Issue to track use-cases for value matching in DCQL.

The payment use-cases requires value matching for performing SCA. In this scenario a webpage would like to confirm the user holds a certain credit card. Here the RP would pass the credit card number as a value that needs to be matched, to filter the results to that given card.

c2bo:
Closes #307 

I tried to clarify a bit more on the results of the discussion in the issues, but it felt like it didn't need any significant changes - I guess the more important part would be the clarification on how to apply an credential formats not based on json (#420)?

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -752,7 +752,9 @@ within the Verifiable Credential, as defined in (#claims_path_pointer).
 `values`:
 : OPTIONAL. An array of strings, integers or boolean values that specifies the expected values of the claim.
 If the `values` property is present, the Wallet SHOULD return the claim only if the
-type and value of the claim both match for at least one of the elements in the array. Details of the processing rules are defined in (#selecting_claims).
+type and value of the claim both match for at least one of the elements in the array. Details of the processing
+rules are defined in (#selecting_claims). Note that the expected behavior for value matching is only strict matching
+of the values, not any sort of fuzzy matching.
 
 ### Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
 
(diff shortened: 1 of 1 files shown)


Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -850,7 +850,9 @@ within the Verifiable Credential, as defined in (#claims_path_pointer).
 `values`:
 : OPTIONAL. An array of strings, integers or boolean values that specifies the expected values of the claim.
 If the `values` property is present, the Wallet SHOULD return the claim only if the
-type and value of the claim both match for at least one of the elements in the array. Details of the processing rules are defined in (#selecting_claims).
+type and value of the claim both match for at least one of the elements in the array. Details of the processing
+rules are defined in (#selecting_claims). Note that the expected behavior for value matching is only strict matching
+of the values, not any sort of fuzzy matching.
```
tplooker, line 855: Im not sure whether we need this addition here, its pretty clear from the first sentence that the matching is exact.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -879,9 +881,10 @@ elements in `values` i.e., the claim should be treated the same as if it did not
 exist in the Credential. Implementing this restriction may not be possible in
 all cases, for example, if the Wallet does not have access to the claim value
 before presentation or user consent or if another component routing
-the request to the Wallet does not have access to the claim value. Therefore, Verifiers
-must treat restrictions expressed using `values` as a best-effort way to improve
-user privacy, but MUST NOT rely on it for security checks.
+the request to the Wallet does not have access to the claim value. It might also
+come down to a decision of the Wallet and/or the user if the value matching request
+is followed. Therefore, Verifiers must treat restrictions expressed using `values` as a
+best-effort way to improve user privacy, but MUST NOT rely on it for security checks.
```
awoie, line 887: it may ultimately be up to the Wallet and/or the user to decide whether th
```suggestion
the request to the Wallet does not have access to the claim value. It is ultimately up to the
Wallet and/or the End-user if the value matching request
is followed. Therefore, Verifiers MUST treat restrictions expressed using `values` as a
best-effort way to improve user privacy, but MUST NOT rely on it for security checks.
```

awoie (approved):
lgtm, some minor editorial suggestions.

openid-4-verifiable-presentations-1_0.md
paulbastian, line 855:
```suggestion
type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
rules are defined in (#selecting_claims).
```

leecam: Yeah agree, not sure this is needed. Paul's suggestion looks good

leecam: Yeah I prefer this.

danielfett (approved):

paulbastian (approved):

PR #456, opened 2025-03-13
Additional reasons why Requests using Redirect URI Scheme cannot be signed

peppelinux:
This PR resolves https://github.com/openid/OpenID4VP/issues/33, removing the requirement to not sign the authz request when the redirect_uri scheme is used.

Nothing should prevent implementations to sign the authz request when the trust framework in use or their will wants this.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2834,7 +2834,7 @@ The technology described in this specification was made available from contribut
 
    -25
    
-   * redirect_uri scheme can be signed
+   * authz requests using redirect_uri scheme can be signed
    * clarify DCQL case of `claims` and `claim_sets` being absent
    * add language on client ID and nonce binding for ISO mdocs and W3C VCs
    * clarify the behavior is not to sign when authorization_signed_response_alg is omitted
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2834,6 +2834,7 @@ The technology described in this specification was made available from contribut
 
    -25
    
+   * redirect_uri scheme can be signed
    * clarify DCQL case of `claims` and `claim_sets` being absent
    * add language on client ID and nonce binding for ISO mdocs and W3C VCs
    * clarify the behavior is not to sign when authorization_signed_response_alg is omitted
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -538,9 +538,9 @@ Confusing Verifiers using a Client Identifier Scheme with those using none can l
 
 This specification defines the following Client Identifier Schemes, followed by the examples where applicable: 
 
-* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). Example Client Identifier value is `redirect_uri:https://client.example.org/cb`.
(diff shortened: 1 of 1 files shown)


jogu (changes requested):
I don't see any consensus on the issue for making this change, and I don't understand this change.

redirect_uri does not define any mechanism for communicating the signing key to the wallet. The wallet hence can't verify a signed request.

If the wallet is going to obtain keys from federation, the federation client id scheme would be used. If the wallet is going to obtain keys from somewhere else, a different client id scheme (not redirect_uri) should be used. There's no  defined nor interoperable way of obtaining keys for the redirect_uri client id scheme. The redirect_uri client id scheme isn't really intended for any production usage.

peppelinux:
@jogu I can therefore revert the changes and use this pr to describe the reasons why the redirect uri scheme cannot be signed, replacing the normative MUST with a simple "cannot"

Wdyt?

jogu:
> @jogu I can therefore revert the changes and use this pr to describe the reasons why the redirect uri scheme cannot be signed, replacing the normative MUST with a simple "cannot"

That seems fine to me. I'd suggest clarifying that the "cannot" with "as this client id scheme defines no way for a trusted key to be obtained by the wallet to verify the request" or something like that.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -538,9 +538,9 @@ Confusing Verifiers using a Client Identifier Scheme with those using none can l
...
 
-* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Authorization Request MUST NOT be signed. The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). Example Client Identifier value is `redirect_uri:https://client.example.org/cb`.
+* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). Example Client Identifier value is `redirect_uri:https://client.example.org/cb`.
```
peppelinux, line 541:
```suggestion
* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). An example Client Identifier value is `redirect_uri:https://client.example.org/cb`. Requests using the `redirect_uri` Client Identifier cannot be signed because there is no method for the Wallet to obtain a trusted key for verification. This scheme lacks a mechanism to communicate the signing key, rendering it unsuitable for signed requests. Therefore, implementations requiring signed requests MUST NOT use the `redirect_uri` Client ID scheme. Due to the absence of a defined method for key management, the `redirect_uri` scheme is not intended for production use.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -538,9 +538,9 @@ Confusing Verifiers using a Client Identifier Scheme with those using none can l
...
 
-The following is a non-normative example of a request with this Client Identifier Scheme:
+The following is a non-normative example of an unsigned request with this Client Identifier Scheme:
```
peppelinux, line 543:
```suggestion
The following is a non-normative example of an unsigned request with the `redirect_uri` Client Identifier Scheme:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2956,7 +2956,8 @@ The technology described in this specification was made available from contribut
    [[ To be removed from the final specification ]]
 
    -25
-
+   
+   * authz requests using redirect_uri scheme can be signed
```
peppelinux, line 2960:
```suggestion
   * additional reasons why requests using redirect_uri scheme cannot be signed
```

jogu:
```suggestion
   * clarify why requests using redirect_uri scheme cannot be signed
```

openid-4-verifiable-presentations-1_0.md
jogu: I think we can probably say the same thing with fewer words and fewer words is generally good :) 

```suggestion
* `redirect_uri`: This value indicates that the Client Identifier (without the prefix `redirect_uri:`) is the Verifier's Redirect URI (or Response URI when Response Mode `direct_post` is used). The Verifier MAY omit the `redirect_uri` Authorization Request parameter (or `response_uri` when Response Mode `direct_post` is used). All Verifier metadata parameters MUST be passed using the `client_metadata` parameter defined in (#new_parameters). An example Client Identifier value is `redirect_uri:https://client.example.org/cb`. Requests using the `redirect_uri` Client Identifier Scheme cannot be signed because there is no method for the Wallet to obtain a trusted key for verification. Therefore, implementations requiring signed requests cannot use the `redirect_uri` Client ID scheme.
```

jogu (approved):

Sakurann (approved):

Issue #436, opened 2025-02-24
Is OpenID Federation supported for DC API requests?


TimoGlastra:
The spec mentions for OpenID Federation client id scheme, that the request MAY contain a `trust_chain` parameter. 

However Appendix A.2. does not include the `trust_chain` parameter in the supported parameters. 

Should Appendix A.2. also include the `trust_chain` parameter?

Sakurann:
signed openid4vp request over DC API works with any client_id_scheme :)

PR #458, opened 2025-03-15
add small note about client id parameters for dc api

TimoGlastra:
Fixes https://github.com/openid/OpenID4VP/issues/436

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2245,6 +2245,8 @@ Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_toke
 
 The `client_id` parameter MUST be omitted in unsigned requests defined in (#unsigned_request). The Wallet MUST ignore any `client_id` parameter that is present in an unsigned request.
 
+Parameters defined by a specific client identifier scheme (such as the `trust_chain` parameter for the OpenID Federation client id scheme) are also supported over the W3C Digital Credentials API.
+
 The `client_id` parameter MUST be present in signed requests defined in (#signed_request), as it communicates to the wallet which Client Identifier Scheme and Client Identifier to use when authenticating the client through verification of the request signature or retrieving client metadata.
 
 The value of the `response_mode` parameter MUST be `dc_api` when the response is neither signed nor encrypted and `dc_api.jwt` when the response is signed and/or encrypted as defined in (#jarm).


TimoGlastra:
Some notes I also shared with @peppelinux over email:
> After re-reading the section on DC API, I must say that there is already a line indicating the set of parameters can be extended: "Additional request parameters MAY be defined and used with OpenID4VP over the DC API.".
>
> So it may not be needed to merge the clarification, although me and another engineer on our team both were in doubt about the restrictions from the supported parameters.

jogu:
> Some notes I also shared with @peppelinux over email:
> 
> > After re-reading the section on DC API, I must say that there is already a line indicating the set of parameters can be extended: "Additional request parameters MAY be defined and used with OpenID4VP over the DC API.".
> > So it may not be needed to merge the clarification, although me and another engineer on our team both were in doubt about the restrictions from the supported parameters.

The list of parameters that can be used in the DC API defined in [section A.2](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-A.2) is explicitly prefixed with:

> Out of the Authorization Request parameters defined in [@!RFC6749] and (#vp_token_request), the following are supported with OpenID4VP over the W3C Digital Credentials API

I'm not actually sure how to interpret  that text, given that `trust_chain` is defined in OID Federation.

The intent of all this text was to make sure that not all the request parameters from https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#parameters automatically became allowable to use in the DC API. For example, `state` and `code_challenge`  are deliberately excluded from the list as they serves no purpose in the DC API.

So likely your proposed text makes sense as I can't currently think of a case where a client id scheme might define a parameter that doesn't make sense in the DC API.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2181,7 +2181,7 @@ In addition to the above-mentioned parameters, a new parameter is introduced for
 
 * `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). An array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet can detect replay of the request from a malicious Verifier by comparing values in this parameter to the Origin. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
 
-Additional request parameters MAY be defined and used with OpenID4VP over the DC API.
+Additional request parameters MAY be defined and used with OpenID4VP over the DC API, including parameters defined by a specific client identifier scheme (such as the `trust_chain` parameter for the `https` client id scheme).
```
jogu, line 2184: I think it actually makes sense to add this note further up - either before line 2176 or before 2180? And word it more like:

> Parameters defined by a specific client identifier scheme (such as the `trust_chain` parameter for the OpenID Federation client id scheme) are also supported over the W3C Digital Credentials API.

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
c2bo: I think 2176 would be good, directly after the list of parameters

c2bo (approved):
I think it would be good to move the whole sentence up

TimoGlastra:
Okay thanks for the feedback, i will update the PR

jogu (approved):

Issue #329, opened 2024-11-19
Mitigating a same-device / cross-device security issue


patatoid:
There are security issues with same-device and cross-device flows where as stated in oid4vc_sectrust P-40 (https://openid.github.io/OpenID4VC_SecTrust/draft-oid4vc-security-and-trust.html#name-security-requirement-p-40) an attacker might share a QR code or a link to a victim that would authenticate for him.

The issue is said to be mitigated via the browser API (https://digitalcredentials.dev/docs/wallets/android/#the-provider-api) giving the ability to link a user session to a nonce in case of same-device but also with cross-device by enabling the different devices to communicate one with the other. (@jogu tell me if I am correct on this)

An other way to mitigate the issue would be to provide a short time to live for QR codes and links using the resource created for handling direct_post requests. During the direct_post requests the verifier would be enforced to check the validity of the request by verifying the expiration set on the before step creating the QR code or the link.

The solution would help in both ways, using browser API or not, preventing the authorization request share by setting a short expiration to the direct_post request associated with the verifiable credential demand.

There is a working version of verifier enabling this with a little updates, here is the associated commit https://github.com/malach-it/boruta_auth/commit/a104d2e5f77e646d8338884669df88b2648abecf

danielfett:
Unfortunately, short-lived codes/URLs is an [insufficient mitigation](https://www.ietf.org/archive/id/draft-ietf-oauth-cross-device-security-08.html#name-short-lived-timebound-qr-or) that can easily be circumvented by simple adaptations by the attacker. For example, an attacker might provide a static QR code to his website were then a redirect to a fresh auth request URL is performed. Or the QR code can be updated on a website.

patatoid:
For now, it is the only one I see since even using browser API the devices have to be interconnected somehow (Bluetooth or other means) which is not the case for most of the users setups.

Would you see other ways to mitigate this?

danielfett:
I don't know exactly how the browser API will handle this, but ideally, no persistent connection between the devices is needed - just a proof that they are in proximity.

In the same-device flow, the problem can be mitigated with existing protocol features (essentially checking that the nonce matches the session that was started in the browser) as described [here](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-13.2).

For the cross-device flow, I don't see any mitigation other than OS/browser support for a secure cross-device  transfer that includes an origin check.

patatoid:
Do you know how current cross-device implementations solve the issue?

[...]

PR #467, opened 2025-03-22
Authorization link share Security Consideration

patatoid:
Following the discussions on #329, here is a suggestion of addition in the Security Considerations section in accordance with the discussed facts.

Hoping the formulation is clear enough.

Being my first contribution, I am not sure about the requirements. If there are any I missed, do not hesitate to tell.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1836,6 +1836,18 @@ The OpenID Foundation provides tools that can be used to confirm that an impleme
 
 https://openid.net/certification/conformance-testing-for-openid-for-verifiable-presentations/
 
+## Authorization link share
+
+An attacker could perform an authorization request and share the resulting link to a victim to perform its presentation flow if not mitigated. The proximity between the verifier and the holder is a fundamental security factor for a presentation, it proves the response is tied to the presentation request and is associated to it, verifiers and wallets being compliant with the protocol. That proximity enabled by the link between the holder delegated by his wallet and the verifier can be enforced using security practices.
+
+### Using web protocols
+
+Using the `direct_post` Response Mode allows the wallet to reply back to the verfier through the `response_uri` endpoint, the verifier can then initiate a presentation session. A short time to live for that session can minimize the attack surface reducing the attack frame within malicious actors can act. Preventing replay attacks can be done by authorizing a single presentation within a session. Those mitigations would act on both same and cross device flows.
+
+### Digital Credentials API
+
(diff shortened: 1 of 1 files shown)


Sakurann (changes requested):
how does this add to what is already documented in https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-14.2 and https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-14.3

we should probably add how DC API prevents session fixation, but I would expect more details than is currently in the PR (cc @danielfett )

danielfett:
I agree with @Sakurann. We should add something on DC API and session fixation, but any solution that is based on short-lived codes/URLs doesn't solve the problem (see my latest comment on https://github.com/openid/OpenID4VP/issues/329#issuecomment-2750597422).

patatoid:
The mitigated issue here would concern verifiers using presentations as a way of authentication. It is a session fixation as described in the cited section but the attack would be different.

@tlodderstedt do you think it is possible to improve the Session Fixation section to add a paragraph about such attacks.

@danielfett as discussed in #329 I think DC API is a long term solution not applicable in a middle term since devices OS are to be updated. The suggested mitigations are practical given the tools available today.

danielfett:
> @danielfett as discussed in #329 I think DC API is a long term solution not applicable in a middle term since devices OS are to be updated. The suggested mitigations are practical given the tools available today.

The problem is, that this is not a mitigation. If it works for a user, so will it work for an attacker. Since the attack is automated and happens in real-time, time limits will simply not be a mitigation. We should instead warn that the problem exists and that, without the DC API, there is no good mitigation. We could add that having a user verify the domain in the browser is closer to a practical mitigation, but will work only for some users.

patatoid:
Those way of mitigation are taken from https://github.com/danielfett/oauth-cross-device-security/blob/main/draft-ietf-oauth-cross-device-security.md#short-livedtimebound-codes, you are one of the authors. Does it changed since then?

danielfett:
That is an excellent point, actually. The language in the draft should be more strict and more precise. It's fine IMO if an implementer additionally adds, e.g., a time limit as an additional precaution (that's what the the draft tries to say). But on a protocol level, it should not be positioned as a proper mitigation.

Sakurann:
there seems to be an agreement that the cross-device BCP should use slightly different language, but we need a new PR addressing that. this is not a breaking change that is time critical for Final 1.0. closing this PR  - let's discuss on issue.

PR #468, opened 2025-03-24
slight tweak to default treatment of https URLs

aaronpk:
unambiguously defines the plain https Client Identifier URLs as a fallback to mean OpenID Federation

This is a PR against the open PR #401 

Context of this PR is documented in https://github.com/openid/OpenID4VP/pull/401#issuecomment-2747041478

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -528,7 +528,7 @@ If a `:` character is not present in the Client Identifier, the Wallet MUST trea
 
 For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
-If the Client Identifier begins with `https://`, the Wallet MUST treat the Client Identifier as an Entity Identifier defined in OpenID Federation [@!OpenID.Federation].
+If the Client Identifier begins with `https://` the full Client Identifier URL represents an Entity Identifier defined in OpenID Federation [@!OpenID.Federation] and MUST be processed as though it was prefixed with `openid_federation` as described below.
 
 If a `:` character is present in the Client Identifier but the value preceding it is not a recognized and supported Client Identifier Scheme value, the Wallet MAY treat the Client Identifier as having a default Client Identifier Scheme. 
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -582,7 +582,9 @@ Body
...
+* `openid_federation`: This value indicates that the Client Identifier is an Entity Identifier defined in OpenID Federation [@!OpenID.Federation]. Processing rules given in [@!OpenID.Federation] MUST be followed. Automatic Registration as defined in [@!OpenID.Federation] MUST be used. The Authorization Request MAY also contain a `trust_chain` parameter. The final Verifier metadata is obtained from the Trust Chain after applying the policies, according to [@!OpenID.Federation]. The `client_metadata` parameter, if present in the Authorization Request, MUST be ignored when this Client Identifier scheme is used. Example Client Identifier: `openid_federation:https://federation-verifier.example.com`.
+
+To use the Client Identifier Schemes `openid_federation`, `decentralized_identifier`, `verifier_attestation`, `x509_san_dns` and `x509_hash`, Verifiers MUST be confidential clients. This might require changes to the technical design of native apps as such apps are typically public clients.
```
Sakurann, line 587: what is the rationale behind this? We have never mandated any particular architecture, yet alone native apps being able to securely store credentials for authentication. I am extremely uncomfortable with this

```suggestion
```

Sakurann (changes requested):
looks good other than a confidential client part

openid-4-verifiable-presentations-1_0.md
```diff
@@ -528,11 +528,13 @@ If a `:` character is not present in the Client Identifier, the Wallet MUST trea
 
 For example, if an Authorization Request contains `client_id=example-client`, the Wallet would interpret the Client Identifier as referring to a pre-registered client.
 
+If the Client Identifier begins with `https://` the full Client Identifier URL represents an Entity Identifier defined in OpenID Federation [@!OpenID.Federation] and MUST be processed as though it was prefixed with `openid_federation` as described below.
```
Sakurann, line 531: > I personally don't have an issue with OpenID4VP defining the default behavior of https URLs as long as it's defined only within the context of OpenID4VP and doesn't spill out to unrelated communities and ecosystems.

if this is the case, it must be much more explicit that this https:// behavior is only for openid4vp context

jogu:
I got a request for a rendered version of this PR so uploaded html that would result from this PR here in case it's useful to others:

https://www.heenan.me.uk/~joseph/openid-4-verifiable-presentations-1_0-pr468-2025-04-26.html#section-5.10

selfissued (approved):
This improves the PR that it would be merged into by removing the breaking change.

Sakurann:
WG discussion:
- be clear that `https` is not a prefix, because it is not.
- default is `openid_federation:` prefix is used for openid federation deployments.
    - dids also get prefixed with decentralzied-identifier:
- but! if client_id does not have a prefix, and it happens to start with https://, there is one expected behavior, which is the same as `openid_federation:`
    - do we need to be explicit about this "no prefix, but it's https" scenario? if we are trying to discourage no-prefix scenario at all? if not harmful, might be better to define a behavior rather than keeping ?
- add security considerations on this
    - as long as someone is using `https` prefix, security issue that led to prefixing client id remains? vs this issue exists only when there are multiple ways to interpret, but not when there is one way to interpret https prefix.

Sakurann:
chair hat off, we should also define behavior when client_id does not have a prefix, **and does not happen to start with https://**

jogu:
Just for info for readers: there is some overlap with the federation spec owned connect wg here, and this is scheduled to be discussed on Thursday's Connect WG call, we're trying to the best of our ability to have alignment between Connect, DCP & OAuth working groups.

sloops77:
i listened to the discussion and it seems strange to have two ways of being able to utilize openid federation. What are the consequences?

- The wallets in the existing ecosystems may end up implementing the openid_federation prefix for ecosystem interop anyway. 
- For new implementors how do they navigate having two ways to implement open id fed? Should they implement the prefix or rely on the default behaviour or should they do both? The right answer will depends on their use case - but the decision can be quite nuanced and therefore its opening up room for incorrect decisions by implementors and therefore interop issues

openid-4-verifiable-presentations-1_0.md
```diff
@@ -528,11 +528,14 @@ If a `:` character is not present in the Client Identifier, the Wallet MUST trea
...
+Note that like all requirements in this specification, this requirement only applies when using this specification.
```
sloops77, line 532: this sentence is a tautology

Sakurann:
please re-review PR #401 in light of new understanding in DCP WG that is documented in https://github.com/openid/OpenID4VP/pull/401#issuecomment-2776912290

bc-pi (changes requested):
the underlying [openid:equal-treatment-under-the-prefix](https://github.com/openid/OpenID4VP/tree/equal-treatment-under-the-prefix) branch and open PR https://github.com/openid/OpenID4VP/pull/401 against which this one was created no longer has default so this has effectively been overcome by events

aaronpk:
I don't think this PR is necessary anymore since the changes to #401, as has been discussed in the WG call and in the issue comments.

PR #470, opened 2025-03-24
add verifier public key to the session transcript

### LINKED ISSUE #400: include verifier's public encryption key into SessionTranscript to prevent some possible attacks in case of unsigned requests
On today's ISO mDL WG call, Martijn asked if there was a reason why the verifier's public encryption key isn't included in the SessionTranscript.

Including it there might prevent some 'man in the browser' type attacks (or any case where a component between relying party and wallet changes the encryption key) for unsigned requests where the encryption key has been replaced by the attacker. (Martijn I believe also noted that in these kind of scenarios the attacker has already gained significant access, so preventing this exact attack may not be worthwhile. But also mentioned that including the key in the SessionTranscript might not be harmful so maybe it should be done anyway.)

awoie:
Depends on #448 (on the removal of `client_id` from `SessionTranscript`).

This PR includes the thumbprint of the public verification key of the verifier in the `SessionTranscript`.

Fixes #400


hlozi (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2522,15 +2522,15 @@ OpenID4VPDCAPIHandoverInfoBytes = bstr .cbor OpenID4VPDCAPIHandoverInfo
...
+  jwk_thumbprint
 ] ; Array containing handover parameters
 
-client_id = tstr
-
 origin = tstr
 
 nonce = tstr
+
+jwk_thumbprint = tstr
```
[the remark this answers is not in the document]
awoie: it is the intention to encode as JWK Thumbprint since the key is a JWK.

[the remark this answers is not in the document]
awoie: My apologies, you are right, it is a `bstr`.

awoie:
> Type should either be bstr if we use the hash directly, or we need to clarify that the hash is base6url encoded

changed to `bstr`.

c2bo (approved):

Sakurann (approved):

paulbastian (changes requested):
I think this is a substantial important changes, that needs further elaboration on the motivation:
- add a note under this section that this is a security measurement for unsigned requests
- please add security consideration that the RP should check that jwk_thumbprint matches its key and what to do when it does not match

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2606,6 +2609,7 @@ The `OpenID4VPDCAPIHandover` structure has the following elements:
 * The `OpenID4VPDCAPIHandoverInfo` has the following elements:
   * The first element MUST be the string representing the Origin of the request as described in (#dc_api_request). It MUST NOT be prefixed with `origin:`.
   * The second element MUST be the value of the `nonce` request parameter.
+  * For the Response Mode `dc_api.jwt`, the third element MUST be the JWK SHA-256 Thumbprint as defined in [@!RFC7638], encoded as a CBOR Byte String, of the Verifier's public key used to encrypt the response. If the Response Mode is `dc_api`, the third element MUST be `null`. For unsigned requests, including the JWK Thumbprint in the `SessionTranscript` allows the Verifier and/or the Wallet to detect whether the response was re-encrypted by a third party, potentially leading to the leakage of sensitive information. While this does not prevent such an attack, it makes it detectable and helps preserve the confidentiality of the response.
```
paulbastian, line 2612: How is it detectable by the Wallet? I was thinking it's only detectable by the RP

awoie, line 2612:
```suggestion
  * For the Response Mode `dc_api.jwt`, the third element MUST be the JWK SHA-256 Thumbprint as defined in [@!RFC7638], encoded as a CBOR Byte String, of the Verifier's public key used to encrypt the response. If the Response Mode is `dc_api`, the third element MUST be `null`. For unsigned requests, including the JWK Thumbprint in the `SessionTranscript` allows the Verifier to detect whether the response was re-encrypted by a third party, potentially leading to the leakage of sensitive information. While this does not prevent such an attack, it makes it detectable and helps preserve the confidentiality of the response.  
```

awoie: Fixed

paulbastian:
I believe a dedicated section in the security considerations would be appropriate, but that shouldn't block this from being merged.

Sakurann:
> I believe a dedicated section in the security considerations to give guidance for Verifier in these cases would be appropriate, but that shouldn't block this from being merged.

I agree.. we should be more diligent in adding security considerations when we do PRs that apply cc @danielfett

paulbastian (approved):

PR #477, opened 2025-03-26
Changing JARM references to JWT directly (Issue #463)

### LINKED ISSUE #438: Clarify key selection for client_metadata jwks
Feedback from the ISO mdl DC API interop preparation:

The way that keys should be passed in client_metadata / the way that the key in the client_metadata jwks to use for encryption is selected is unclear.

Some verifiers are including use: enc in the encryption jwk, some are not. Some wallets were requiring a key with use:enc, some were not.

I think there's perhaps a general assumption currently that a "normal" style key selection algorithm would be used, something like described in FAPI: https://openid.net/specs/openid-financial-api-part-2-1_0.html#duplicate-key-identifiers but we should be explicit about the expected behaviour.

### LINKED ISSUE #463: Changing JARM references to JWT directly
Currently, we use JARM as the means for response encryption. This works very well for signed responses, but is expressly forbids unsigned responses. This has resulted in a lot of text that says, in the case of unsigned, encrypted responses to ignore large portions of the spec. 

Currently, there doesn't seem to be well understood uses for signed responses at all. Given that, removing the reference to JARM and changing the text that we currently use to countermand JARM to instead say 'use an unsigned, encrypted JWT' would simplify life for readers/implementers substantially while being minimal/no change for existing implementations. 

Example New Text: 

section 5.1
…
Client_metadata
…
authorization_encrypted_response_alg: [OPTIONAL] The JWE [RFC7516] alg algorithm REQUIRED for encrypting authorization responses. If omitted, then no encryption is performed.
[gco.] Could replace this with jwk kid instead (as the  jwk specifies alg), which would be more explicit and help from an implementation side when jwks are passed in-band, but when it is oob this might still make sense?
authorization_encrypted_response_enc: [OPTIONAL] The JWE [RFC7516] enc algorithm REQUIRED for encrypting authorization responses. If authorization_encrypted_response_alg is specified, the default for this value is A128CBC-HS256. When authorization_encrypted_response_enc is included, authorization_encrypted_response_alg MUST also be provided.
…
8.3. Encrypted Responses
This section defines how an Authorization Response containing a VP Token can be encrypted at the application level when the Response Type value is vp_token or vp_token id_token. Encrypting the Authorization Response can prevent personal data in the Authorization Response from leaking, when the Authorization Response is returned through the front channel (e.g., the browser).
To encrypt the Authorization Response, implementations MUST use an unsigned, encrypted JWT as described in [RFC7519]. For security considerations see Section 14.5.
This MUST be used for the following response_modes:
dc_api.jwt
direct_post.jwt
[gco: Are the existing jarm response modes relevant? If so, we should include them here]
Note that for the ECDH JWE algorithms (from section 4.6 of [RFC7518]), the apu and apv values are inputs into the key derivation process that is used to derive the content encryption key. Regardless of algorithm used, the values are always part of the AEAD tag computation so will still be bound to the encrypted response. The following is a non-normative example of the payload of a JWE used in an Authorization Response:
{ … example unchanged … }
The JWE response document MUST include the vp_token and, if [DIF.PresentationExchange] was used in the request, the presentation_submission parameters as defined in Section 8.1.
The key material used for encryption SHOULD be determined using existing metadata mechanisms.
To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use jwks claim within the client_metadata request parameter, or within the metadata defined in the Entity Configuration when [OpenID.Federation] is used, or other mechanisms.
Note: For encryption, implementers have a variety of options available through JOSE, including the use of Hybrid Public Key Encryption (HPKE) as detailed in [I-D.ietf-jose-hpke-encrypt].
8.3.1. Response Mode "direct_post.jwt"
This specification also defines a new Response Mode direct_post.jwt, which allows for encryption to be used with Response Mode direct_post defined in Section 8.2.

The Response Mode direct_post.jwt causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in Section 8.2. The Wallet adds the response parameter containing the JWT as defined in Section 8.3 in the body of an HTTP POST request using the application/x-www-form-urlencoded content type. The names and values in the body MUST be encoded using UTF-8.

If a Wallet is unable to generate an encrypted response, it MAY send an error response without encryption as per Section 8.2.

The following is a non-normative example of a response using the presentation_submission and vp_token values from Appendix B.1.1. (line breaks for display purposes only):

POST /post HTTP/1.1
Host: client.example.org
Content-Type: application/x-www-form-urlencoded

  response=eyJra...9t2LQ
The following is a non-normative example of the payload of the JWT used in the example above before base64url encoding and encrypting:
{ … example unchanged … }

GarethCOliver:
Removes the JARM option for response signing, referencing JWT encryption directly

This continues to use the JARM client_metadata parameters, for compatibility and because we can not simply re-define them. How they are used for response encryption is updated and the option for signing removed. 

This closes #463, closes #438


openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,9 +295,8 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
+    * `jwks`: OPTIONAL. A JWKS as defined in [@!RFC7591]. It MAY contain one or more public keys, such as those used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#response_encryption)), or where the Wallet will require the public key of the Verifier to generate the Verifiable Presentation. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests.
     * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
-    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM], with an adjustment to the default behavior when this parameter is absent: instead of defaulting to `RS256`, the Authorization Response is not signed.
     * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
     * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
```
Sakurann, line 301: I agree it reads much cleaner without references to JARM. the only remaining awkward part is probably these two metadata parameters that are defined by JARM. a small note why we still use them might be helpful?

c2bo: I was wondering about the same - a small note that this mechanism originated from JARM but doesn't use the signing part might make sense?

GarethCOliver: Makes sense: I've updated to add a note on both how to use within this spec and explicitly say that JARM doesn't apply. Also included the definition text from JARM (to make reading it a bit clearer).

openid-4-verifiable-presentations-1_0.md
```diff
@@ -124,7 +124,10 @@ Verifiable Presentation (VP):
 W3C Verifiable Credential:
 : A Verifiable Credential compliant to the [@VC_DATA] specification.
 
-W3C Verifiable Presentation:
+W3C Verifiable Presentation:    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] enc algorithm REQUIRED for encrypting authorization responses. If authorization_encrypted_response_alg is specified, the default for this value is A128CBC-HS256. When authorization_encrypted_response_enc is included, authorization_encrypted_response_alg MUST also be provided.
+
+    Note: The encryption parameters `authorization_encrypted_response_alg` and `authorization_encrypted_response_enc` from [@!JARM] are reused to avoid redefining them. They are intended to be used for an unsigned, encrypted response (see (#response_encryption)). As signing is unnecessary, the mechanism specified in [@!JARM] do not apply.
+
```
Sakurann, line 130:
```suggestion
W3C Verifiable Presentation:
```
is this a copy paste error?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,11 +298,12 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
-    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] alg algorithm REQUIRED for encrypting authorization responses. See (#response_encyption) for usage. The default, if omitted, is that no encryption is performed.
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] enc algorithm REQUIRED for encrypting authorization responses. If authorization_encrypted_response_alg is specified, the default for this value is A128CBC-HS256. When authorization_encrypted_response_enc is included, authorization_encrypted_response_alg MUST also be provided.
+
+    Note: The encryption parameters `authorization_encrypted_response_alg` and `authorization_encrypted_response_enc` from [@!JARM] are reused to avoid redefining them. They are intended to be used for an unsigned, encrypted response (see (#response_encryption)). As signing is unnecessary, the mechanism specified in [@!JARM] do not apply.
```
Sakurann, line 306:
```suggestion
    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] algorithm for encrypting authorization responses. See (#response_encyption) for usage. The default, if omitted, is that no encryption is performed.
    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] encryption algorithm for encrypting authorization responses. If `authorization_encrypted_response_alg` is specified, the default for this value is A128CBC-HS256. When `authorization_encrypted_response_enc` is included, `authorization_encrypted_response_alg` MUST also be provided.

Note: Since the response is encrypted (see (#response_encryption)), but not signed, the mechanism specified in [@!JARM] does not apply. However, `authorization_encrypted_response_alg` and `authorization_encrypted_response_enc` metadata parameters from [@!JARM] are reused to avoid redefining them. ```

File: openid-4-verifiable-presentations-1_0.md [hunk 4]
```diff
@@ -295,11 +298,12 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
-    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM], with an adjustment to the default behavior when this parameter is absent: instead of defaulting to `RS256`, the Authorization Response is not signed.
-    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] alg algorithm REQUIRED for encrypting authorization responses. See (#response_encyption) for usage. The default, if omitted, is that no encryption is performed.
```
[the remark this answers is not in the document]
GarethCOliver: applied manually

Sakurann (approved):

babisRoutis (approved):

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,11 +295,12 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
-    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM], with an adjustment to the default behavior when this parameter is absent: instead of defaulting to `RS256`, the Authorization Response is not signed.
-    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] algorithm for encrypting authorization responses. See (#response_encryption) for usage. The default, if omitted, is that no encryption is performed.
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] encryption algorithm for encrypting authorization responses. If `authorization_encrypted_response_alg` is specified, the default for this value is A128CBC-HS256. When `authorization_encrypted_response_enc` is included, `authorization_encrypted_response_alg` MUST also be provided.
```
awoie, line 301: Any reason to choose A128CBC-HS256 over A128GCM? We are using A128GCM already in HAIP and it is more modern, faster, and robust (when implemented correctly).
```suggestion
    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] encryption algorithm for encrypting authorization responses. If `authorization_encrypted_response_alg` is specified, the default for this value is `A256GCM`. When `authorization_encrypted_response_enc` is included, `authorization_encrypted_response_alg` MUST also be provided.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,11 +295,12 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
+
+Note: Since the response is encrypted (see (#response_encryption)), but not signed, the mechanism specified in [@!JARM] does not apply. However, `authorization_encrypted_response_alg` and `authorization_encrypted_response_enc` metadata parameters from [@!JARM] are reused to avoid redefining them. ```
```
awoie, line 303:
```suggestion
Note: Since the response is encrypted (see (#response_encryption)), but not signed, the mechanism specified in [@!JARM] does not apply. However, `authorization_encrypted_response_alg` and `authorization_encrypted_response_enc` metadata parameters from [@!JARM] are reused to avoid redefining them.```
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2249,7 +2239,7 @@ Parameters defined by a specific client identifier scheme (such as the `trust_ch
 
 The `client_id` parameter MUST be present in signed requests defined in (#signed_request), as it communicates to the wallet which Client Identifier Scheme and Client Identifier to use when authenticating the client through verification of the request signature or retrieving client metadata.
 
-The value of the `response_mode` parameter MUST be `dc_api` when the response is neither signed nor encrypted and `dc_api.jwt` when the response is signed and/or encrypted as defined in (#jarm).
+The value of the `response_mode` parameter MUST be `dc_api` when the response is neither signed nor encrypted and `dc_api.jwt` when the response is encrypted as defined in (#response_encryption).
```
awoie, line 2242:
```suggestion
The value of the `response_mode` parameter MUST be `dc_api` when the response is not encrypted and `dc_api.jwt` when the response is encrypted as defined in (#response_encryption).
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,11 +295,12 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
-    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM], with an adjustment to the default behavior when this parameter is absent: instead of defaulting to `RS256`, the Authorization Response is not signed.
-    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] algorithm for encrypting authorization responses. See (#response_encryption) for usage. The default, if omitted, is that no encryption is performed.
```
awoie, line 300: Shouldn't the `response_mode` orchestrate whether encryption should be performed? If `dc_api.jwt` and there is no `alg`, then this is a error case. Same if the `jwks` does not contain a suitable encryption key.
```suggestion
    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] algorithm for encrypting authorization responses. See (#response_encryption) for usage.
```

awoie (commented):
Overall looks good but I had two comments: 1) default behaviour if `authorization_encrypted_response_alg` is not present, and 2) default `enc` value.

openid-4-verifiable-presentations-1_0.md
babisRoutis: > Shouldn't the `response_mode` orchestrate whether encryption should be performed? If `dc_api.jwt` and there is no `alg`, then this is a error case. Same if the `jwks` does not contain a suitable encryption key.

Totally agree @awoie

When `response_mode` requires encryption (`*.jwt`) then there should be  
- `authorization_encrypted_response_alg` (default or explicit)
- `authorization_encrypted_response_enc` (default or explicit)
- `jwks` should contain at least one suitable encryption key
- wallet should support all the above.

If any of the above conditions is not met then an error should be raised (It would be difficult though to communicated it back to the verifier) 

I believe the same goes, for the opposite case

When `response_mode` doesn't require encryption then
- `authorization_encrypted_response_alg` should be absent 
- `authorization_encrypted_response_enc` should be absent
- `jwks` should be absent or empty

PS: To be frank, I don't find so useful the defaults  (when an explicit value is not provided)

bc-pi, line 300: > Shouldn't the response_mode orchestrate whether encryption should be performed? 

I'd think so, yes. And I'd further suggest that `authorization_encrypted_response_alg` shouldn't be used at all with this change. The response_mode and public keys in the jwks have the necessary info. In fact doesn't the use of `authorization_encrypted_response_alg` preculde the utiltiy of providing more than one key in the jwks to give the wallet options for encryption alg?

bc-pi (changes requested):
authorization_encrypted_response_alg shouldn't be used

openid-4-verifiable-presentations-1_0.md
awoie: Would you then require the `alg` value to be present in the JWK, e.g., `ECDH-ES`, and then just mandate a specific encryption algorithm, e.g., `A128GCM`?

bc-pi: > Would you then require the `alg` value to be present in the JWK, e.g., `ECDH-ES`,

I wasn't thinking of going so far as requiring it but I suppose that could be done. Or at least mention it.

> and then just mandate a specific encryption algorithm, e.g., `A128GCM`?

using authorization_encrypted_response_enc with a reasonable default seems okay to me

openid-4-verifiable-presentations-1_0.md
GarethCOliver: Default purely came from JARM spec (so unchanged from the existing spec just repeated here for clarity.) I prefer babisRoutis's approach and mandate a value be specified, when there is no clear 'best' default, and will update to do that.

openid-4-verifiable-presentations-1_0.md
GarethCOliver: Agree this approach is better, was minimizing changes but this plus clearer language makes sense, will give a pass at updating the text.

martijnharing (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,11 +295,11 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
-    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM], with an adjustment to the default behavior when this parameter is absent: instead of defaulting to `RS256`, the Authorization Response is not signed.
-    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] enc algorithm REQUIRED for encrypting authorization responses. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present. Otherwise this should be absent.
+
+Note: Since the response is encrypted (see (#response_encryption)), but not signed, the mechanism specified in [@!JARM] does not apply. However, the  `authorization_encrypted_response_enc` metadata parameter from [@!JARM] is reused to avoid redefining it.```
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
Note: Since the response is encrypted (see (#response_encryption)), but not signed, the mechanism to sign, encrypt, or both sign and encrypt the Authorization Response specified in [@!JARM] does not apply. However, the  `authorization_encrypted_response_enc` metadata parameter from [@!JARM] is reused to avoid redefining it.```
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -340,7 +340,7 @@ The following additional considerations are given for pre-existing Authorization
 : OPTIONAL. Defined in [@!RFC6749]. The Wallet MAY allow Verifiers to request presentation of Verifiable Credentials by utilizing a pre-defined scope value. See (#request_scope) for more details.
 
 `response_mode`:
-: REQUIRED. Defined in [@!OAuth.Responses]. This parameter is used (through the new Response Mode `direct_post`) to ask the Wallet to send the response to the Verifier via an HTTPS connection (see (#response_mode_post) for more details). It is also used to request signing and encrypting (see (#jarm) for more details). 
+: REQUIRED. Defined in [@!OAuth.Responses]. This parameter is used (through the new Response Mode `direct_post`) to ask the Wallet to send the response to the Verifier via an HTTPS connection (see (#response_mode_post) for more details). It is also used to request encrypting (see (#response_encryption) for more details).
```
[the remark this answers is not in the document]
Sakurann: but those are two uses that are both required. so i think `it is also used` is beter. "to request encrypting" is not an optional use of this parameter, right?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1328,50 +1328,37 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
...
-
-This specification also defines how to encrypt an unsigned Authorization Response by adapting the mechanisms defined in [@!JARM]. The JSON containing the Authorization Response parameters can be encrypted as the payload of the JWE.
+## Encrypted Responses {#response_encryption}
 
-The advantage of an encrypted but not signed Authorization Response is that it prevents the signing key from being used as a correlation factor. It can also be a challenge to establish trust in the signing key to ensure authenticity. For security considerations with encrypted but unsigned responses, see (#encrypting_unsigned_response).
+This section defines how an Authorization Response containing a VP Token can be encrypted at the application level when the Response Type value is `vp_token` or `vp_token id_token`. Encrypting the Authorization Response can prevent personal data in the Authorization Response from leaking, when the Authorization Response is returned through the front channel (e.g., the browser). For security considerations see (#encrypting_unsigned_response).
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
This section defines how an Authorization Response containing a VP Token can be encrypted at the application level when the Response Type value is `vp_token` or `vp_token id_token`. Encrypting the Authorization Response can, for example, prevent personal data in the Authorization Response from leaking. For security considerations see (#encrypting_unsigned_response).
```
yeah i think that's ok. 

@martijnharing could you also please take a look at privacy considerations Nat has drafted in https://github.com/openid/OpenID4VP/issues/24#issuecomment-2779356349

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1328,50 +1328,37 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
...
 
-If the JWT is only a JWE, the following processing rules MUST be followed:
+To encrypt the Authorization Response, implementations MUST use an unsigned, encrypted JWT as described in [@!RFC7519]. The `enc` algorithm used MUST be obtained from `authorization_encrypted_response_enc` claim within the `client_metadata`. The `alg` algorithm used MUST be the `alg` value of the chosen `jwk`.
 
-- `iss`, `exp` and `aud` MUST be omitted in the JWT Claims Set of the JWE, and the processing rules as per [@!JARM] Section 2.4 related to these claims do not apply.
-- The processing rules as per [@!JARM] Section 2.4 related to JWS processing MUST be ignored.
+The Verifier's public key for input into the key agreement MUST be obtained from the `jwks` claim within the `client_metadata` request parameter, within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. The Wallet MAY select any key for which encryption is a valid operation.
```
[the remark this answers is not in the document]
Sakurann:
```suggestion
The Verifier's public key for input into the key agreement SHOULD be obtained from the `jwks` claim within the `client_metadata` request parameter, within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. The Wallet MAY select any key for which encryption is a valid operation.
```
sgtm

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1328,50 +1328,37 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
...
+This specification also defines a new Response Mode `direct_post.jwt`, which allows for encryption to be used with Response Mode `direct_post` defined in (#response_mode_post).
 
-The Response Mode `direct_post.jwt` causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in (#response_mode_post). The Wallet adds the `response` parameter containing the JWT as defined in Section 4.1. of [@!JARM] and (#jarm) in the body of an HTTP POST request using the `application/x-www-form-urlencoded` content type. The names and values in the body MUST be encoded using UTF-8.
+The Response Mode `direct_post.jwt` causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in (#response_mode_post). The Wallet adds the `response` parameter containing the JWT as defined in (#response_encryption) in the body of an HTTP POST request using the `application/x-www-form-urlencoded` content type. The names and values in the body MUST be encoded using UTF-8.
```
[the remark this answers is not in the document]
Sakurann: yeah, this section https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-A.4, right? 
(this comment is probably out of scope of this PR)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3204,6 +3191,7 @@ The technology described in this specification was made available from contribut
    * clarify that `dcql_query` and `presentation_definition` are passed as JSON objects (not strings) in request objects
    * support returning multiple presentations for a single dcql credential query when requested using `multiple`
    * Added support for multiple Client Identifiers and corresponding Request Signature to the DC API profile
+   * remove JARM and response signing, using JWT directly for unsigned, encrypted responses.
```
Sakurann, line 3194: i can move this to -26 in #497 or you can do it here

bc-pi (approved):
thanks @GarethCOliver for addressing my prior comments, LGTM now.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -295,11 +295,11 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
-    * `authorization_signed_response_alg`: OPTIONAL. As defined in [@!JARM], with an adjustment to the default behavior when this parameter is absent: instead of defaulting to `RS256`, the Authorization Response is not signed.
-    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM].
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM].
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] enc algorithm REQUIRED for encrypting authorization responses. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present. Otherwise this should be absent.
```
paulbastian, line 300:
```suggestion
    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] enc algorithm is REQUIRED for encrypting authorization responses. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present. Otherwise this SHOULD be absent.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1328,50 +1328,37 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
...
+## Encrypted Responses {#response_encryption}
 
-The advantage of an encrypted but not signed Authorization Response is that it prevents the signing key from being used as a correlation factor. It can also be a challenge to establish trust in the signing key to ensure authenticity. For security considerations with encrypted but unsigned responses, see (#encrypting_unsigned_response).
+This section defines how an Authorization Response containing a VP Token can be encrypted at the application level when the Response Type value is `vp_token` or `vp_token id_token`. Encrypting the Authorization Response can, for example, prevent personal data in the Authorization Response from leaking. For security considerations see (#encrypting_unsigned_response).
 
-If the JWT is only a JWE, the following processing rules MUST be followed:
+To encrypt the Authorization Response, implementations MUST use an unsigned, encrypted JWT as described in [@!RFC7519]. The `enc` algorithm used MUST be obtained from `authorization_encrypted_response_enc` claim within the `client_metadata`. The `alg` algorithm used MUST be the `alg` value of the chosen `jwk`.
```
paulbastian, line 1335: I think the last sentence is not clear enough, we mean a key from the `jwks` claim within `client_metadata` as written in the next section?

paulbastian (changes requested):

openid-4-verifiable-presentations-1_0.md
paulbastian: "the chosen `jwk`" is not easy to understand, as it has not been mentioned before I believe

bc-pi: I think this all touches on clarifications I (was) volunteered to make yesterday kinda per https://github.com/openid/OpenID4VP/pull/477#discussion_r2031613035 but I'll definitely try to address this kind of point. Hopefully a bit later today.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1328,50 +1328,37 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
...
 
-If a Wallet is unable to generate a JARM response, it MAY send an error response without using JARM as per (#response_mode_post).
+If a Wallet is unable to generate an encrypted response, it MAY send an error response as per (#response_mode_post).
 
 The following is a non-normative example of a response using the `presentation_submission` and `vp_token` values from (#jwt_vc). (line breaks for display purposes only):
 
 <{{examples/response/jarm_jwt_vc_json_post.txt}}
 
-The following is a non-normative example of the payload of the JWT used in the example above before base64url encoding and signing:
+The following is a non-normative example of the payload of the JWT used in the example above before base64url encoding and encrypting:
```
bc-pi, line 1361: to my eye, an order of operations was implied here and with the change from signing to encrypting it's not right anymore 

```suggestion
The following is a non-normative example of the payload of the JWT used in the example above before encrypting and base64url encoding:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1328,50 +1328,37 @@ Note: In the Response Mode `direct_post` or `direct_post.jwt`, the Wallet can ch
...
+The Response Mode `direct_post.jwt` causes the Wallet to send the Authorization Response using an HTTP POST request instead of redirecting back to the Verifier as defined in (#response_mode_post). The Wallet adds the `response` parameter containing the JWT as defined in (#response_encryption) in the body of an HTTP POST request using the `application/x-www-form-urlencoded` content type. The names and values in the body MUST be encoded using UTF-8.
 
-If a Wallet is unable to generate a JARM response, it MAY send an error response without using JARM as per (#response_mode_post).
+If a Wallet is unable to generate an encrypted response, it MAY send an error response as per (#response_mode_post).
 
 The following is a non-normative example of a response using the `presentation_submission` and `vp_token` values from (#jwt_vc). (line breaks for display purposes only):
```
bc-pi, line 1357: a "direct_post.jwt" example of a response using the `presentation_submission` and `vp_token` values... with PE having gone away and this having been a JWS rather than JWE... and merge conflicts but not here. I'm not sure I have the skills to make the mostly agreed on changes in the context of this PR. Might have to start a new one borrowing from Gareth's work as reference.

PR #478, opened 2025-03-27
Require `kid` from `jwks` to be included in JWE response header

### LINKED ISSUE #441: Requirement of `kid` in JARM response header?
Initially we supported the encryption rules for JARM based on 18013-7. The new definition in OID4VP is less strict (other related issue to this:  https://github.com/openid/OpenID4VP/issues/438).

For JARM responses we used the `kid` in the encrypted JWE header to determine which key should be used for decryption. There is no requirement on the `kid` in HAIP / OID4VP, meaning this approach won't work anymore.

The JARM spec defines:
> (OPTIONAL) The client decrypts the JWT using the key determined by the kid JWT header parameter. The key might be a private key, where the corresponding public key is registered with the expected issuer of the response ("use":"enc" via the client's metadata jwks or jwks_uri) or a key derived from its client secret (see section 4.2).

But since this step is marked as optional, I'm not 100% sure whether we can expect a `kid` in the response (is it optional because encryption is optional, or is it optional even if encryption is used?). Also Since there is no requirement to include the `kid` in the client metadata JWKs AFAIK.

I think it can be useful to make `kid` required to identify the encryption key used, but otherwise there might need to be some guidance on how to associate the encrypted response to the created request. For DC API it's easier (as the response is submitted by the frontend). For non-DC API requests I could see an unique response_uri per request working, however this conflicts with e.g. the requirements around `x509_san_uri` (which it seems is about to be removed in #430) where the `response_uri` must match a `uniformResourceIdentifier` in the x509 cert. It doesn't seem feasible to generate a new cert for every request.

TimoGlastra:
Fixes #441

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1355,7 +1355,7 @@ The JWT response document MUST include the `vp_token` and, if [@!DIF.Presentatio
 
 The key material used for encryption and signing SHOULD be determined using existing metadata mechanisms. 
 
-To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE header of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
+To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in Section 4.1.6 of [@!RFC7516]) of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
 
 To sign the Authorization Response, the Wallet MUST use a private key that corresponds to a public key made available in its metadata.
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1355,7 +1355,7 @@ The JWT response document MUST include the `vp_token` and, if [@!DIF.Presentatio
 
 The key material used for encryption and signing SHOULD be determined using existing metadata mechanisms. 
 
-To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the public key obtained from the `jwks` claim contains a `kid` parameter, the JWE MUST include the same value for the `kid` JWE response header parameter to hint to the Verifier which public key is used to encrypt the JWE.
+To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the public key obtained from the `jwks` claim contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE header of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
 
 To sign the Authorization Response, the Wallet MUST use a private key that corresponds to a public key made available in its metadata.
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1355,7 +1355,7 @@ The JWT response document MUST include the `vp_token` and, if [@!DIF.Presentatio
 
 The key material used for encryption and signing SHOULD be determined using existing metadata mechanisms. 
 
-To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms.
+To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the public key obtained from the `jwks` claim contains a `kid` parameter, the JWE MUST include the same value for the `kid` JWE response header parameter to hint to the Verifier which public key is used to encrypt the JWE.
```
awoie, line 1358:
```suggestion
To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the public key obtained from the `jwks` claim contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE header of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
```

awoie (approved):
LGTM, just a few minor suggestions.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1355,7 +1355,7 @@ The JWT response document MUST include the `vp_token` and, if [@!DIF.Presentatio
 
 The key material used for encryption and signing SHOULD be determined using existing metadata mechanisms. 
 
-To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms.
+To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the public key obtained from the `jwks` claim contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE header of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
```
jogu, line 1358: I think this text should apply if the key comes from federation too, so I suggest removing the "obtained from the 'jwks' claim".

```suggestion
To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE header of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
```

Sakurann (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1355,7 +1355,7 @@ The JWT response document MUST include the `vp_token` and, if [@!DIF.Presentatio
 
 The key material used for encryption and signing SHOULD be determined using existing metadata mechanisms. 
 
-To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms.
+To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE header of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
```
c2bo, line 1358: I guess it would make sense to reference the JWE RFC here?
```suggestion
To obtain Verifier's public key for the input to the key agreement to encrypt the Authorization Response, the Wallet MUST use `jwks` claim within the `client_metadata` request parameter, or within the metadata defined in the Entity Configuration when [@!OpenID.Federation] is used, or other mechanisms. If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in Section 4.1.6 of [@!RFC7516]) of the encrypted response. This enables the Verifier to identify the specific public key used to encrypt the response.
```

c2bo (approved):

charsleysa (approved):

Issue #300, opened 2024-10-31
Add important privacy/security considerations for DCQL


danielfett:
- When credential matching via claim_sets, the presence/absence of certain claims might leak data (all claims used for deciding response must be included in consent)
- Response must be the same, whether there is no match for credentials or the user didn't consent.

danielfett:
Also, issue #300 🎉

bc-pi:
> Also, issue #300 🎉

This is that, no?

danielfett:
> > Also, issue [#300](https://github.com/openid/OpenID4VP/issues/300) 🎉
> 
> This is that, no?

This was just for those who celebrate round numbers. 

Anyway, there's a PR now: #469

PR #479, opened 2025-03-28
Remove PE as a query language

### LINKED ISSUE #211: co-existence of multiple query languages (to PE, or not to PE)

danielfett:
This PR does the following:

- Remove Presentation Exchange as a query language incl. the auth request parameters presentation_definition and presentation_definition_uri
- <strike>Introduce the DCQL feature `contains` to enable matching W3C VC types, which so far was not possible with DCQL</strike> - reverted, see discussion
- Update the DCQL rules and extra parameters for W3C VCs and AnonCreds to ensure that useful queries can be made
- Update all relevant examples

Fixes #211 

merging this unblocks PR on #6


leecam (approved):
Might be better to split out the support for contains in a separate PR. But looks good to me either way

selfissued (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -916,6 +824,11 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+`contains`:
+: OPTIONAL. An array of strings or integers that, when specified, must appear in the matched array claim in any order.
+If the `contains` property is present, the Wallet SHOULD return the claim only if it is an array
+and if it contains all of the elements in the `contains` array, without taking into account ordering of elements.
+
```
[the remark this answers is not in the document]
danielfett: I don't think we do. I doubt there are many use cases for it, but if there are, they can be covered by providing multiple credential queries and then defining the alternatives on the credential set level.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -434,99 +417,23 @@ wallet_metadata=%7B%22vp_formats_supported%22%3A%7B%22jwt_vc_json%22%3A%7B%22alg
...
-```
 
 ## Using `scope` Parameter to Request Verifiable Credential(s) {#request_scope}
 
 Wallets MAY support requesting presentation of Verifiable Credentials using OAuth 2.0 scope values.
 
-Such a scope value MUST be an alias for 
-- a well-defined DCQL query, or
-- a well-defined Presentation Definition (for [@!DIF.PresentationExchange]) that will be referred to in the `presentation_submission` response parameter. 
+Such a scope value MUST be an alias for a well-defined DCQL query. Since more than one scope value can be used at the same time, the IDs for credentials (see (#credential_query)) and for claims (see (#claims_query)) within the DCQL queries in all scope values supported by a Wallet MUST be disjoint.
```
Sakurann, line 425:
```suggestion
Such a scope value MUST be an alias for a well-defined DCQL query. Since multiple scope values can be used at the same time, the identifiers for credentials (see (#credential_query)) and claims (see (#claims_query)) within each DCQL query associated with a different scope value MUST be unique.
```
trying to improve readability..

openid-4-verifiable-presentations-1_0.md
```diff
@@ -916,6 +823,11 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+`contains`:
+: OPTIONAL. An array of strings or integers that, when specified, must appear in the matched array claim in any order.
+If the `contains` property is present, the Wallet SHOULD return the claim only if it is an array
+and if it contains all of the elements in the `contains` array, without taking into account ordering of elements.
+
```
Sakurann, line 830:
```suggestion
`contains`:
: OPTIONAL. An array of strings or integers that, when specified, MUST appear in the matched array claim in any order.
When the `contains` property is present, the Wallet SHOULD return the claim only if the claim is an array that includes all elements listed in the `contains` array, irrespective of their order.

```
trying to improve readability

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
Sakurann: can this be used with sd-jwt vc and mdocs or is this specific to w3c vcdm? if the former, it is not clear to me how and would like some exampels... if the latter, it should be in w3c vcdm specific section, no?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1011,14 +924,14 @@ presentations.
...
+A claims path pointer is a pointer into the Verifiable Credential or Presentation, identifying one or more claims.
 A claims path pointer MUST be a non-empty array of strings, nulls and non-negative integers.
-A claims path pointer can be processed, which means it is applied to a credential. The results of
+A claims path pointer can be processed, which means it is applied to a Credential or Presentation. The results of
 processing are the referenced claims.
 
 ## Semantics for JSON-based credentials
 
-This section defines the semantics of a claims path pointer when applied to a JSON-based credential.
+This section defines the semantics of a claims path pointer when applied to a JSON-based credential or presentation.
```
Sakurann, line 934:
```suggestion
This section defines the semantics of a claims path pointer when applied to a JSON-based Credential or Presentation.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1421,14 +1303,6 @@ This document also defines the following additional error codes and error descri
...
 - The Wallet does not support any of the formats requested by the Verifier, such as those included in the `vp_formats` registration parameter.
 
-`invalid_presentation_definition_uri`:
-
-- The Presentation Definition URL cannot be reached.
-
-`invalid_presentation_definition_reference`:
-
-- The Presentation Definition URL can be reached, but the specified `presentation_definition` cannot be found at the URL.
-
```
Sakurann, line 1431: I support it, and we never defined `dcql_query_uri` parameter but I think it's worth noting that with this PR, we are losing the ability to pass query language by reference

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1605,9 +1474,7 @@ The following is a non-normative example of a set of static configuration values
 
 ## Nested Verifiable Presentations
 
-Current version of this document does not support presentation of a Verifiable Presentation nested inside another Verifiable Presentation, even though [@!DIF.PresentationExchange] specification theoretically supports this by stating that the nesting of `path_nested` objects "may be any number of levels deep".
-
-One level of nesting `path_nested` objects is sufficient to describe a Verifiable Credential included inside a Verifiable Presentation.
+This specification does not support presentation of a Verifiable Presentation nested inside another Verifiable Presentation
```
Sakurann, line 1477:
```suggestion
This specification does not support presentation of a Verifiable Presentation nested inside another Verifiable Presentation.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2348,19 +2182,33 @@ The security properties that are normally provided by the Client Identifier are
...
+OpenID for Verifiable Presentations is Credential Format agnostic, i.e., it is designed to allow applications to request and receive Verifiable Presentations and Verifiable Credentials in any Credential Format. This section defines a set of Credential Format specific parameters and rules for some of the known Credential Formats. For the Credential Formats that are not mentioned in this specification, other specifications or deployments can define their own set of Credential Format specific parameters.
 
 ## W3C Verifiable Credentials
 
-W3C Verifiable Credentials may use an additional parameter for the `descriptor_map` with the `presentation_submission`: The `path_nested` object inside an Input Descriptor Mapping Object is used to describe how to find a returned Credential within a Verifiable Presentation, and contains a `format` parameter with the Credential format identifier as a value and a `path` parameter with a relative path to the Verifiable Credential. Non-normative examples can be found further in this section.
+todo: define where the claims path operates
+todo: is proof_type_value correct?
```
Sakurann, line 2192: please make sure there are issues opened for this before we merge this PR

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2348,19 +2182,33 @@ The security properties that are normally provided by the Client Identifier are
...
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
+
+The following is a W3C Verifiable Credentials specific parameter in the meta parameter in a Credential Query as defined in (#credential_query):
```
Sakurann, line 2198:
```suggestion
The following is a W3C Verifiable Credentials specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query):
```

(sorry, let me be a backtick police..)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2518,18 +2337,17 @@ ISO/IEC TS 18013-7 Annex B [@ISO.18013-7] and ISO/IEC 23220-4 [@ISO.23220-4] Ann
 
 [@ISO.18013-7] defines the following elements:
 
-* Rules for the `presentation_definition` Authorization Request parameter.
-* Rules for the `presentation_submission` Authorization Response parameter.
+* Rules for the `presentation_definition` Authorization Request parameter and the `presentation_submission` Authorization Response parameter (no longer supported in this specification)
```
Sakurann, line 2340:
```suggestion
* Rules for the `presentation_definition` Authorization Request parameter and the `presentation_submission` Authorization Response parameter (which were supported until draft -24 of this specification)
```
this is probably more helpful..?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2764,57 +2576,29 @@ be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
...
-The following is a non-normative example of the content of the `presentation_submission` parameter:
-
-<{{examples/response/ps_sd_jwt_vc.json}}
-
-The following is a non-normative example of the `vp_token` parameter provided in the same response and referred to by the `presentation_submission` above:
-
-<{{examples/response/token_response_vp_token_sd_jwt_vc.txt}}
-
-In this example the `vp_token` contains only the disclosures for the claims specified in the `presentation_submission`, along with a Key Binding JWT.
+The `transaction_data_hashes` response parameter defined in (#transaction_data), if Transaction Data is used, MUST be included in the Key Binding JWT as a top level claim. This means that transaction data mechanism cannot be used with SD-JWT VCs without cryptographic key binding (i.e., which do not use the KB-JWT).
```
Sakurann, line 2591: this will probably conflict with #421. tho #421 probably getting merged first. @sander

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3183,6 +2921,8 @@ The technology described in this specification was made available from contribut
    * clarify that `dcql_query` and `presentation_definition` are passed as JSON objects (not strings) in request objects
    * support returning multiple presentations for a single dcql credential query when requested using `multiple`
    * Added support for multiple Client Identifiers and corresponding Request Signature to the DC API profile
+   * remove presentation exchange
```
Sakurann, line 2924:
```suggestion
   * remove DIF Presentation Exchange as a query language option
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3183,6 +2921,8 @@ The technology described in this specification was made available from contribut
...
+   * added `contains` to ensure features of PE can be mapped
```
Sakurann, line 2925: can be mapped to what..?

Sakurann (commented):
very much trusting Daniel that all the examples are correct. 

I am a little bit on the fence for a new `contains` parameter. would wait for a wg discussion..

danielfett:
Note so that we don't forget: This PR has an overlap with #300. We need to adapt the text added in #300 before/after we merge this.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: This is generic.

openid-4-verifiable-presentations-1_0.md
danielfett: The 'each' changes the meaning though.

openid-4-verifiable-presentations-1_0.md
danielfett: Removed - should not have made it into the commit.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2348,19 +2182,33 @@ The security properties that are normally provided by the Client Identifier are
...
-W3C Verifiable Credentials may use an additional parameter for the `descriptor_map` with the `presentation_submission`: The `path_nested` object inside an Input Descriptor Mapping Object is used to describe how to find a returned Credential within a Verifiable Presentation, and contains a `format` parameter with the Credential format identifier as a value and a `path` parameter with a relative path to the Verifiable Credential. Non-normative examples can be found further in this section.
+todo: define where the claims path operates
+todo: is proof_type_value correct?
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
+
+The following is a W3C Verifiable Credentials specific parameter in the meta parameter in a Credential Query as defined in (#credential_query):
+
```
TimoGlastra, line 2196: I think a custom `type_values` which is an array (or array of arrays to allow multiple type combinations) would be good to add here. My reasoning is that there is `doctype_values` and `vct_values`, so this will ensure consistency. Also, the `doctype_values` and `vct_values` are required, so also adding `type_values` means we also have a required meta type filter for w3c credentials (which was an issue with PEX)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2454,57 +2284,46 @@ AnonCreds is a Credential format defined as part of the Hyperledger Indy project
...
+`schema_id_values`:
+: OPTIONAL. An array of strings that specifies the allowed values for the `schema_id` of the requested Verifiable Credential. It MUST be a valid scheme identifier as defined in [@Hyperledger.Indy].
 
-<{{examples/request/request.txt}}
+`cred_def_id_values`:
+: OPTIONAL. An array of strings that specifies the allowed values for the `cred_def_id` of the requested Verifiable Credential. It MUST be a valid credential definition identifier as defined in [@Hyperledger.Indy].
 
-The following is a non-normative example of the content of the `presentation_definition` parameter:
+`proof_type_values`:
+: OPTIONAL. An array of strings that specifies the types of proofs that the Verifier accepts to be used in the Verifiable Presentation, for example `CLSignature2019`.
```
TimoGlastra, line 2301: I would not add this TBH. AnonCreds only supports CL signatures, I've never seen a deployment that does not CL signatures. There is effort ongoing for an AnonCreds v2, but this has a different model than AnonCreds v1,  so it should have a different format identifier IMO. So just the `ac_vp` identifier should be enough

examples/request/dcql_ac_vc_sd.json
```diff
@@ -0,0 +1,20 @@
+{
+  "credentials": [
+    {
+      "id": "example_vc_ac_sd",
+      "format": "ac_vp",
+      "meta": {
+        "proof_type_values": ["CLSignature2019"],
+        "schema_id_values": ["did:indy:idu:test:3QowxFtwciWceMFr7WbwnM:2:BasicScheme:0.1"]
```
TimoGlastra, line 8: The structure is not valid for `did:indy`

```suggestion
        "schema_id_values": ["did:indy:idu:test:3QowxFtwciWceMFr7WbwnM/anoncreds/v0/SCHEMA/BasicScheme/0.1"]
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2454,57 +2281,46 @@ AnonCreds is a Credential format defined as part of the Hyperledger Indy project
...
 
-<{{examples/request/pd_ac_vc.json}}
+#### Claims Matching
 
-The `format` object in the `input_descrioptor` element uses the format identifier `ac_vc` as defined above and sets the `proof_type` to `CLSignature2019` to denote this descriptor requires a Credential in AnonCreds format signed with a CL signature (Camenisch-Lysyanskaya signature). The rest of the expressions operate on the AnonCreds JSON structure.
+When used in the context of AnonCreds, the `claims_path` parameter always matches on the contents of the `values` key in the JSON-representation of the Verifiable Credential.
```
TimoGlastra, line 2305: This is a change of how the PEX definition handled Anoncreds credentials. The matching would be done on the outer wrapper (so not within the `values`), which would allow you to also query for e.g. the `issuer_did` / `issuer_id`. 

With this change you can only use DCQL to query on the values, and the meta defines `schema_id_values`, and `cred_def_id_values`. 

So I think we should either add a meta value to filter by did, allow dids to be used as trusted authorities, or query based on the outer object (so `['values', 'name']`), to not break issued based matching for anoncreds credentials

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2348,19 +2181,31 @@ The security properties that are normally provided by the Client Identifier are
...
+OpenID for Verifiable Presentations is Credential Format agnostic, i.e., it is designed to allow applications to request and receive Verifiable Presentations and Verifiable Credentials in any Credential Format. This section defines a set of Credential Format specific parameters and rules for some of the known Credential Formats. For the Credential Formats that are not mentioned in this specification, other specifications or deployments can define their own set of Credential Format specific parameters.
 
 ## W3C Verifiable Credentials
 
-W3C Verifiable Credentials may use an additional parameter for the `descriptor_map` with the `presentation_submission`: The `path_nested` object inside an Input Descriptor Mapping Object is used to describe how to find a returned Credential within a Verifiable Presentation, and contains a `format` parameter with the Credential format identifier as a value and a `path` parameter with a relative path to the Verifiable Credential. Non-normative examples can be found further in this section.
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
```
TimoGlastra, line 2193: One issue i see for w3c VCs is that issuer based matching with DCQl can become quite complex, as you don't know whether to query `['issuer']` or `['issuer', 'id']`. 

So if you would use claim sets, you'd have to duplicate all claim sets. I know the WG decided against adding `did` as a trusted authority mechanism, but it would make issuer matching a lot easier IMO.

openid-4-verifiable-presentations-1_0.md
danielfett:
```suggestion
   * added `contains` to ensure features of PE can be mapped to DCQL syntax
```

openid-4-verifiable-presentations-1_0.md
danielfett: Thanks, I will remove it.

openid-4-verifiable-presentations-1_0.md
danielfett: Thanks, I'll change it to match on the outer object then.

c2bo:
General remark regarding AnonCreds that I realised while checking this PR: If we decide to keep AnonCreds in for 1.0, then we should update that part of the spec and fix things like references (moved out of Indy a long time ago), nonce/audience. Happy to do so in a separate Issue/PR.

openid-4-verifiable-presentations-1_0.md
c2bo: Yeah, fehlt the same way to me for the W3C Parts: https://github.com/openid/OpenID4VP/pull/479#discussion_r2019986109. I do believe it would be a good idea to pull json-ld types matching out into format specific parts.

danielfett: The point about making it mandatory is a good one, as well as the lack of consistency at the moment. I'll change it to type_values.

c2bo: That will change with this PR: #480

c2bo (commented):
Looks mostly fine to me. My main concerns are the W3C / AnonCreds sections right now and I agree with Timo that a selection on type values similar to vct_values would make sense imho for W3C.

The AnonCreds related concerns don't really belong to this PR and we can move that into a separate issue/PR (fixing existing things not related to query language).

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2348,19 +2182,33 @@ The security properties that are normally provided by the Client Identifier are
...
+todo: is proof_type_value correct?
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
+
+The following is a W3C Verifiable Credentials specific parameter in the meta parameter in a Credential Query as defined in (#credential_query):
+
+`proof_type_values`:
+: OPTIONAL. An array of strings that specifies the types of proofs that the Verifier accepts to be used in the Verifiable Presentation, for example `RsaSignature2018`.
```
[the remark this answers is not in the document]
danielfett: I suggest to tackle this in a separate issue

examples/request/dcql_ac_vc_sd.json
```diff
@@ -0,0 +1,19 @@
...
+  "credentials": [
+    {
+      "id": "example_vc_ac_sd",
+      "format": "ac_vp",
+      "meta": {
+        "schema_id_values": ["did:indy:idu:test:3QowxFtwciWceMFr7WbwnM/anoncreds/v0/SCHEMA/BasicScheme/0.1"]
+      },
+      "claims": [
+        {
+          "path": ["values", "given_name"]
```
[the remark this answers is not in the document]
danielfett: See https://github.com/openid/OpenID4VP/pull/479#pullrequestreview-2730181906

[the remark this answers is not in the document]
TimoGlastra: I think directly querying on the values would be nice, but i think it's important to have a way to filter based on the issuer did.

Sakurann (approved):
approving given the latest changes that are W3C VCDM specific

Sakurann:
could we make a change proposed here in this PR..? https://github.com/openid/OpenID4VP/pull/480#issuecomment-2769647331

examples/request/dcql_ac_vc_sd.json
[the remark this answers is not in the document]
c2bo: Yeah I guess that makes sense.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
TimoGlastra: I would suggest to remove the `proof_type_values` for now then and add it back in if that's desired, as it clashes with `vp_formats`. I don't fully understand why we need this for W3C vcs, but not in the meta of SD-JWT / mDOC?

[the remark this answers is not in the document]
danielfett: That would work for me as well - let's discuss on the call.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2343,19 +2172,36 @@ The security properties that are normally provided by the Client Identifier are
...
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
+
+The following is a W3C Verifiable Credentials specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query):
+
+`proof_type_values`:
+: OPTIONAL. An array of strings that specifies the types of proofs that the Verifier accepts to be used in the Verifiable Presentation, for example `["RsaSignature2018"]`.
+
```
Sakurann, line 2190:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: I move this to metadata - please review this commit: https://github.com/openid/OpenID4VP/pull/479/commits/daee54150dde226443f965c6ad70a6887647ac90

openid-4-verifiable-presentations-1_0.md
danielfett:
```suggestion
Such a `scope` parameter value MUST be an alias for a well-defined DCQL query. Since multiple `scope` values can be used at the same time, the identifiers for credentials (see (#credential_query)) and claims (see (#claims_query)) within the DCQL queries associated with `scope` values MUST be unique.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -742,7 +649,7 @@ Each entry in `credentials` MUST be an object with the following properties:
 the constraints in `credential_sets`. The value MUST be a non-empty string
 consisting of alphanumeric, underscore (`_`) or hyphen (`-`) characters.
 Within the Authorization Request, the same `id` MUST NOT
-be present more than once.
+be present more than once (whether for credentials or for claim sets, defined below).
```
charsleysa, line 652: Is this change saying that ids must be completely unique within a query so a claim query id cannot use the same value as a credential query id and that the claim query id cannot be reused across credential queries?

e.g. would this be invalid?
```
{
  "credentials: [
    {
      "id": "query_1"
      ...
      "claims": [
        {"id": "query_1", "path": ["last_name"]},
        {"id": "query_2", "path": ["first_name"]},
      ]
    },
    {
      "id": "query_2",
      ...
      "claims": [
        {"id": "query_1", "path": ["last_name"]},
        {"id": "query_2", "path": ["first_name"]},
      ]
    }
  ]
}
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2343,19 +2172,42 @@ The security properties that are normally provided by the Client Identifier are
...
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
+
+The following is a W3C Verifiable Credentials specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query):
+
+`type_values`:
+: OPTIONAL. An array of arrays of strings that specifies the types of Verifiable Credentials that the Verifier accepts to be used in the Verifiable Presentation. Each of the top-level arrays specifies one alternative to match the type of the Verifiable Credential against. One level deeper, each array of strings specifies a set of types that must be present in the `type` property of the Verifiable Credential, irrespective of order and whether additional types are present.
```
sloops77, line 2189:
```suggestion
: OPTIONAL. An array of arrays of strings that specifies the types of Verifiable Credentials that the Verifier accepts to be used in the Verifiable Presentation. Each of the top-level arrays specifies one alternative to match the type of the Verifiable Credential against. Each inner array specifies a set of types that must be present in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

sloops77: I guess I get to be that guy - but that this PR is updating the DCQL spec as well as editorial changes as a consequence of removing PE, makes reviewing it difficult

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3204,6 +2943,7 @@ The technology described in this specification was made available from contribut
    * clarify that `dcql_query` and `presentation_definition` are passed as JSON objects (not strings) in request objects
    * support returning multiple presentations for a single dcql credential query when requested using `multiple`
    * Added support for multiple Client Identifiers and corresponding Request Signature to the DC API profile
+   * remove DIF Presentation Exchange as a query language option
```
sloops77, line 2946: add that there were updates to the w3c and anoncred specific dcql subsections

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2846,29 +2631,27 @@ GET /authorize?
   &id_token_type=subject_signed
   &client_id=x509_san_dns%3Aclient.example.org
   &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
-  &presentation_definition=...
```
sloops77, line 2849: @danielfett should this be replaced by a `dcq_queryl=...`?

danielfett: Good point - yes

openid-4-verifiable-presentations-1_0.md
danielfett: That's what it says, but not what I originally intended. I will revert this change.

openid-4-verifiable-presentations-1_0.md
danielfett: Will do.

openid-4-verifiable-presentations-1_0.md
TimoGlastra: Looking at the AnonCreds specification again and there is no `issuer_did` 
 property on the credential. It is allowed as a query param in an AnonCreds proof request restriction. 

So the change I proposed is not sufficient to query by issuer did. I've opened #494 to support `did` as trusted authorities.

TimoGlastra: However with `schema_id_values` and `cred_def_id` values already being part of the meta, i think querying directly on the values would be okay then, as it does not solve the issuer did querying. My apologies for giving a wrong solution.

openid-4-verifiable-presentations-1_0.md
TimoGlastra: I've opened https://github.com/openid/OpenID4VP/issues/494 to support did as trusted authorities, which would solve this issue, but it's not specific to this PR and can be worked around using claim sets

openid-4-verifiable-presentations-1_0.md
```diff
@@ -432,99 +415,23 @@ wallet_metadata=%7B%22vp_formats_supported%22%3A%7B%22jwt_vc_json%22%3A%7B%22alg
...
-```
 
 ## Using `scope` Parameter to Request Verifiable Credential(s) {#request_scope}
 
 Wallets MAY support requesting presentation of Verifiable Credentials using OAuth 2.0 scope values.
 
-Such a scope value MUST be an alias for 
-- a well-defined DCQL query, or
-- a well-defined Presentation Definition (for [@!DIF.PresentationExchange]) that will be referred to in the `presentation_submission` response parameter. 
+Such a `scope` parameter value MUST be an alias for a well-defined DCQL query. Since multiple `scope` values can be used at the same time, the identifiers for credentials (see (#credential_query)) and claims (see (#claims_query)) within the DCQL queries associated with `scope` values MUST be unique.
```
TimoGlastra, line 423: I wrote a whole comment on not understanding this requirement, but I think I understand now. The response contains only the DCQL query id as keys, so if you reuse the same ID across scope values it means you don't know which scope was used for the DCQL presentation. 

I think it could benefit from a small clarification, as it took me quite a while to understand the reasoning for this.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
TimoGlastra: I opened https://github.com/openid/OpenID4VP/issues/495

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2343,19 +2172,42 @@ The security properties that are normally provided by the Client Identifier are
...
+
+#### Claims Matching
+
+The `claims_path` parameter in the Credential Query as defined in (#credential_query) is used to specify the claims that the Verifier wants to receive in the Verifiable Presentation. When used in the context of W3C Verifiable Credentials, the `claims_path` parameter always matches on the root of Verifiable Credential (not the Verifiable Presentation). Examples are shown in the following subsections.
+
+### Verifier Metadata
+
+The `vp_formats` parameter of the Verifier metadata MUST have the key `jwt_vp_json` or `ldp_vp` (according to the format, as defined in (#jwt_vc) and (#ldp_vc)), and the value MUST be an object consisting of the following name/value pair:
+
+* `proof_type_values`: OPTIONAL. A JSON array containing types of proofs that
```
TimoGlastra, line 2201: This won't work for `jwt_vp_json` (see also #495)

TimoGlastra (approved):
This PR looks good to me, especially the removal of the Presentation Exchange part. 

I see some missing gaps related to DCQL querying and the Verifier Metadata which were highlighted because of the changes in this PR, for which I've opened #495 and #494 (I'm happy to a PR for both of those issues).

c2bo (approved):
Not sure about that one sentence to validate vp_token - everything else looks good to me

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3204,6 +2944,8 @@ The technology described in this specification was made available from contribut
    * clarify that `dcql_query` and `presentation_definition` are passed as JSON objects (not strings) in request objects
    * support returning multiple presentations for a single dcql credential query when requested using `multiple`
    * Added support for multiple Client Identifiers and corresponding Request Signature to the DC API profile
+   * remove DIF Presentation Exchange as a query language option
+   * Changes in the DCQL query parameters specific to W3C VCs and AnonCreds
```
jogu, line 2948: These should go against -26 instead of -25

jogu (approved):

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: Thanks!

openid-4-verifiable-presentations-1_0.md
danielfett: I'll change it back to matching on values then :-)

openid-4-verifiable-presentations-1_0.md
danielfett: Thanks, fixed after merging main.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1447,16 +1316,12 @@ This document also defines the following additional error codes and error descri
 
 Verifiers MUST validate the VP Token in the following manner:
 
-1. Validate the format of the VP Token as defined in (#response-parameters) and verify the contents depending on the language used in the Authorization Request:
-   1. If DCQL was used, ensure that the set of VPs returned satisfies all required Credential Sets (and optionally other Credential Sets).
-   1. If [@!DIF.PresentationExchange] was used, determine the number of VPs returned in the VP Token and identify in which VP which requested VC is included, using the Input Descriptor Mapping Object(s) in the Presentation Submission.
+1. Validate the format of the VP Token as defined in (#response-parameters) and check that the set of VPs returned satisfies all required Credential Sets (and optionally other Credential Sets).
```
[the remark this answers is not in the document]
danielfett: Done!

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1009,14 +917,14 @@ presentations.
 
 # Claims Path Pointer {#claims_path_pointer}
 
-A claims path pointer is a pointer into the Verifiable Credential, identifying one or more claims.
+A claims path pointer is a pointer into the Verifiable Credential or Presentation, identifying one or more claims.
```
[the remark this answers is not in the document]
danielfett: Good point, I'll remove it.

openid-4-verifiable-presentations-1_0.md
danielfett: I'll add a clarification!

openid-4-verifiable-presentations-1_0.md
```diff
@@ -264,13 +263,13 @@ that signals to the Wallet that it can make an HTTP POST request to the Verifier
...
 
 The Verifier communicates a Client Identifier Scheme that indicate how the Wallet is supposed to interpret the Client Identifier and associated data in the process of Client identification, authentication, and authorization as a prefix in the `client_id` parameter. This enables deployments of this specification to use different mechanisms to obtain and validate Client metadata beyond the scope of [@!RFC6749]. A certain Client Identifier Scheme MAY require the Verifier to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
 
 Depending on the Client Identifier Scheme, the Verifier can communicate a JSON object with its metadata using the `client_metadata` parameter which contains name/value pairs.
 
-This specification enables the Verifier to send both Presentation Definition JSON object and Client Metadata JSON object by value or by reference.
+This specification enables the Verifier to send the Client Metadata JSON object by value or by reference.
```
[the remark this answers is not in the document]
danielfett: Done.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -232,15 +232,14 @@ Figure: Cross Device Flow
...
+(2.5) The HTTP GET response returns the Request Object containing Authorization Request parameters. It especially contains a DCQL query that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Request Object and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers her consent to present the requested Credentials.
```
bc-pi, line 235: that is not a typical use of the word especially
```suggestion
(2.5) The HTTP GET response returns the Request Object containing Authorization Request parameters. It contains a DCQL query that describes the requirements of the Credential(s) that the Verifier is requesting to be presented. Such requirements could include what type of Credential(s), in what format(s), which individual Claims within those Credential(s) (Selective Disclosure), etc. The Wallet processes the Request Object and determines what Credentials are available matching the Verifier's request. The Wallet also authenticates the End-User and gathers her consent to present the requested Credentials. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2364,19 +2192,42 @@ The security properties that are normally provided by the Client Identifier are
...
-W3C Verifiable Credentials may use an additional parameter for the `descriptor_map` with the `presentation_submission`: The `path_nested` object inside an Input Descriptor Mapping Object is used to describe how to find a returned Credential within a Verifiable Presentation, and contains a `format` parameter with the Credential format identifier as a value and a `path` parameter with a relative path to the Verifiable Credential. Non-normative examples can be found further in this section.
+
+### DCQL
+
+#### Parameters in the `meta` parameter in Credential Query
+
+The following is a W3C Verifiable Credentials specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query):
+
+`type_values`:
+: OPTIONAL. An array of arrays of strings that specifies the types of Verifiable Credentials that the Verifier accepts to be used in the Verifiable Presentation. Each of the top-level arrays specifies one alternative to match the type of the Verifiable Credential against. Each inner array specifies a set of types that must be present in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types.
```
bc-pi, line 2209: On my first read of "array of arrays of strings" I thought it was a mistake. I soon came to see it is not a mistake. But it's not easy to read. Perhaps "An array of sting arrays"?

bc-pi (approved):
swooping in at the 11th hour with some trivial nitpicking suggestions but overall approve

openid-4-verifiable-presentations-1_0.md
bc-pi:
```suggestion
: OPTIONAL. An array of string arrays that specifies the types of Verifiable Credentials that the Verifier accepts to be used in the Verifiable Presentation. Each of the top-level arrays specifies one alternative to match the type of the Verifiable Credential against. Each inner array specifies a set of types that must be present in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

jogu:
Merging as per agreement on yesterday's WG call and noting that there have been no objections to Kristina explicitly asking on the mailing list if anyone had any concerns ( https://lists.openid.net/pipermail/openid-specs-digital-credentials-protocols/Week-of-Mon-20250331/000707.html )

PR #480, opened 2025-03-31
Make doctype_value and vct_values mandatory

### LINKED ISSUE #413: Optionality of mdoc specific DCQL fields
Section B.3.1 defines the doctype_value and intent_to_retain fields. In the ISO/IEC 18013-5 specification, both of these fields are mandatory. Unless there is a strong reason to not have the same requirements, these should be mandatory when an mdoc is requested when using OpenID.

martijnharing:
closes #413

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2534,7 +2534,7 @@ This section defines ISO mdoc specific DCQL Query and Response parameters.
 The following is an ISO mdoc specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query).
 
 `doctype_value`:
-: OPTIONAL. String that specifies an allowed value for the
+: MANDATORY. String that specifies an allowed value for the
 doctype of the requested Verifiable Credential. It MUST
 be a valid doctype identifier as defined in [@ISO.18013-5].
 
@@ -2758,7 +2758,7 @@ This section defines SD-JWT VC specific DCQL Query and Response parameters.
 The following is an SD-JWT VC specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query).
 
 `vct_values`:
-: OPTIONAL. An array of strings that specifies allowed values for
+: MANDATORY. An array of strings that specifies allowed values for
 the type of the requested Verifiable Credential. All elements in the array MUST
 be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
 MAY return credentials that inherit from any of the specified types, following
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2534,7 +2534,7 @@ This section defines ISO mdoc specific DCQL Query and Response parameters.
 The following is an ISO mdoc specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query).
 
 `doctype_value`:
-: OPTIONAL. String that specifies an allowed value for the
+: MANDATORY. String that specifies an allowed value for the
```
Sakurann, line 2537:
```suggestion
: REQUIRED. String that specifies an allowed value for the
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2758,7 +2758,7 @@ This section defines SD-JWT VC specific DCQL Query and Response parameters.
 The following is an SD-JWT VC specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query).
 
 `vct_values`:
-: OPTIONAL. An array of strings that specifies allowed values for
+: MANDATORY. An array of strings that specifies allowed values for
```
Sakurann, line 2761:
```suggestion
: REQUIRED. An array of strings that specifies allowed values for
```

danielfett:
if we merge #479, the new parameter types_values for W3C VCs needs to be updated to be MANDATORY as well.

danielfett (approved):

TimoGlastra (approved):

c2bo (approved):

danielfett:
@martijnharing please make the new `type_values` for W3C credentials mandatory as well and @c2bo can probably tell us which of `schema_id_values` and `cred_def_id_values` we should make mandatory as well

c2bo:
> @martijnharing please make the new `type_values` for W3C credentials mandatory as well and @c2bo can probably tell us which of `schema_id_values` and `cred_def_id_values` we should make mandatory as well

I'd say at least one of  `schema_id_values` or `cred_def_id_values` MUST be present. In theory it would be schema ID and did or cred_def_id as far as I understand the current AnonCreds iteration @TimoGlastra?

TimoGlastra:
Yes we basically use did + schema id to query, as this doesn't limit key rotation through the credential definition. 

I think that at least one of them needs to be defined makes sense. Not sure if it's an AND or an OR, but it's never really needed to include both.

danielfett:
In that case, requiring one of them to be present seems reasonable. @martijnharing

Sakurann:
WG discussion
for AnonCreds, one of the schema_id_values or cred_def_id_values must be present.
for W3C, type_values must be present

paulbastian (approved):

Sakurann (approved):

PR #481, opened 2025-03-31
Specify value-matching for ISO mdoc-based credentials.

### LINKED ISSUE #420: Value matching clarifications

davidz25:
closes #420

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -819,7 +819,7 @@ If the `values` property is present, the Wallet SHOULD return the claim only if 
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
-If a Wallet implements value matching and the credential to match against is
+If a Wallet implements value matching and the Credential to match against is
 an ISO mdoc-based credential, the following rules apply:
 
 - If the value to match against is a string, only claims encoded as a `tstr` (major
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -819,6 +819,23 @@ If the `values` property is present, the Wallet SHOULD return the claim only if 
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+If a Wallet implements value matching and the credential to match against is
+an ISO mdoc-based credential, the following rules apply:
+
+- If the value to match against is a string, only claims encoded as a `tstr` (major
+  type 2 according to [@RFC8949]) are considered and the value of the claim MUST
+  match the value to match against exactly. CBOR tags (major type 6) MUST be ignored.
+  For example, if the value to match against is `2025-03-31`, both the CBOR
+  values (in diagnostic notation) `"2025-03-31"` and `#6.1004("2025-03-31")` would
+  match.
+- If the value to match against is an integer, only claims encoded as unsigned
+  integers (major type 0), negative integers (major type 1), and `bignum` (as per
+  section 3.4.3 of [@RFC8949]) are considered. The CBOR value MUST match exactly against the
+  provided value. CBOR tags (major type 6) MUST be ignored.
+- If the value to match against is a boolean, only claims encoded as major type
+  7, additional information 20 (`false`) and 21 (`true`) are considered. The CBOR value 
+  MUST match exactly against the provided value. CBOR tags (major type 6) MUST be ignored.
+
 ### Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
 
 The following section describes the logic that applies for selecting claims 


openid-4-verifiable-presentations-1_0.md
```diff
@@ -948,6 +948,23 @@ Wallet and/or the End-User if the value matching request
...
+  values (in diagnostic notation) `"2025-03-31"` and `#6.1004("2025-03-31")` would
+  match.
+- If the value to match against is a integer, only claims encoded as unsigned
+  integers (major type 0), negative integers (major type 1), and `bignum` (as per
+  section 3.4.3 of [@RFC8949]) are considered. The CBOR value must match the
+  provided value to match against exactly.
+- If the value to match against is a boolean, only claims encoded as major type
+  7, additional information 20 (`false`) and 21 (`true`) are considered. The CBOR
+  value must match the provided value to match against exactly.
+
```
Sakurann, line 967:
```suggestion
If a Wallet implements value matching and the credential to match against is
an ISO mdoc-based credential, the following rules apply:

- If the value to match against is a string, only claims encoded as a `tstr` (major
  type 2 according to [@RFC8949]) are considered and the value of the claim MUST
  match the value to match against exactly. CBOR tags (major type 6) MUST be ignored.
  For example, if the value to match against is `2025-03-31`, both the CBOR
  values (in diagnostic notation) `"2025-03-31"` and `#6.1004("2025-03-31")` would
  match.
- If the value to match against is an integer, only claims encoded as unsigned
  integers (major type 0), negative integers (major type 1), and `bignum` (as per
  section 3.4.3 of [@RFC8949]) are considered. The CBOR value MUST match exactly against the
  provided value.
- If the value to match against is a boolean, only claims encoded as major type
  7, additional information 20 (`false`) and 21 (`true`) are considered. The CBOR value MUST match exactly against the provided value.

```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -948,6 +948,23 @@ Wallet and/or the End-User if the value matching request
 is followed. Therefore, Verifiers MUST treat restrictions expressed using `values` as a
 best-effort way to improve user privacy, but MUST NOT rely on it for security checks.
 
+If a Wallet implements value matching and the credential to match against is
```
danielfett, line 951: The text is fine, but this should be moved below the `values` definition in Line 917. The processing rules in the current section define details on selecting claims and credentials, it's not a perfect fit here.

Sakurann:
@davidz25 would it be possible to incorporate changes, if you agree with them, before monday hybrid dcp wg meeting, if possible? thank you!

openid-4-verifiable-presentations-1_0.md
```diff
@@ -948,6 +948,23 @@ Wallet and/or the End-User if the value matching request
...
+an ISO mdoc-based credential, the following rules apply:
+
+- If the value to match against is a string, only claims encoded as a `tstr` (major
+  type 2 according to [@RFC8949]) are considered and the value of the claim must
+  match the value to match against exactly. CBOR tags (major type 6) shall be ignored.
+  For example if the value to match against is `2025-03-31` both the values CBOR
+  values (in diagnostic notation) `"2025-03-31"` and `#6.1004("2025-03-31")` would
+  match.
+- If the value to match against is a integer, only claims encoded as unsigned
```
martijnharing, line 960: This text allows negative integers, however the definition of 'values' says that it can "strings, integers or boolean values". Looking at the JSON RFC, integer is defined as "int = zero / ( digit1-9 *DIGIT )" which does not include negative numbers.
I'm not sure which of this was the intention, but with the proposed text there is misalignment between the two formats.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -948,6 +948,23 @@ Wallet and/or the End-User if the value matching request
...
+  integers (major type 0), negative integers (major type 1), and `bignum` (as per
+  section 3.4.3 of [@RFC8949]) are considered. The CBOR value must match the
```
martijnharing, line 962: We should say here as well to ignore tags.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -948,6 +948,23 @@ Wallet and/or the End-User if the value matching request
 is followed. Therefore, Verifiers MUST treat restrictions expressed using `values` as a
 best-effort way to improve user privacy, but MUST NOT rely on it for security checks.
 
+If a Wallet implements value matching and the credential to match against is
+an ISO mdoc-based credential, the following rules apply:
```
martijnharing, line 952: Currently there are a number of items missing that can be encoded in CBOR. Some of these are probably not too critical (e.g. the 'undefined' item), others I'm not sure of (floating points, null), but make sense to have. But most importantly byte strings should be included, as this is a major tag that otherwise does not have a way to be matched against.

davidz25:
Updated PR to incorporate Kristina's changes and Daniel's suggestion of moving it to below values. Still need to address Martijn's comments, doing that now.

openid-4-verifiable-presentations-1_0.md
davidz25: Hmm, yeah, this is a problem. I think it's possible someone someday creates an ISO mdoc DocType with data elements with negative integers, it's not even that far-fetched... are people really not using negative integers in JSON documents?

openid-4-verifiable-presentations-1_0.md
davidz25: Should we though? Are there any real use-cases when you use a tagged integer and you then wish to compare against it? Only thing I can think about right now is Epoch-based time and I think for that you want to use the functions that @c2bo talked about during IIW this week for 1.1 so you can say e.g. fn_to_datestring() which returns e.g. an ISO 8601 datestring you can match against.

Either way I guess I'm fine with saying those tags are ignored, I don't think it costs anything. I'll incorporate that change.

openid-4-verifiable-presentations-1_0.md
davidz25: I think we can always add support for matching more values in 1.1, right?

For floating point comparison, this could be especially tricky when you only have `==` so that's why I didn't include this originally. Comparison against bstr seems weird, when would you ever do this? I guess it's possible someone could design a doctype that encodes e.g. bitmasks using a bstr but I think that's just bad doctype design...

openid-4-verifiable-presentations-1_0.md
davidz25: I don't think we can easily change the JSON RFC 🤣  ... so maybe we should also allow e.g. strings "42" and "-42" to also match CBOR major type 0 and 1? I think that would be pretty natural and this way users can use "-42" to match -42 as a CBOR value... it's not pretty but it will work. Thoughts?

openid-4-verifiable-presentations-1_0.md
martijnharing: I think bytestring is important to have as that’s a fairly common type.
It may require a normative change in how to identify the type in the request, so we should do that now instead of later.

martijnharing: Other solution would be to match against encoded cbor for all cases?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -819,6 +819,23 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+If a Wallet implements value matching and the credential to match against is
```
paulbastian, line 822:
```suggestion
If a Wallet implements value matching and the Credential to match against is
```

Sakurann:
WG discussion.
- replace this current text with a pointer to cbor-json conversion section 6.1 in the cbor spec, so that the current definition of `values` parameter applies both to json and cbor. in order to make sure this works for the bite string

Sakurann:
Superseded by #538

PR #482, opened 2025-03-31
add verifier_attestations parameter to authz req

### LINKED ISSUE #396: Add attachments in the presentation request
The [Verifier Attestation JWT](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-verifier-attestation-jwt) is good for authentication, but limited in further usage. Current requirements from the [Implementing Acts](https://ec.europa.eu/info/law/better-regulation/have-your-say/initiatives/14399-European-Digital-Identity-Wallets-registration-of-relying-parties_en) require that the Relying Party need to present a so called `Registration Certificate`, where the intended use is attested by the Wallet-Relying Party Registrar. To avoid requests from the wallet to the registrar, the Relying Party should be able to attach this certificate to the presentation request.

The request includes a new `credentials` attribute, that allows to attach
certificates like registration and authorization certificates.

- **`type`**: STRING, media type of the attached credential, helping to parse the credential by the Wallet. The value has to be registered in the [IANA media types](https://www.iana.org/assignments/media-types/media-types.xhtml) registry.
- **`data`**: ANY, content of the credential. Will be parsed by the Wallet based on the type.
- **`credentialsMatch`**: STRING[], when the provided certificate is used for authorization, it will list the credential ids from the `dcql_query`.

```json
{
  "x5c": ["mcnwmx...."], certificate to prove the signature
  "typ": "JWT",
  "alg": "ES256"
}
.
{
  "response_type":"vp_token",
  "client_id":"x509_san_dns:client.example.org",
  "response_mode":"direct_post",
  "dcql_query":{
    "credentials":[
      {
        "id":"my_credential",
        "format":"vc+sd-jwt",
        "meta":{
          "vct_values":[
            "https://credentials.example.com/identity_credential"
          ]
        },
        "claims":[
          {
            "path":[
              "last_name"
            ]
          },
          {
            "path":[
              "first_name"
            ]
          },
          {
            "path":[
              "address",
              "street_address"
            ]
          }
        ]
      }
    ]
  },
  "credentials":[
    {
      "type:" "jose",
      "data": "base64-encoded-registrationCertificate-jws"
    },
    {
      "type:" "jose",
      "data": "base64-encoded-authorizationCertificate-jws",
      "credentialsMatch": ["my_credential"]
    }
  ]
}.ejsdbwjhsbs
```

cre8:
Closes #396


Sakurann:
Please add a change log :)

cre8:
> Please add a change log :)

done :)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3169,6 +3202,7 @@ The technology described in this specification was made available from contribut
 
    -25
 
+   * add `attechments` to list of authorization parameters
```
jogu, line 3205:
```suggestion
   * add `attachments` to list of authorization parameters
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -331,6 +331,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
```
jogu, line 334: Please add to IANA considerations too, [Appendix D.2](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-D.2)

I think you also want to add it to the list of allowed parameters for Browser DC API?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -708,6 +726,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
 
+## Attachments in Authorization Requests {#attachments-in-authorization-requests}
+
+This section provides guidance on processing and security considerations for the `attachments` parameter defined in (#new_parameters).
+
+### Proof of Possession
+
+In some cases, the Verifier may want to include attachments that are bound to a key or require verification of their origin.
+
+It is up to the ecosystem profile to define how proof of possession is conveyed. Verifiers SHOULD use the `nonce` parameter from the Authorization Request in any proof-of-possession calculation to prevent replay attacks. The Wallet MUST validate such proofs if defined by the profile and ignore or reject attachments that fail validation.
```
jogu, line 737: A verifier generated nonce doesn't prevent replay of a verifier generated object? It definitely does something but I think it's closer to binding it to the response.

jogu (changes requested):

openid-4-verifiable-presentations-1_0.md
cre8: good point, added to both

openid-4-verifiable-presentations-1_0.md
```diff
@@ -331,6 +331,24 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. In [@!DIF.PresentationExchange], the string matches the `id` field in the Input Descriptor. In the Digital Credentials Query Language, the string matches the `id` field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
+
+The Wallet MUST ignore any unrecognized attachment formats, see (#attachments-in-authorization-requests) for more details.
+
+The following is a non-normative example of an attachment object:
+
+```
+{
+  "format": "eudi_registration_certificate",
+  "data": "eyJhbGciOiJFUzI1....EF0RBtvPClL71TWHlIQ"
```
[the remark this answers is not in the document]
cre8: updated it to say it can have any type

openid-4-verifiable-presentations-1_0.md
cre8: You are right, I changed it to bind it to the transaction. Open for another good approach

openid-4-verifiable-presentations-1_0.md
```diff
@@ -331,6 +331,24 @@ The following is a non-normative example of a transaction data content, after ba
...
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. Each object contains the following parameters:
+
+    * `format`: REQUIRED: String that identifies the format of the attachment.
+    * `data`: REQUIRED: Object that contains the data of the attachment. It will be parsed by the Wallet according to the `format` parameter.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. In [@!DIF.PresentationExchange], the string matches the `id` field in the Input Descriptor. In the Digital Credentials Query Language, the string matches the `id` field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
+
+The Wallet MUST ignore any unrecognized attachment formats, see (#attachments-in-authorization-requests) for more details.
```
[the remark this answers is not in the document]
cre8: I merged the first sentence in the paragraph above to make clear that more information are in the other chapter (and that the reference is not limited to the attachment format)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3188,8 +3228,12 @@ The technology described in this specification was made available from contribut
 
    [[ To be removed from the final specification ]]
 
-   -25
+   -26
+  
+   * add `attachments` to list of authorization parameters
 
+   -25
+
```
Sakurann, line 3236:
```suggestion
   -25
   
   * add `attachments` to list of authorization parameters
```
why is this suddenly -26? what did i miss?

cre8: I thought because 25 got already published I needed to put it into 26. But I can change it back to 25

cre8: I reverted that change, it's back in 25

openid-4-verifiable-presentations-1_0.md
```diff
@@ -329,6 +329,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. The Wallet MUST ignore any unrecognized attachment formats. Each object contains the following parameters:
+
+    * `format`: REQUIRED: String that identifies the format of the attachment.
+    * `data`: REQUIRED: The structure and content of data are format-dependent and not constrained by this specification.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. In [@!DIF.PresentationExchange], the string matches the `id` field in the Input Descriptor. In the Digital Credentials Query Language, the string matches the `id` field in the Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
```
TimoGlastra, line 337: This needs to  be synced with changes from #479 where PresentationExchange is removed

TimoGlastra, line 337:
```suggestion
    * `credential_ids`: REQUIRED. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. The string matches the `id` field in the DCQL Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
```

openid-4-verifiable-presentations-1_0.md
jogu: I think you were right the first time @cre8 , this should be in -26. Although technically you only became correct this morning when -25 was actually published :)

jogu (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -329,6 +329,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. The Wallet MUST ignore any unrecognized attachment formats. Each object contains the following parameters:
+
+    * `format`: REQUIRED: String that identifies the format of the attachment.
+    * `data`: REQUIRED: The structure and content of data are format-dependent and not constrained by this specification.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. The string matches the `id` field in the DCQL Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
```
c2bo, line 337:
```suggestion
    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. Each string matches the `id` field in a DCQL Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -706,6 +724,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+### Proof of Possession
+
+In some cases, the Verifier may want to include attachments that are bound to a key or require verification of their origin.
+
+It is up to the ecosystem profile to define how proof of possession is conveyed. Verifiers SHOULD use the `nonce` parameter from the Authorization Request in any proof-of-possession calculation to bind to to this transaction. The Wallet MUST validate such proofs if defined by the profile and ignore or reject attachments that fail validation.
+
+### Multi-Relying Party Behavior
+
+In some cases, the Verifier may want to include attachments based on a specific eco system profile. In such a scenario, the Authorization Request MUST sent a signed request in the JWS JSON Serialization format, see (#jws-json-serialization).
```
c2bo, line 739: not entirely sure, but I think we should introduce the binding to client id (prefix) here
```suggestion
In some cases, the Verifier may want to include attachments based on specific ecosystem profiles bound to specific Client Identifier Prefixes. In such a scenario, the Authorization Request MUST sent a signed request in the JWS JSON Serialization format, see (#jws-json-serialization).
```

openid-4-verifiable-presentations-1_0.md
cre8: added to v26 again

openid-4-verifiable-presentations-1_0.md
cre8: good point to clarify it, I added it

openid-4-verifiable-presentations-1_0.md
```diff
@@ -706,6 +724,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
 
+## Attachments in Authorization Requests {#attachments-in-authorization-requests}
+
+This section provides guidance on processing and security considerations for the `attachments` parameter defined in (#new_parameters).
+
+### Proof of Possession
+
+In some cases, the Verifier may want to include attachments that are bound to a key or require verification of their origin.
+
+It is up to the ecosystem profile to define how proof of possession is conveyed. Verifiers SHOULD use the `nonce` parameter from the Authorization Request in any proof-of-possession mechanism to bind to this transaction. The Wallet MUST validate such proofs if defined by the profile and ignore or reject attachments that fail validation.
```
jogu, line 735: Having thought about this more I'm not convinced nonce is sufficient to bind to the transaction. I think you need at least client id embedded in to the proof of possessions too.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -329,6 +329,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. The Wallet MUST ignore any unrecognized attachment formats. Each object contains the following parameters:
+
+    * `format`: REQUIRED: String that identifies the format of the attachment.
```
jogu, line 335: I think we should be clearer that this isn't defined in OID4VP - perhaps we repeat the same text we use for `transaction_data` types?

```suggestion
    * `format`: REQUIRED: String that identifies the format of the attachment. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for type values.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -329,6 +329,24 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `data`: REQUIRED: The structure and content of data are format-dependent and not constrained by this specification.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. Each string matches the `id` field in a DCQL Credential Query. If there is more than one element in the array, the Wallet MUST use only one of the referenced Credentials for transaction authorization.
```
jogu, line 337: "transaction authorization" is wrong here right?

jogu, line 337: > Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant

It feels like we should probably say something about what the meaning of omitting `credential_ids` is? I guess something like "If omitted the attachment is relevant to all requested credentials."

openid-4-verifiable-presentations-1_0.md
```diff
@@ -706,6 +724,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+## Attachments in Authorization Requests {#attachments-in-authorization-requests}
+
+This section provides guidance on processing and security considerations for the `attachments` parameter defined in (#new_parameters).
```
jogu, line 729: I think a slightly longer introduction here would be helpful - e.g. in the same way as https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-8.4 does at least give an example of what it could be used for?

danielfett:
@cre8 Feel free to also add your name to the acknowledgements!

openid-4-verifiable-presentations-1_0.md
cre8: I updated it to make clear that the usage depends on the `format` value and therefore the context how to use it.

openid-4-verifiable-presentations-1_0.md
cre8: Added some examples. As mentioned above the logic if something should be aborted or so is defined by the value of the format and therefore by the context.

openid-4-verifiable-presentations-1_0.md
jogu: > Or not at all. You could also pass something like a CRL list or so that has nothing to do directly with the requested credentials, but is needed to process the request (in this case validating the status of the x509 that was used for signing).

I think I feel nervous about this. It feels like we're defining an extension point on top of something that's already an extension point. If you need to pass a CRL list why wouldn't it be better to define an OAuth level parameter for passing the CRL in? (And profiles can absolutely define extra ecosystem-specific oauth parameters.)

cre8: I also thought about this, so I only gave examples that are relevant in the credentials context like the requirements from the EU (registration certificates and authorization attestations).

We can also write to limit it that you only need attachments there that are required to fetch credentials.

openid-4-verifiable-presentations-1_0.md
cre8: including the client id reduces the risk of the replay attack. I would prefer to add some suggestions to build the proof than leaving it open to the implementations, risking vulnerabilities.

cre8: Added the `client_id` as a parameter.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. The Wallet MUST ignore any unrecognized attachment formats. Each object contains the following parameters:
+
+    * `format`: REQUIRED: String that identifies the format of the attachment. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for type values.
+    * `data`: REQUIRED: The structure and content of data are format-dependent and not constrained by this specification.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted the attachment is relevant to all requested credentials.
```
c2bo, line 328: I think it would be highly beneficial if the wallet would be able to identify an attachment without having to parse it. My assumption right now would be that data and format doesn't uniquely identify an attachment? 

I do believe adding something like an ID that is uniquely defined in an ecosystem on this layer would help to make sure the wallet is not parsing payloads it doesn't need or won't be able to understand?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +629,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+## Attachments in Authorization Requests {#attachments-in-authorization-requests}
+
+The attachment data mechanism enables to pass additional information to the Wallet to request credentials. The Verifier MAY include attachments in the Authorization Request, for example to pass a registration certificate to inform about the registered intended use or an authorization attestation to request a specific credential from the Wallet to comply with the embedded disclosure policy.
```
c2bo, line 634:
```suggestion
The attachment data mechanism enables a Verifier to pass additional information to the Wallet. The Verifier MAY include attachments in the Authorization Request, for example to pass a registration certificate to inform about the registered intended use or an authorization attestation to request a specific credential from the Wallet to comply with the embedded disclosure policy.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +629,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+### Proof of Possession
+
+In some cases, the Verifier may want to include attachments that are bound to a key or require verification of their origin.
+
+It is up to the ecosystem profile to define how proof of possession is conveyed. Verifiers SHOULD use the `nonce` and `client_id` parameter from the Authorization Request in any proof-of-possession mechanism to bind to this transaction. The Wallet MUST validate such proofs if defined by the profile and ignore or reject attachments that fail validation.
```
c2bo, line 640:
```suggestion
It is up to the ecosystem profile to define how a proof of possession is conveyed. Verifiers SHOULD use the `nonce` and `client_id` parameter from the Authorization Request in any proof-of-possession mechanism to bind to this transaction. The Wallet MUST validate such proofs if defined by the profile and ignore or reject attachments that fail validation.
```

c2bo (commented):
I think this PR is in a decent state. 

The only thing that worries me a bit is the question if we should add some form of ID on the root level of attachments to help a Wallet make a decision whether it needs to parse an attachment or not. I don't think it is a good idea to force a wallet to parse all attachments (in formats it understands) before being able to make a decision if it needs or understands them or not.
Would feel cleaner to me if we added said `id` parameter to entries of `attachments`. Ecosystems would then need to define ids in a collision-resistant fashion.

openid-4-verifiable-presentations-1_0.md
TimoGlastra: As i understood it, format should be enough, and i think we should encourage that. Attachments are ecosystem / use case specific. 

The `eudi_registration_certificate` already acts like a kind of ID here (And i think we should avoid overloading the id parameter). ID feels more appropriate for an instance of an attachment, not the type of an attachment.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. The Wallet MUST ignore any unrecognized attachment formats. Each object contains the following parameters:
+
+    * `format`: REQUIRED: String that identifies the format of the attachment. The specific values are out of scope of this specification. It is RECOMMENDED to use collision-resistant names for type values.
```
TimoGlastra, line 326: I don't have an issue with this being "format", but is there a specific reason transaction data uses "type" and this uses "format"? I think using "type" could make sense so that attachments/transaction data work somewhat consistently (also with "credential_ids"). Then "format" is sort of reserved for credential formats. 

I also think type makes sense as the format here is kinda like a (media) type.

But again, not too important

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +629,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+### Proof of Possession
+
+In some cases, the Verifier may want to include attachments that are bound to a key or require verification of their origin.
+
+It is up to the ecosystem profile to define how proof of possession is conveyed. Verifiers SHOULD use the `nonce` and `client_id` parameter from the Authorization Request in any proof-of-possession mechanism to bind to this transaction. The Wallet MUST validate such proofs if defined by the profile and ignore or reject attachments that fail validation.
+
+### Multi-Relying Party Behavior
+
+In some cases, the Verifier may want to include attachments based on specific ecosystem profiles bound to specific Client Identifier Prefixes. In such a scenario, the Authorization Request MUST sent a signed request in the JWS JSON Serialization format, see (#jws-json-serialization).
```
TimoGlastra, line 644:
```suggestion
In some cases, the Verifier may want to include attachments based on specific ecosystem profiles bound to specific Client Identifier Prefixes. In such a scenario, the Authorization Request MUST send a signed request in the JWS JSON Serialization format, see (#jws-json-serialization).
```

openid-4-verifiable-presentations-1_0.md
TimoGlastra, line 328: Isn't credential_ids already very specific to a particular attachment format? Is there a specific reason to define this as part of the base structure instead of leaving it to the format specific data field?

TimoGlastra (approved):
LGTM, some questions but mostly curious about reasoning

openid-4-verifiable-presentations-1_0.md
c2bo: I'd agree that ID was probably not the best proposal here.

Let's rename to something like type if we expect this parameter to convey more than a data format imho and make that more clear in the description? I am happy for things to stay the same if this parameter conveys both
- type of the payload (e.g., EUD RP Registration)
- encoding format (e.g. JWT) to allow the Wallet to decode
but it should be a bit more clear in the description imho.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: I think the big difference is that this is on a different meta level. An attachment will have both, a format (credential format) and a type (expressed in terms determined by the format). It's basically like in DCQL queries. Transaction data types, in contrast, are always JSON, so we can omit the format.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,24 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`attachments`:
+: OPTIONAL. Array of objects, where each object includes an extra resource that needs to be passed to the Wallet. These resources are intended to provide additional context to the Wallet or the End-User. The Wallet MUST ignore any unrecognized attachment formats. Each object contains the following parameters:
```
tlodderstedt, line 324: As far as I understand, the intention of this parameter is to convey attestations/assertions/certificates confirming attributes of the Relying Party. Given this assumption, I find the name of the parameter and the description rather generic. Why is that?
I would rather expect something along the lines of "rp_credentials" or "client_credentials" or "client_attestations" or even "vp_token" since for me the difference between the `vp_token` response parameter and this parameter is just the direction and the entity the credential (presentations) are related to.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: why don't we structure this like DCQL queries or VCI requests? The format determines how the type is specified and the rest is determined by the type. I don't think we need an ID as long as the `credential_ids` allow the wallet to determine which attachments shall be used for what context, which is especially important if multiple attachment of same format/type are in the request.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,24 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `data`: REQUIRED: The structure and content of data are format-dependent and not constrained by this specification.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attachment is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted the attachment is relevant to all requested credentials.
+
+See (#attachments-in-authorization-requests) for more details.
+
+The following is a non-normative example of an attachment object:
+
+```
+{
+  "format": "eudi_registration_certificate",
```
tlodderstedt, line 336: That suprises me. I would have expected this to be "jwt" (for example) and then an additional attribute "type" determining the type as "eudi_registration_certificate". The type would also be determined from `typ`JSON header or a `vct` claim within the JWT.

openid-4-verifiable-presentations-1_0.md
tlodderstedt, line 634: this needs more explanation, right now it is too much EUDIW lingo and does not describe the broader context.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +629,20 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+## Attachments in Authorization Requests {#attachments-in-authorization-requests}
+
+The attachment data mechanism enables to pass additional information to the Wallet to request credentials. The Verifier MAY include attachments in the Authorization Request, for example to pass a registration certificate to inform about the registered intended use or an authorization attestation to request a specific credential from the Wallet to comply with the embedded disclosure policy.
+
+### Proof of Possession
+
+In some cases, the Verifier may want to include attachments that are bound to a key or require verification of their origin.
```
tlodderstedt, line 638: isn't that the normal case?
And what does verification of origin mean? Origin of the credential or origin of the request?

openid-4-verifiable-presentations-1_0.md
tlodderstedt, line 640: Please explain you mean the creation and provision of a digital signature as PoP. 

I think the section must zoom out a bit. First of all, it needs to discuss that the wallet must (in one way or the other) be able to bind the attachment to the relying party. There are two ways, either the attachment is bound to attributes confirmed by the client authentication or the client sends another proof of possession for a key included in the attachment. Please add those considerations.

openid-4-verifiable-presentations-1_0.md
tlodderstedt, line 644: Why don't you just add this to the multi RP section in the DC API Annex? Then the context would be cristal clear.

tlodderstedt (changes requested):

openid-4-verifiable-presentations-1_0.md
c2bo: type is absolutely fine for me and a better name. When I wrote ID, i basically meant type information - some unique identifier that allows the wallet to figure out what this attachment would be used for.

I agree that if you have type+format information, then the `credential_ids` reference is enough for the required logic 👍

cre8: type is better than format. But will there be a case where an attachment of a specific type can be represented in different formats? In case of JWT vs. json serialization, it could then be like `eudi_registration_certificate+json` or. `eudi_registration_certificate+jwt`. IMHO I think the amount of the multiple formats are so rar that we should just use the `type` (ex format field` for identification and the wallet will only process the ones it knows.

cre8: In case of an authorization credential, you need to tell the Wallet to which credential it is mapped, because there is no direct link from the authorization credential to the requested credential.
It seems to be a better approach than looping over all credentials to check if there is at least one that could match

openid-4-verifiable-presentations-1_0.md
cre8: I am fine renaming it. Joseph told me that we should limit it only for the purpose of credential requests and not to attach something like an ocsp response that can be used to validate the x509 that was used to sign the request.

`client_attestations` sounds better than `vp_token` because this name is already used (to have a better separation)

openid-4-verifiable-presentations-1_0.md
cre8: I don't see the case why why an eco system should define different formats (json, string) for the same type. When defining a specific attachment, the unique identifier like `eudi_registration_certificate` will be defined in the ecosystem how to process it (so also how to decode the information if required and how to consume the information).

Defining a generic `format` field with the values like `string` | `json`, ... does not seam good, especially since the type needs to be string based to embed it in json. That's why the iana media type was removed for the format field and replaced by a unique identifier.

openid-4-verifiable-presentations-1_0.md
cre8: I am open for other use cases, but I think the main points is for authentication/authorization.

In most cases authorization

openid-4-verifiable-presentations-1_0.md
cre8: From a readers perspective this would be the first place for me to search for it (like the specific behavior when adding a feature. But we can move it down.

openid-4-verifiable-presentations-1_0.md
cre8: This phrase was still there when it was not limited to the credential request. In this case, POP is relevant for all attestations, so I will update it.

cre8:
> I think this PR is in a decent state.
> 
> The only thing that worries me a bit is the question if we should add some form of ID on the root level of attachments to help a Wallet make a decision whether it needs to parse an attachment or not. I don't think it is a good idea to force a wallet to parse all attachments (in formats it understands) before being able to make a decision if it needs or understands them or not. Would feel cleaner to me if we added said `id` parameter to entries of `attachments`. Ecosystems would then need to define ids in a collision-resistant fashion.

I think the best approach would be to follow the JWT approach where the `typ` field defines the type of JWT and then there is a registry for the eco system or a global one where the values and how to deal with the attachment is defined. Since this field is on the root level of the attachment, the Wallet only needs to parse this value without anything else and can then decide how to continue.

openid-4-verifiable-presentations-1_0.md
TimoGlastra: I agree @cre8, i think adding a separate type and format overcomplicates things. The type can mandate the format. And if there need to be multiple formwts we can use +jwt, +cwt, ..

openid-4-verifiable-presentations-1_0.md
cre8: Removed this section

openid-4-verifiable-presentations-1_0.md
cre8: Updated to make it mandatory

openid-4-verifiable-presentations-1_0.md
cre8: updated

openid-4-verifiable-presentations-1_0.md
cre8: used type, format value is not relevant

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`verifier_attestations`:
+: OPTIONAL. An array of signed objects used by the Relying Party (Verifier) to convey attested information relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
```
c2bo, line 324:
```suggestion
: OPTIONAL. An array of signed objects used by the Verifier to convey attested information relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
...
+
+    * `type`: REQUIRED. A string that identifies the type of the attestation. Ecosystems SHOULD use collision-resistant identifiers.
```
c2bo, line 326:
```suggestion
    * `type`: REQUIRED. A string that identifies the type of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `data`: REQUIRED. A signed object (e.g. a JWT), whose payload structure is defined by the ecosystem or profile. The Wallet MUST validate this signature and ensure binding.
+    * `credential_ids`: OPTIONAL. Array of strings each referencing a Credential requested by the Verifier for which the attestation is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted the attestation is relevant to all requested credentials.
```
c2bo, line 328:
```suggestion
    * `credential_ids`: OPTIONAL. An array of strings each referencing a Credential requested by the Verifier for which the attestation is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted, the attestation is relevant to all requested credentials.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`verifier_attestations`:
+: OPTIONAL. An array of signed objects used by the Relying Party (Verifier) to convey attested information relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
+
+    * `type`: REQUIRED. A string that identifies the type of the attestation. Ecosystems SHOULD use collision-resistant identifiers.
+    * `data`: REQUIRED. A signed object (e.g. a JWT), whose payload structure is defined by the ecosystem or profile. The Wallet MUST validate this signature and ensure binding.
```
c2bo, line 327:
```suggestion
    * `data`: REQUIRED. An object or string representing an attestation (e.g. a JWT). The payload structure is defined on a per type level. The Wallet MUST validate this signature and ensure binding.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+## Verifier Attestations {#verifier-attestations}
+
+Attachments allow the Verifier to provide additional context or metadata as part of the Authorization Request. These inputs can support a variety of use cases, such as helping the Wallet apply policy decisions, validating eligibility, or presenting more meaningful information to the End-User during consent.
+
+Each attachment is an object containing a type identifier and associated data. The format and semantics of these objects are intentionally flexible and defined by ecosystem or profile-specific rules.
```
c2bo, line 637:
```suggestion
Each Verifier Attestation is an object containing a type identifier, associated data and optionally references to credential ids. The format and semantics of these attestations are defined by ecosystems or profiles.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+## Verifier Attestations {#verifier-attestations}
+
+Attachments allow the Verifier to provide additional context or metadata as part of the Authorization Request. These inputs can support a variety of use cases, such as helping the Wallet apply policy decisions, validating eligibility, or presenting more meaningful information to the End-User during consent.
```
c2bo, line 635:
```suggestion
Verifier Attestations allow the Verifier to provide additional context or metadata as part of the Authorization Request. These inputs can support a variety of use cases, such as helping the Wallet apply policy decisions, validating eligibility, or presenting more meaningful information to the End-User during consent.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+Each attachment is an object containing a type identifier and associated data. The format and semantics of these objects are intentionally flexible and defined by ecosystem or profile-specific rules.
+
+For example, a Verifier might include:
+
+- A **registration certificate** issued by a trusted authority, to prove that it is authorized to request certain credentials.
+- An **attestation of intended use**, declaring why a particular credential is being requested.
+- A **policy statement**, such as a signed document describing acceptable use, retention periods, or access rights.
+
+Attachments are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they MUST ignore any unrecognized or unsupported attachment types.
```
c2bo, line 645:
```suggestion
Verifier Attestations are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Attestation types.
```

c2bo (commented):
Not sure about the proof of possession part tbh

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2950,12 +3000,13 @@ The technology described in this specification was made available from contribut
...
 
-   * renamed "Client ID Scheme" to "Client ID Prefix", and updated metadata (`client_id_prefixes_supported`) and `error_description` to match
+   * renamed "Client ID Scheme" to "Client ID Prefix", and updated metadata (`client_id_prefixes_supported`) and `error_description` to match  
+   * add `verifier_attestations` to list of authorization parameters
    * remove DIF Presentation Exchange as a query language option
    * Changes in the DCQL query parameters specific to W3C VCs and AnonCreds
 
    -25
-
+
```
cre8, line 3009:
```suggestion

```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`verifier_attestations`:
+: OPTIONAL. An array of signed objects used by the Verifier to convey attested information relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
```
tlodderstedt, line 324:
```suggestion
: OPTIONAL. An array of attestations about the Verifier relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
...
+
+    * `type`: REQUIRED. A string that identifies the type of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers.
```
tlodderstedt, line 326: I miss the format/type differentiation. The way it is defined now, the type determines both (which is different from all other places in OID4VC)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `data`: REQUIRED. An object or string representing an attestation (e.g. a JWT). The payload structure is defined on a per type level. The Wallet MUST validate this signature and ensure binding.
```
tlodderstedt, line 327:
```suggestion
    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per type level. The Wallet MUST validate this signature and ensure binding.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
 If the Verifier responds with any HTTP error response, the Wallet MUST terminate the process.
 
+## Verifier Attestations {#verifier-attestations}
+
+Verifier Attestations allow the Verifier to provide additional context or metadata as part of the Authorization Request. These inputs can support a variety of use cases, such as helping the Wallet apply policy decisions, validating eligibility, or presenting more meaningful information to the End-User during consent.
```
tlodderstedt, line 635:
```suggestion
Verifier Attestations allow the Verifier to provide additional context or metadata as part of the Authorization Request attested by a trusted third party. These inputs can support a variety of use cases, such as helping the Wallet apply policy decisions, validating eligibility, or presenting more meaningful information to the End-User during consent.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+Each Verifier Attestation is an object containing a type identifier, associated data and optionally references to credential ids. The format and semantics of these attestations are defined by ecosystems or profiles.
+
+For example, a Verifier might include:
+
+- A **registration certificate** issued by a trusted authority, to prove that it is authorized to request certain credentials.
```
tlodderstedt, line 641:
```suggestion
- A **registration certificate** issued by a trusted authority, to prove that the verifier has publicly registered its intend to request certain credentials.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+- An **attestation of intended use**, declaring why a particular credential is being requested.
```
tlodderstedt, line 642: difference to first bullet?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+- A **policy statement**, such as a signed document describing acceptable use, retention periods, or access rights.
```
tlodderstedt, line 643: - the conformation of a role of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+Verifier Attestations are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Attestation types.
+
+### Proof of Possession
+
+This specification supports two models for proof of possession:
+
+- **claim-bound attestations**: The attestation is not signed by the Relying Party, but bound to it. The exact binding mechanism is defined by the type of the definition. For example for JWTs, the `sub` claim is including the distinguished name of the Certificate that was used to sign the request. The binding may also include the client_id parameter.
+- **key-bound attestations**: The attestation is signed by the Relying Party and bound to it. To bind the attestation to the presentation request, it should include the `nonce` and `client_id` parameters in the signature.
```
tlodderstedt, line 652:
```suggestion
- **key-bound attestations**: The attestation is signed by the Relying Party with a key contained or related to the attestation . To bind the signature to the presentation request, the respective signature object should include the `nonce` and `client_id` request parameters.
```

tlodderstedt (changes requested):

openid-4-verifiable-presentations-1_0.md
cre8: I don't think that we need the `format` field like it is used in DCQL. Each attestation object hast the three mentioned fields, but the value of `type` will help you to understand the value of data and how to process it. E.g. when the value is `eudi_registration_certificate` and it's a known value to the Wallet, it will know that the `data` field of type string, including a JWT with a known structure.
There was no demand from others that a type of attestation can be passed in different formats (like JWT or JWS JSON serialization). If this is a demand, the eco system should include the encoding information into the value of `type` like `eudi_registration_certificate+jwt`.

openid-4-verifiable-presentations-1_0.md
cre8: It's not a clear different. I would remove the first point since the second one is more generic.

openid-4-verifiable-presentations-1_0.md
cre8: Are you suggesting to add this as an example?

cre8, line 643:
```suggestion
- A **policy statement**, such as a signed document describing acceptable use, retention periods, or access rights.
- The **conformation of a role** of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.
```

openid-4-verifiable-presentations-1_0.md
cre8:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+Verifier Attestations allow the Verifier to provide additional context or metadata as part of the Authorization Request attested by a trusted third party. These inputs can support a variety of use cases, such as helping the Wallet apply policy decisions, validating eligibility, or presenting more meaningful information to the End-User during consent.
+
+Each Verifier Attestation is an object containing a type identifier, associated data and optionally references to credential ids. The format and semantics of these attestations are defined by ecosystems or profiles.
+
+For example, a Verifier might include:
+
+- A **registration certificate** issued by a trusted authority, to prove that the verifier has publicly registered its intend to request certain credentials.
+- A **policy statement**, such as a signed document describing acceptable use, retention periods, or access rights.
+- The **conformation of a role** of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.
```
tlodderstedt, line 643:
```suggestion
- The **confirmation of a role** of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.
```

tlodderstedt (changes requested):

openid-4-verifiable-presentations-1_0.md
danielfett: I think Torsten's suggestion makes sense - use a type to enable decoding, then use format-specific mechanisms to transport the purpose of the attestation.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`verifier_attestations`:
+: OPTIONAL. An array of attestations about the Verifier relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
+
+    * `type`: REQUIRED. A string that identifies the type of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers.
```
tlodderstedt, line 326:
```suggestion
    * `format`: REQUIRED. A string that identifies the format of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers. Further processing of the attestation is determined by the type of the attestation, which is specified in a format-specific way. 
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per type level. The Wallet MUST validate this signature and ensure binding.
```
tlodderstedt, line 327:
```suggestion
    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. The Wallet MUST validate this signature and ensure binding.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
...
+    * `credential_ids`: OPTIONAL. An array of strings each referencing a Credential requested by the Verifier for which the attestation is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted, the attestation is relevant to all requested credentials.
+
+See (#verifier-attestations) for more details.
+
+The following is a non-normative example of an attested object:
+
+```json
+{
+  "type": "eudi_registration_certificate",
```
tlodderstedt, line 336:
```suggestion
  "format": "jwt",
```

tlodderstedt (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+- The **confirmation of a role** of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.
+
+Verifier Attestations are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Attestation types.
+
+### Proof of Possession
+
+This specification supports two models for proof of possession:
+
+- **claim-bound attestations**: The attestation is not signed by the Relying Party, but bound to it. The exact binding mechanism is defined by the type of the definition. For example for JWTs, the `sub` claim is including the distinguished name of the Certificate that was used to sign the request. The binding may also include the client_id parameter.
+- **key-bound attestations**: The attestation is signed by the Relying Party with a key contained or related to the attestation . To bind the signature to the presentation request, the respective signature object should include the `nonce` and `client_id` request parameters.
```
cre8, line 652:
```suggestion
- **key-bound attestations**: The attestation's proof of possession is signed by the Relying Party with a key contained or related to the attestation . To bind the signature to the presentation request, the respective signature object should include the `nonce` and `client_id` request parameters.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+- The **conformation of a role** of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.
+
+Verifier Attestations are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Attestation types.
+
+### Proof of Possession
+
+This specification supports two models for proof of possession:
+
+- **claim-bound attestations**: The attestation is not signed by the Relying Party, but bound to it. The exact binding mechanism is defined by the type of the definition. For example for JWTs, the `sub` claim is including the distinguished name of the Certificate that was used to sign the request. The binding may also include the client_id parameter.
+- **key-bound attestations**: The attestation is signed by the Relying Party with a key contained or related to the attestation . To bind the signature to the presentation request, the respective signature object should include the `nonce` and `client_id` request parameters.
```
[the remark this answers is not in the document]
cre8: yes, I made a suggestion

openid-4-verifiable-presentations-1_0.md
cre8: We agreed it's better to define the format for decoding than the type. This has the disadvantage that the Wallet needs to decode all attestations to get to type (decoding the sd jwt vc and then looking at the vct value). Since this is only decoding and not validating, the required process is affordable. This reduces the risk that there is a difference between an attribute on the root level of the attestation and the actual passed object

tlodderstedt (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,6 +320,25 @@ The following is a non-normative example of a transaction data content, after ba
 }
 ```
 
+`verifier_attestations`:
+: OPTIONAL. An array of attestations about the Verifier relevant to the Credential Request. These attestations MAY include Verifier metadata, policies, trust status, or authorizations. Attestations are intended to support authorization decisions, inform Wallet policy enforcement, or enrich the End-User consent dialog. Each object has the following structure:
+
+    * `format`: REQUIRED. A string that identifies the format of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers. Further processing of the attestation is determined by the type of the attestation, which is specified in a format-specific way. 
+    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. The Wallet MUST validate this signature and ensure binding.
+    * `credential_ids`: OPTIONAL. An array of strings each referencing a Credential requested by the Verifier for which the attestation is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted, the attestation is relevant to all requested credentials.
```
c2bo, line 328: feels a little cleaner to me:
```suggestion
    * `credential_ids`: OPTIONAL. An array of strings each referencing a Credential requested by the Verifier for which the attestation is relevant. Each string matches the `id` field in a DCQL Credential Query. If omitted, the attestation is relevant to the whole request.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -611,6 +630,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+
+Verifier Attestations are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Attestation types.
+
+### Proof of Possession
+
+This specification supports two models for proof of possession:
+
+- **claim-bound attestations**: The attestation is not signed by the Relying Party, but bound to it. The exact binding mechanism is defined by the type of the definition. For example for JWTs, the `sub` claim is including the distinguished name of the Certificate that was used to sign the request. The binding may also include the client_id parameter.
+- **key-bound attestations**: The attestation's proof of possession is signed by the Relying Party with a key contained or related to the attestation . To bind the signature to the presentation request, the respective signature object should include the `nonce` and `client_id` request parameters. The attestation and the proof of possession have to be passed in the attachment.
```
c2bo, line 652:
```suggestion
- **key-bound attestations**: The attestation's proof of possession is signed by the Relying Party with a key contained or related to the attestation. To bind the proof of possession to the presentation request, it should include the `nonce` and `client_id` request parameters. The attestation and the proof of possession have to be passed in the attachment.
```

c2bo (approved):
I am not really happy with wallets having to parse everything first (to discover type) before understanding if it's relevant, but overall this variant seems to achieve the requirements

openid-4-verifiable-presentations-1_0.md
cre8: I think this should be as it is. The attestation is limited to be used only for credentials and could be abused for other stuff when extending it to the whole request.

cre8:
> I am not really happy with wallets having to parse everything first (to discover type) before understanding if it's relevant, but overall this variant seems to achieve the requirements

Me neither, but since it's only parsing and not validating, it should be fine. If there occur attestations where it becomes more complex, we can add an optional attribute that can be used as an identifier. But at this point we were not able to find a good solution that pleased everyone.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -609,6 +628,29 @@ The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
...
+- A **policy statement**, such as a signed document describing acceptable use, retention periods, or access rights.
+- The **confirmation of a role** of the verifier in a certain domain, e.g. the verifier might be a certified payment service provider under the EU's Payment Service Directive 2.
+
+Verifier Attestations are optional. Wallets MAY use them to make authorization decisions or to enhance the user experience, but they SHOULD ignore any unrecognized or unsupported Verifier Attestation types.
+
+### Proof of Possession
+
+This specification supports two models for proof of possession:
+
+- **claim-bound attestations**: The attestation is not signed by the Relying Party, but bound to it. The exact binding mechanism is defined by the type of the definition. For example for JWTs, the `sub` claim is including the distinguished name of the Certificate that was used to sign the request. The binding may also include the client_id parameter.
```
c2bo, line 649:
```suggestion
- **claim-bound attestations**: The attestation is not signed by the Verifier, but bound to it. The exact binding mechanism is defined by the type of the definition. For example for JWTs, the `sub` claim is including the distinguished name of the Certificate that was used to sign the request. The binding may also include the client_id parameter.
```

PR #483, opened 2025-04-01
Clarify key selection for client_metadata jwks

deshmukhrajvardhan:
## PR Checkpoints
- [x] official html rendering, which is generated using mmark - Download it from the GitHub action tab, https://github.com/openid/OpenID4VP/pull/<PR number>/checks - by clicking on 'artifacts' and then 'output'
- [ ] >=4 approvals OR >=2 if editorial PR
- [ ] >=1 week since PR's final state
- [ ] Only Chairs can merge the PR

## Summary
Clarify key selection for client_metadata jwks like described in FAPI: https://openid.net/specs/openid-financial-api-part-2-1_0.html#duplicate-key-identifiers

## Related Issue(s)
resolves https://github.com/openid/OpenID4VP/issues/438

## Special notes for your reviewer

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -330,13 +330,8 @@ The following is a non-normative example of a transaction data content, after ba
   // other transaction data type specific parameters
 }
 ```
-### Key selection for client_metadata jwks {#client_metadata_key_selection}
-JWK sets should not contain multiple keys with the same kid. However, to increase interoperability when there are multiple keys with the same kid, the verifier shall consider other JWK attributes, such as kty, use, alg, etc., when selecting the verification key for the particular JWS message. For example, the following algorithm could be used in selecting which key to use to verify a message signature:
-
-
-1. find keys with a kid that matches the kid in the JOSE header;
-2. if a single key is found, use that key;
-3. if multiple keys are found, then the verifier should iterate through the keys until a key is found that has a matching alg, use, kty, or crv that corresponds to the message being verified.
+### Key Selection for `jwks` Client Metadata Parameter {#client_metadata_key_selection}
+JWK sets SHOULD NOT contain multiple keys with the same `kid`. However, when there are multiple keys with the same `kid`, the verifier shall consider other JWK attributes, such as `kty`, `use`, `alg`, etc., when selecting the verification key for the particular JWS message.
 
 ## Existing Parameters
 
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -295,7 +295,7 @@ In the context of an authorization request according to [@RFC6749], parameters c
 `client_metadata`:
 : OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -330,6 +330,13 @@ The following is a non-normative example of a transaction data content, after ba
   // other transaction data type specific parameters
 }
 ```
+### Key selection for client_metadata jwks {#client_metadata_key_selection}
+JWK sets should not contain multiple keys with the same kid. However, to increase interoperability when there are multiple keys with the same kid, the verifier shall consider other JWK attributes, such as kty, use, alg, etc., when selecting the verification key for the particular JWS message. For example, the following algorithm could be used in selecting which key to use to verify a message signature:
+
+
+1. find keys with a kid that matches the kid in the JOSE header;
+2. if a single key is found, use that key;
+3. if multiple keys are found, then the verifier should iterate through the keys until a key is found that has a matching alg, use, kty, or crv that corresponds to the message being verified.
```
Sakurann, line 339:
```suggestion

### Key Selection for `jwks` Client Metadata Parameter {#client_metadata_key_selection}

JWK sets SHOULD NOT contain multiple keys with the same `kid`. However, when there are multiple keys with the same `kid`, the verifier shall consider other JWK attributes, such as `kty`, `use`, `alg`, etc., when selecting the verification key for a particular JWS message. ```

I know this is the exact text from [FAPI](https://openid.net/specs/openid-financial-api-part-2-1_0.html#duplicate-key-identifiers), but I feel like 1 ~3 currently repeats what is already described in the text, no? just a suggestion

Comment (deshmukhrajvardhan) [reply openid-4-verifiable-presentations-1_0.md:339]: I agree, it would be good to remove steps 1-3 as the text describes it well enough.

**Review by charsleysa — APPROVED** [review]

**Comment by bc-pi** [comment]
Honestly not sure this works well with anticipated changes from  #477

**Review by Sakurann — APPROVED** [review]

File: openid-4-verifiable-presentations-1_0.md [hunk 2]
```diff
@@ -330,6 +330,8 @@ The following is a non-normative example of a transaction data content, after ba
   // other transaction data type specific parameters
 }
 ```
+### Key Selection for `jwks` Client Metadata Parameter {#client_metadata_key_selection}
+JWK sets SHOULD NOT contain multiple keys with the same `kid`. However, when there are multiple keys with the same `kid`, the verifier shall consider other JWK attributes, such as `kty`, `use`, `alg`, etc., when selecting the verification key for the particular JWS message.
```
Sakurann, line 334:
```suggestion

### Key Selection for `jwks` Client Metadata Parameter {#client_metadata_key_selection}

JWK sets SHOULD NOT contain multiple keys with the same `kid`. However, when there are multiple keys with the same `kid`, the verifier shall consider other JWK attributes, such as `kty`, `use`, `alg`, etc., when selecting the verification key for the particular JWS message.
```

Sakurann:
@bc-pi umm this is based on the feedback from the implementers who have implemented only response encryption (no JARM kind of signing at all) so i still think this PR is valuable regardless of #477

bc-pi:
> @bc-pi umm this is based on the feedback from the implementers who have implemented only response encryption (no JARM kind of signing at all) so i still think this PR is valuable regardless of #477

I think I was largly reacting to the text (now gone) that was borrowed from FAPI. But the "selecting the verification key for the particular JWS message" in this PR is in a new little section that is refrenced only from the context of the client_metadata/jwks param where it says "Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests".  

Admittedly I don't know specifically what that feedback from the implementers was but the text introduced here is incorrect so probably doesn't meaningfully address it.

bc-pi (changes requested):
see prior comments https://github.com/openid/OpenID4VP/pull/483#issuecomment-2783815808 & https://github.com/openid/OpenID4VP/pull/483#issuecomment-2775499923

PR #485, opened 2025-04-01
note on pre-final specs

### LINKED ISSUE #409: add compatibility note about pre-final specifications
Multiple non-final specifications are referenced in OpenID4VP. we should add a note about their versions.

Th precedent is OpenID Connect Core final placing such compatibility note in the spec at https://openid.net/specs/openid-connect-core-1_0-final.html#PreFinalIETFSpecs :

> Implementers should be aware that this specification uses several IETF specifications that are not yet final specifications. Those specifications are:
> 
> JSON Web Token (JWT) draft -13 [JWT]
> JSON Web Signature (JWS) draft -18 [JWS]
> JSON Web Encryption (JWE) draft -18 [JWE]
> JSON Web Key (JWK) draft -18 [JWK]
> JSON Web Algorithms draft -18 [JWA]
> Assertion Framework for OAuth 2.0 Client Authentication and Authorization Grants draft -13 [OAuth.Assertions]
> JSON Web Token (JWT) Profile for OAuth 2.0 Client Authentication and Authorization Grants draft -07 [OAuth.JWT]
> 
> While every effort will be made to prevent breaking changes to these specifications, should they occur, OpenID Connect implementations should continue to use the specifically referenced draft versions above in preference to the final versions, unless using a possible future OpenID Connect profile or specification that updates some or all of these references.

c2bo:
Closes #409 

Not entirely sure about this one. Especially the text about things like w3c DC API was written in a way where there should be no breaking changes that impact this specification.

Also: Is there a stable url to reference for the current version of DC API?


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1704,6 +1704,17 @@ Note: If the Verifier's Response URI did not return a `redirect_uri` in step (6)
...
+## Pre-Final Specifications
+
+Implementers should be aware that this specification uses several specifications that are not yet final specifications. Those specifications are:
+
+- OpenID Federation 1.0 draft -42 [@!OpenID.Federation]
+- SIOPv2 draft -13 [@!SIOPv2]
+- SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -08 [I-D.ietf-oauth-sd-jwt-vc]
+- W3C Digital Credentials API - [@!W3C.Digital_Credentials_API]
+
+While every effort will be made to prevent breaking changes to these specifications, should they occur, OpenID4VP implementations should continue to use the specifically referenced draft versions above in preference to the final versions, unless using a possible future OpenID Connect profile or specification that updates some or all of these references.
```
danielfett, line 1716:
```suggestion
While breaking changes to these specifications are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced draft versions above in preference to the final versions, unless updated by a profile or new version of this specification.
```

danielfett (approved):

selfissued:
@c2bo Can you please update the date in the OpenID Federation reference to "5 March 2025", matching the draft 42 date?

Thanks.

selfissued (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1704,6 +1704,17 @@ Note: If the Verifier's Response URI did not return a `redirect_uri` in step (6)
 
 (10) The Verifier checks whether the `nonce` received in the Credential(s) in the VP Token in step (9) corresponds to the `nonce` value from the session. The Verifier then consumes the VP Token and invalidates the `transaction-id`, `request-id` and `nonce` in the session.
 
+## Pre-Final Specifications
+
+Implementers should be aware that this specification uses several specifications that are not yet final specifications. Those specifications are:
+
+- OpenID Federation 1.0 draft -42 [@!OpenID.Federation]
+- SIOPv2 draft -13 [@!SIOPv2]
+- SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -08 [I-D.ietf-oauth-sd-jwt-vc]
```
jogu, line 1713: Should a specific version of SD-JWT be listed too? I don't think SD-JWT VC refers to a specific draft.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1704,6 +1704,17 @@ Note: If the Verifier's Response URI did not return a `redirect_uri` in step (6)
...
+- W3C Digital Credentials API - [@!W3C.Digital_Credentials_API]
```
jogu, line 1714: This is a slightly odd one - I'm not sure the "only use the specifically referenced draft versions" really applies here as really your only option is to use whatever the browsers/OS actually implement?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1704,6 +1704,17 @@ Note: If the Verifier's Response URI did not return a `redirect_uri` in step (6)
 
 (10) The Verifier checks whether the `nonce` received in the Credential(s) in the VP Token in step (9) corresponds to the `nonce` value from the session. The Verifier then consumes the VP Token and invalidates the `transaction-id`, `request-id` and `nonce` in the session.
 
+## Pre-Final Specifications
+
+Implementers should be aware that this specification uses several specifications that are not yet final specifications. Those specifications are:
+
+- OpenID Federation 1.0 draft -42 [@!OpenID.Federation]
+- SIOPv2 draft -13 [@!SIOPv2]
```
jogu, line 1712: If we mean draft 13, shouldn't the normative reference point to draft 13? i.e. https://openid.net/specs/openid-connect-self-issued-v2-1_0-13.html (the contents of which will never change)

Whereas it currently points to: https://openid.net/specs/openid-connect-self-issued-v2-1_0.html - the contents of which can change.

(I guess this comment may apply to other references too, I didn't check)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1704,6 +1704,17 @@ Note: If the Verifier's Response URI did not return a `redirect_uri` in step (6)
...
+- SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -08 [I-D.ietf-oauth-sd-jwt-vc]
+- W3C Digital Credentials API - [@!W3C.Digital_Credentials_API]
+
+While breaking changes to these specifications are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced draft versions above in preference to the final versions, unless updated by a profile or new version of this specification.
```
jogu, line 1716: It might be worth clarifying whether this applies to non-draft versions too? e.g. we reference ISO/IEC 18013-5:2021 - is that the only one you can use? (Whereas for the BCP195 reference I'm relatively sure we definitely would expect people to follow the latest version.)

openid-4-verifiable-presentations-1_0.md
Sakurann: good point

openid-4-verifiable-presentations-1_0.md
Sakurann: can we do that in a different PR? the purpose of this PR is to clarify draft specs.

jogu: I guess we could.

openid-4-verifiable-presentations-1_0.md
jogu: So we're saying that for SD-JWT we *do* want people to use the latest rather than a particular numbered draft? That seems strange to me. I think we should explicitly state a version for SD-JWT too.

c2bo: I wondered about that as well, but openid connect core spec worked the same way: The reference was to latest (stable url to latest), but the note then specified specific draft versions with this note.

--> This is why i only adjusted the text here and left the references to latest.

openid-4-verifiable-presentations-1_0.md
c2bo: I wondered about that as well, but openid connect core spec worked the same way: The reference was to latest (stable url to latest), but the note then specified specific draft versions with this note.

--> This is why i only adjusted the text here and left the references to latest.

openid-4-verifiable-presentations-1_0.md
Sakurann: we do not reference sd-jwt spec from openid4vp, only sd-jwt vc.
plus, this is what i meant - no need for a draft version for sd-jwt vc because ietf tooling will give the latest one:
```suggestion
- SD-JWT-based Verifiable Credentials (SD-JWT VC) draft [@!I-D.ietf-oauth-sd-jwt-vc]
```

openid-4-verifiable-presentations-1_0.md
Sakurann: i think that's confusing and we should do better than OIDC?

openid-4-verifiable-presentations-1_0.md
Sakurann: @leecam do you have an opinion here?

openid-4-verifiable-presentations-1_0.md
c2bo: I would still provide that draft version in the section explicitly if you don't mind - seems cleaner to me to have that in text for all non-final specs

openid-4-verifiable-presentations-1_0.md
c2bo: Yeah DC API felt really weird to me as well and there isn't really a specific draft to reference?

openid-4-verifiable-presentations-1_0.md
c2bo: I would assume the intention was to reference the latest draft for the likely case that there are no breaking changes -> then it would be better to have a reference to the final specification.

I am happy to change this to whatever we decide if there is consensus which way to go

openid-4-verifiable-presentations-1_0.md
jogu: I thought the point was to refer to explicit versions, as Christian had done, so that if breaking changes are made it's clear which one should be followed to be conformant.

If that is the case I think we need to specify explicit versions of both SD-JWT VC and SD-JWT to achieve the stated aim.

(Although equally it is going to be confusing if we're saying "use draft 8" alongside a link to (say) draft 9. I guess that was Kristina's point...)

openid-4-verifiable-presentations-1_0.md
jogu: > I wondered about that as well, but openid connect core spec worked the same way: The reference was to latest (stable url to latest), but the note then specified specific draft versions with this note.

The OIDC link points to 1.0 final. The only change that might happen to that are errata which should be non-breaking. It's up to us what we point to though, if we want to point to an explicit revision of OIDC (e.g. 1.0 errata 3) then we could.

Sakurann: wg discussion:
- use URL to the version that is being referenced

openid-4-verifiable-presentations-1_0.md
Sakurann: WG discussion remove this from here

openid-4-verifiable-presentations-1_0.md
Sakurann: WG discussion: 
make clear 
- a specific version to be used with openid4vp final.
- allow using updated versions of already published specifications if there is no breaking changes
- do an errata when an updated version of these specs are published
 
```suggestion
While breaking changes to these specifications are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced versions above in preference to the final versions, unless updated by a profile or new version of this specification.
```

openid-4-verifiable-presentations-1_0.md
c2bo:
```suggestion
```

openid-4-verifiable-presentations-1_0.md
c2bo: > wg discussion:
> 
> * use URL to the version that is being referenced

I think we reached a different consensus at the end of the discussion?

Sakurann: i don't think so

Sakurann (approved):

TimoGlastra:
https://github.com/openid/OpenID4VP/pull/500 needs to be synced with this PR once either this PR or #500 is merged.

I added a reference to https://datatracker.ietf.org/doc/draft-ietf-jose-fully-specified-algorithms/

c2bo:
I'll update this PR once yours is merged

openid-4-verifiable-presentations-1_0.md
c2bo: changed to fixed versions for drafts

Issue #455, opened 2025-03-13
Clarify alg_values_supported parameter


andprian:
Clause  [10.1. ](https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#section-10.1)[Additional Wallet Metadata Parameters](https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#name-additional-wallet-metadata-) states:

> alg_values_supported: OPTIONAL. An object where the value is an array of case sensitive strings that identify the cryptographic suites that are supported.

Clarify for which cryptographic operations we need to specify the cryptographic suites within `alg_values_supported`.

Sakurann:
#500 merged

PR #488, opened 2025-04-01
Scvenema/455/clarify alg values supported parameter

scvenema:
Clarify the `alg_values_supported` parameter per Issue #455


charsleysa (approved):

bc-pi (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1478,7 +1478,7 @@ This specification defines new metadata parameters according to [@!RFC8414].
...
-    * `alg_values_supported`: OPTIONAL. An object where the value is an array of case sensitive strings that identify the cryptographic suites that are supported. Parties will need to agree upon the meanings of the values used, which may be context-specific. For specific values that can be used depending on the Credential format, see (#format_specific_parameters). If `alg_values_supported` is omitted, it is unknown what cryptographic suites the wallet supports.
+    * `alg_values_supported`: OPTIONAL. An object whose value is an array of case sensitive strings that identify the signature algorithms that are supported for wallet-signed presentations. Parties will need to agree upon the meanings of the values used, which may be context-specific. For specific values that can be used depending on the Credential format, see (#format_specific_parameters). If `alg_values_supported` is omitted, it is unknown what cryptographic suites the wallet supports.
```
TimoGlastra, line 1481: Shouldn't the definition of this just follow the format-specific definition for Verifier Metadata?

#479 adds a Verifier Metadata definition for W3C vcs `proof_type_values`.  I guess we can just reuse that definition of this structure as well? I also opened #495, which is related to this.

TimoGlastra: So instead of calling it Verifier Metadata, we call it Metadata, and we define the `vp_formats` / `vp_formats_supported` structure. I see no reason for those to be separate definitions really.

So e.g. B.4.3 becomes:


> The `format` value in the `vp_formats` parameter in the Verifier metadata, or the `vp_formats_supported` in the Wallet metadata MUST have the key `dc+sd-jwt`, and the value is an object consisting of the following name/value pairs:
>
> * `sd-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms supported for an Issuer-signed JWT of an SD-JWT. If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the Issuer-signed JWT of the presented SD-JWT MUST match one of the array values.
> * `kb-jwt_alg_values`: OPTIONAL. A JSON array containing identifiers of cryptographic algorithms supported for a Key Binding JWT (KB-JWT). If present, the `alg` JOSE header (as defined in [@!RFC7515]) of the presented KB-JWT MUST match one of the array values.
> 
> The following is a non-normative example of `client_metadata` request parameter value in a request to present an IETF SD-JWT VC.

TimoGlastra: I think that is also what @Sakurann commented in https://github.com/openid/OpenID4VP/issues/455#issuecomment-2739683436, but i don't think this PR clarifies that. Since the format specific appendixes specifically refers to Verifier Metadata and what the verifier supports. Not what the wallet supports.

TimoGlastra: My suggestion would be for both `vp_formats` and `vp_formats_supported` have no default parameters. An empty object indicates support for the format. Additional parameters can be specified in the format specific registry (such as `alg_values_supported` for `dc+sd-jwt` and `jwt_vc_json`. For `ldp_vc` we can define `proof_types_supported` along with `cryptosuites_supported`. For anoncreds we don't need additional metadata, for mdoc it should be COSE alg values.

Sakurann (changes requested):
it's probably because i did not document it clearly enough in the issue but my expectations from this PR was closer to what Timo describes #495

Sakurann:
superceded by #500. will close once that one is merged

scvenema:
I'm closing this PR as obsoleted by PR #500 
However, the parent Issue #455 may still be relevant however. I'll add a new comment to that issue in a moment. (@Sakurann  )

PR #492, opened 2025-04-02
Change processing rules to allow same credential fulfilling different queries

### LINKED ISSUE #397: same credential fulfilling multiple credential queries

Sakurann:
resolves #397 

- adds a processing rule that `When the same Credential fulfills more than one Credential Query in a request, that Credential MUST contain the combination of claims requested by those Credential Queries. However, that Credential MUST NOT contain claims not requested by either of the Credential Queries.`
    - I added it in a place that in my opinion made it most clear that this transcends claims and credential selection, but it can also be moved to after [line 966](https://github.com/openid/OpenID4VP/compare/same-cred-diff-query?expand=1#diff-3118c9756a1d7b361bb53af8bd9d65666476a27cfeeced56c47a5dccc38eac55L966) 

editorial changes
- consolidated security considerations section in one place (already using client identifier prefix term)
- some whitespace nits


.DS_Store
charsleysa, line 1: shouldn't commit this file

openid-4-verifiable-presentations-1_0.md
```diff
@@ -914,12 +910,16 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
...
-### Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
+## Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
 
 The following section describes the logic that applies for selecting claims 
-and for selecting credentials. 
+and for selecting credentials.
 
-#### Selecting Claims {#selecting_claims}
+When the same Credential fulfills more than one Credential Query in a request, that Credential MUST contain
+the combination of claims requested by those Credential Queries. However, that Credential MUST NOT contain
```
paulbastian, line 919: Doesn't this disallow the simple option to make two distinct presentations for the same credential?

paulbastian (changes requested):

openid-4-verifiable-presentations-1_0.md
GarethCOliver: I guess the question would be 'what defines a credential'? Would it be more accurate to talk about this in terms of presentations, rather than credentials:

I.e. 

'When the same Credential Presentation fulfills more than one Credential Query in a request, that Credential Presentation MUST contain the combination of claims requested by those Credential Queries. However, that Credential Presentation MUST NOT contain claims not requested by any of the Credential Queries that it fulfills."

That would allow the wallet to make two distinct or a single one.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -914,12 +910,16 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
...
+claims not requested by any of the Credential Queries.
```
Sakurann, line 920:
```suggestion
When the same Credential fulfills more than one Credential Query in a request, that Credential MAY contain
the combination of claims requested by those Credential Queries. However, in this case, the Credential MUST NOT contain
claims not requested by any of the Credential Queries.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: good catch, Paul. I was not sure if we wanted to allow "the simple option to make two distinct presentations for the same credential". if we do, Verifier would need to support both cases, which might be harder?

paulbastian: Yes, but I didn't see why we should make wallet implementations harder either. Most verified implementations may not even check if more data than necessary was delivered

GarethCOliver: I don't think it's actually more work for the verifier:

AIUI it is always valid in these cases to answer the different credential queries with *different* credentials (I don't believe we even have a way to express that two different credential queries *must* be the same credential, there are arguably use-cases but they are pretty thin).

As a consequence, verifiers must be able to support different credentials presented for the different credential queries (at least to be correct). If they support that, then they will naturally support @paulbastian's desired mechanism (as it is  likely indistinguishable from a verifiers perspective). So being permissive for wallets lowers their burden at no interop cost, so seems like win.

(Verifier processing logic is really just: for each queryId  verify credential, extract relevant claims from credential to internal format. All this text changes is to prevent Verifiers being overly strict (ones who want to be now have to do some extra work to de-dupe credentials before running their check for unexpected claims, but for those who don't it is no more work.))

Sakurann: > As a consequence, verifiers must be able to support different credentials presented for the different credential queries (at least to be correct). If they support that, then they will naturally support @paulbastian's desired mechanism 

that is a fair point.

I'll commit my suggestion below that I believe fulfills @paulbastian's original point?

.DS_Store
charsleysa, line 1: This file should be removed from the PR

charsleysa (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -914,12 +910,16 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
...
+## Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
 
 The following section describes the logic that applies for selecting claims 
-and for selecting credentials. 
+and for selecting credentials.
 
-#### Selecting Claims {#selecting_claims}
+When the same Credential fulfills more than one Credential Query in a request, that Credential MAY contain
+the combination of claims requested by those Credential Queries. However, in this case, the Credential MUST NOT contain
+claims not requested by any of the Credential Queries.
```
danielfett, line 920: This should go below "selecting claims".

danielfett:
This changes a lot of unrelated things, so should probably be rebased and checked again.

openid-4-verifiable-presentations-1_0.md
Sakurann, line 920:
```suggestion
When a Verifiable Credential fulfills more than one Credential Query in a request, the Verifiable Presentation MAY contain
the combination of claims requested by those Credential Queries. However, in this case, the Verifiable Presentation MUST NOT contain
claims not requested by any of the Credential Queries if the Credential Format supports selective disclosure or ZKP.
```

paulbastian: btw, we do not differentiate between Credential and Verifiable Credential in OpenID4VCI, I think the terminology should match here better.

danielfett:
```suggestion
For formats supporting selective disclosure, these rules support selecting a minimal dataset to fulfill the Verifier's request in a privacy-friendly manner (see (#privacy-considerations) for additional considerations). Wallets MUST NOT send selectively disclosable claims that have not been selected according to the rules below. A presentation of a credential MAY contain more than the claims selected in the particular DCQL Credential Query if the same credential is selected with the additional claims in a separate Credential Query in the same request.
```

NOTE: The anchor in privacy considerations needs to be added!

martijnharing: The proposal from Daniel makes sense to me.

martijnharing (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -914,12 +910,14 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
...
 
-### Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
+## Selecting Claims and Credentials {#dcql_query_lang_processing_rules}
 
 The following section describes the logic that applies for selecting claims 
-and for selecting credentials. 
+and for selecting credentials.
 
-#### Selecting Claims {#selecting_claims}
+For formats supporting selective disclosure, these rules support selecting a minimal dataset to fulfill the Verifier's request in a privacy-friendly manner (see (#privacy-considerations) for additional considerations). Wallets MUST NOT send selectively disclosable claims that have not been selected according to the rules below. A presentation of a credential MAY contain more than the claims selected in the particular DCQL Credential Query if the same credential is selected with the additional claims in a separate Credential Query in the same request.
```
Sakurann, line 918:
```suggestion
For formats supporting selective disclosure, these rules support selecting a minimal
dataset to fulfill the Verifier's request in a privacy-friendly manner
(see (#privacy-considerations) for additional considerations). Wallets MUST NOT send
selectively disclosable claims that have not been selected according to the rules below. A presentation of a credential MAY contain more than the claims selected in the
particular DCQL Credential Query if the same credential is selected with the additional
claims in a separate Credential Query in the same request, or the additional claims are
not selectively disclosable.
```

paulbastian (approved):

charsleysa (approved):
as long as the .DS_Store file is removed

martijnharing (approved):

GarethCOliver (approved):

c2bo (approved):

PR #496, opened 2025-04-03
Update openid-4-verifiable-presentations-1_0.md

GarethCOliver:
nit: any rather than either (could be more than two)


Sakurann (approved):
thanks

PR #499, opened 2025-04-04
Clarify iss value in Request Object (JAR)

### LINKED ISSUE #299: `iss` value to be used in signed requests is not clear?
Feedback based on testing verifier conformance tests - I think it's not clear what the value for `iss` should be in signed request objects.

JAR only says:

> If signed, the Authorization Request Object SHOULD contain the Claims iss (issuer) and aud (audience) as members with their semantics being the same as defined in the [JWT](https://www.rfc-editor.org/rfc/rfc9101.html#RFC7519) [[RFC7519](https://www.rfc-editor.org/rfc/rfc9101.html#RFC7519)] specification.

But doesn't actually say what value to give to `iss`.

I believe it can only sensibly be client id. That's certainly the approach that ended up being taken in FAPI conformance tests.

c2bo:
Closes #299

Made sense to me to put it next to the introduction of JAR, but I could also create a section 5.10 below the section on aud if that seems more logical to people.


openid-4-verifiable-presentations-1_0.md
```diff
@@ -256,6 +256,7 @@ Presentation of Verifiable Credentials using OpenID for Verifiable Presentations
 The Authorization Request follows the definition given in [@!RFC6749] taking into account the recommendations given in [@!RFC9700].
 
 The Verifier MAY send an Authorization Request as a Request Object either by value or by reference, as defined in the JWT-Secured Authorization Request (JAR) [@RFC9101]. Verifiers MUST include the `typ` Header Parameter in Request Objects with the value `oauth-authz-req+jwt`, as defined in [@RFC9101]. Wallets MUST NOT process Request Objects where the `typ` Header Parameter is not present or does not have the value `oauth-authz-req+jwt`.
+Note that the `client_id` claim is required as defined below and would be redundant with a possible `iss` claim in the Request Object which is commonly used in JAR. To not break existing JAR implementations, the `iss` claim MAY be present in the Request Object, but MUST ignored by the Wallet if present.
```
```suggestion

The `client_id` claim is required as defined below and would be redundant with a possible `iss` claim in the Request Object which is commonly used in JAR. To not break existing JAR implementations, the `iss` claim MAY be present in the Request Object. However, even if it is present, the Wallet MUST ignore it.
```

sticking to one normative MAY/MUST per sentence. and removing that this is a note since it is normative.
I am not a big fan of `as defined below`. might be cleaner to move this after the definition of `client_id` being included?

bc-pi (approved):
seems legit

PR #507, opened 2025-04-04
Describe nonce for AnonCreds

### LINKED ISSUE #489: Fix AnonCreds text
>> The AnonCreds related concerns don't really belong to this PR and we can move that into a separate issue/PR (fixing existing things not related to query language).

>I suppose we need an Issue to track this either way.

_Originally posted by @danielfett in https://github.com/openid/OpenID4VP/issues/479#issuecomment-2769610290_

The AnonCreds credential format specific parts are somewhat broken. We need to at least describe how to generate the nonce used in AnonCreds as a combination of nonce & audience. Straight forward proposal would be to take nonce & audience inputs, concatenate and hash them and take the frist 80 bits of that input as the nonce.

Also references should not point to Indy, but https://hyperledger.github.io/anoncreds-spec/

c2bo:
Closes #489 

This should fix things with the current AnonCreds (1.0) specification to the best of my knowledge. Any feedback from people actually implementing/using AnonCreds according to the current 1.0 specification very welcome.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -52,7 +52,7 @@ This specification defines a protocol for requesting and presenting Verifiable C
 
 # Introduction
 
-This specification defines a mechanism on top of OAuth 2.0 [@!RFC6749] that enables presentation of Verifiable Credentials as Verifiable Presentations. Verifiable Credentials and Verifiable Presentations can be of any format, including, but not limited to W3C Verifiable Credentials Data Model [@VC_DATA], ISO mdoc [@ISO.18013-5], IETF SD-JWT VC [@!I-D.ietf-oauth-sd-jwt-vc], and AnonCreds [@Hyperledger.Indy].
+This specification defines a mechanism on top of OAuth 2.0 [@!RFC6749] that enables presentation of Verifiable Credentials as Verifiable Presentations. Verifiable Credentials and Verifiable Presentations can be of any format, including, but not limited to W3C Verifiable Credentials Data Model [@VC_DATA], ISO mdoc [@ISO.18013-5], IETF SD-JWT VC [@!I-D.ietf-oauth-sd-jwt-vc], and AnonCreds [@Hyperledger.AnonCreds].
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2340,6 +2340,10 @@ The following is a non-normative example of a DCQL request for an AnonCreds Cred
...
+
+To compute to the `nonce` parameter that is used as an input for Prove and Verify operations of AnonCreds, the `nonce` from the Authorization Request must be concatenated with the audience binding (as defined above) and hashed using sha-256. The first 80 bits of that digest are then used as the `nonce` paramter for the AnonCreds proof. This `nonce` MUST then be used as nonce (also called n_1) for the Presentation generation and verification as defined in sections 9.6 and 9.7 of [@Hyperledger.AnonCreds].
```
Sakurann, line 2345:
```suggestion
The AnonCreds Credential format only allows for a `nonce` that has to be exactly 80 bit long, whereas other Credential formats
allow for the different inputs to be signed over in a proof of possession for Cryptographic Holder Binding. For AnonCreds, everything that should be part of
the input to generate that proof MUST be part of the `nonce`. Currently, this specification does not support transaction data
for AnonCreds and only supports the `nonce` from the Authorization Request and an audience binding as inputs for
the proof generation. The audience binding MUST be the value of the Client Identifier, except for requests over the DC API where it MUST be
the Origin prefixed with `origin:`, as described in (#dc_api_response).

To compute to the `nonce` parameter that is used as an input for the Prove and Verify operations of AnonCreds, the `nonce` from the
Authorization Request must be concatenated with the audience binding (as defined above) and hashed using sha-256. The first
80 bits of that digest are then used as the `nonce` paramter for the AnonCreds proof. This computed nonce MUST then be used as nonce (also called n_1) for
the Presentation generation and verification as defined in sections 9.6 and 9.7 of [@Hyperledger.AnonCreds].
```

Sakurann (approved):
looks good - thank you!

danielfett (approved):

paulbastian (approved):
I don't think that anyway is currently using Anoncreds over OpenID, but the proposal seems legit to me.

PR #500, opened 2025-04-04
revamp vp formats

TimoGlastra:
Supersedes #488

Fixes #495 
Fixes #487 

~~Dependant on #479 (it started from that branch due to the conflicts, once that PR is merged i will rebase). You can see the changes from this PR here: https://github.com/openid/OpenID4VP/commit/c5818cdc480634a058378c6e015052cda9324899~~

- Unified definition of `vp_formats` and `vp_formats_supported` for each format
- Updated inconsistent formats usage between `vp_formats` and `vp_formats_supported`. 
- Extended LDP proofs with `cryptosuite` to make it easier to query/define supported schemes (since most proofs will use `DataIntegrityProof`
- Added a definition for mso_mdoc. I based it on @awoie's PR in OID4VCI, and tried to be consistent and defined both IssuerSigned and DeviceSigned supported algorithms. https://github.com/openid/OpenID4VCI/pull/460. I was doubting if we need it since I think 18013-5 already has requirements on the supported algorithms, but in practice a lot of implementations do not support all algorithms. I'm open to suggestions to make this simpler, but just wanted to show what it would look like for mdoc.

NOTE: this has breaking changes, but I think the previous definitions were really underspecified (and leaning on PEX), and not consistent. So I think it's worth it to make these changes before 1.0

PR #513, opened 2025-04-08
Allow presentation without holder binding

### LINKED ISSUE #6: Presenting VC without a VP (non key-bound credentials)

A use-case being… A doctor is employed by an agency B, who has a contract with hospital A. So to prove that he works for hospital A, the doctor needs to send a VC1 that he works for agency B hospital A and VC2 that the agency B is contracted by hospital A. 

Doctor can create a VP1 of VC1, because it is issued to him, but he cannot create a VP of VC2 because VC2 is issued to an agency. but it should be ok as long as verifier trusts hospital’s signature. 

I think theoretically, it could be expressed in a descriptor\_map in a presentation\_submission that one VC can be found inside a VP and another VC is not inside a VP but as-is?

danielfett:
Fixes #6 

Executive summary:

- in the text, use "Verifiable Presentation" to refer to Presentations with Holder Binding Proofs only; use "Presentation" as the generic term for all kinds of credential presentations
- adapt the terminology section accordingly
- introduce considerations for using `state`: now REQUIRED if **at least one Presentation without Holder Binding is requested**.
   _Alternative: "if **only pres. without holder binding** are requested", but the current rule seems easier to implement considering that credentials can be returned conditionally etc._
- introduce the new DCQL credential query parameter `allow_replay` that requests a Presentation without Holder Binding
- introduce some format-specific considerations:
   - SD-JWT VC needs just allow_replay: true
   - for W3C VCs, we already had four format identifiers distinguishing between presentations and credentials: `jwt_vp_json`, `jwt_vc_json`, `ldp_vp`, `ldp_vc` — I therefore introduced the rule that *vc must be used if and only if allow_replay is true. 
      _Alternative 1: Remove the *vp formats, but that would be another breaking change, I suppose? Alternative 2: Remove the *vc formats, but that would mean requesting a "*vp" when in reality, only a credential is returned._
- adapt the security considerations to make clear that without key binding, replay of credentials is possible

Let the bikeshedding on the parameter name begin, other ideas I had were `insecure_no_binding` and simply `key_binding: false`. I think `allow_replay` captures the risk quite well and to the point, without sounding too scare, and doesn't sound too nice for developers to "just give it a try".


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2257,7 +2312,9 @@ The `vp_formats` parameter of the Verifier metadata MUST have the key `jwt_vp_js
 
 This section illustrates the presentation of a Credential conformant to [@VC_DATA] that is signed using JWS, and does not use JSON-LD.
 
-The Credential format identifiers are `jwt_vc_json` to request a W3C Verifiable Credential and `jwt_vp_json` to request a W3C Verifiable Presentation.
+The Credential Format Identifiers are `jwt_vc_json` to request a W3C Verifiable Credential and `jwt_vp_json` to request a Verifiable Presentation compliant to the [@VC_DATA] specification.
+
+If `jwt_vc_json` is used, the Credential is returned without Holder Binding. Therefore, the Verifier MUST set the `allow_replay` parameter in the Credential Query to `true` if it is using `jwt_vc_json`. Otherwise, `allow_replay` MUST be `false`. Wallets MUST reject requests that do not follow these rules.
```
TimoGlastra, line 2317: I feel like `allow_replay` is more of a result of not requiring key binding. But I don't think it's the best name for a query parameter.

I'm not sure if there's an overarching term. I think in OID4VCI it's called key binding (so something like 'require_key_binding`) or more generic that a presentation is required (so `require_verifiable_presentation`).

danielfett: I don't think `require_key_binding` sounds serious enough :-) I'd prefer a wording that conveys that this is a critial parameter.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,22 +314,65 @@ The following is a non-normative example of a transaction data content, after ba
...
+`state`:
+: REQUIRED under the conditions defined in (#nkb-credentials). Otherwise, `state` is OPTIONAL. `state` values MUST only contain ASCII URL safe characters (uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde).
+
+## Requesting Presentations without Holder Binding Proofs {#nkb-credentials}
+
+The primary use case of this specification is to request and present Verifiable
+Presentations, i.e., Presentations that contain a Holder Binding proof.
+
+However, there are use cases where the Verifier wants to request presentation of
+Credentials without a proof of Holder Binding. Examples for such use cases include
```
tlodderstedt, line 342:
```suggestion
Credentials without a proof of cryptographic Holder Binding. Examples for such use cases include
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,22 +314,65 @@ The following is a non-normative example of a transaction data content, after ba
...
+ticket), Credentials that are bound to a biometric trait, or Credentials that
+are bound to claims (e.g., a diploma). In some cases, Credentials may support
+Holder Binding, but the Verifier may not require it for the Presentation.
+
+A Verifier that requests and accepts a Presentation of a Credential without a
+proof of Holder Binding accepts that the presented Credential may have been
+replayed. (#preventing-replay) contains additional considerations for this case.
+
+To request a Credential without proof of Holder Binding, the Verifier specifies a
+different format in the DCQL request as defined in (#dcql_query) and
```
tlodderstedt, line 353: would it make sense to add an example?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -674,6 +710,14 @@ Note that Relying Parties must verify that the issuer of a received presentation
 trusted on their own and this feature mainly aims to help data minimization by not
 revealing information that would likely be rejected.
 
+`allow_replay`:
```
tlodderstedt, line 713: sounds weird, no one is interested in allowing replay - I would rather call this element "key_binding_required"

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1336,11 +1380,15 @@ This document also defines the following additional error codes and error descri
...
 
 1. Validate the format of the VP Token as defined in (#response-parameters).
-1. Check that the set of VPs returned satisfies all requirements defined in the Verifier's request as described in (#dcql_query_lang_processing_rules).
-1. Validate the integrity, authenticity, and Holder Binding of any Verifiable Presentation provided in the VP Token according to the rules of the respective Presentation format. See (#preventing-replay) for the checks required to prevent replay of a VP.
-1. Perform the checks on the Credential(s) specific to the Credential Format (i.e., validation of the signature(s) on each VC).
-1. Confirm that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
-1. Perform the checks required by the Verifier's policy based on the set of trust requirements such as trust frameworks it belongs to (i.e., revocation checks), if applicable.
+1. Check the individual Presentations according to the specific Credential or Presentation format requested:
+   1. Validate the integrity and authenticity of the Presentation and Credential.
```
tlodderstedt, line 1384: what about the state?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1583,21 +1625,33 @@ Note: If the Verifier's Response URI did not return a `redirect_uri` in step (6)
...
+An attacker could try to inject Presentations obtained from (for example) a previous Authorization Response into another Authorization Response, thus impersonating the End-User that originally presented the respective Verifiable Presentation. Holder Binding aims to prevent such attacks.
+
+### Presentations without Holder Binding Proofs
 
-An attacker could try to inject a VP Token (or an individual Verifiable Presentation), that was obtained from a previous Authorization Response, into another Authorization Response thus impersonating the End-User that originally presented that VP Token or the respective Verifiable Presentation.
+By definition, Presentations without Holder Binding (see (#nkb-credentials)) do
+not prove that the Holder possesses the private Holder Binding key belonging to
+the Credential. A Verifier that consumes Presentations without Holder Binding
+accepts the risk that the Holder may have obtained the Credential from a third
+party (e.g., by playing the role of a Verifier) and that the Holder may not be
```
tlodderstedt, line 1638: I suggest to add considerations how the binding could be achieved, like claims binding to other credentials.

openid-4-verifiable-presentations-1_0.md
tlodderstedt: I agree with @TimoGlastra .

tlodderstedt (changes requested):

paulbastian (changes requested):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,22 +314,65 @@ The following is a non-normative example of a transaction data content, after ba
...
 ```
 
-## Existing Parameters
+## Existing Parameters {#existing_parameters}
 
 The following additional considerations are given for pre-existing Authorization Request parameters:
 
 `nonce`:
-: REQUIRED. Defined in [@!OpenID.Core]. It is used to securely bind the Verifiable Presentation(s) provided by the Wallet to the particular transaction. See (#preventing-replay) for details. Values MUST only contain ASCII URL safe characters (uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde).
+: REQUIRED. Defined in [@!OpenID.Core]. It is used to securely bind Verifiable Presentation(s) provided by the Wallet to the particular transaction. The Verifier MUST create a fresh, cryptographically random number with sufficient entropy for every Authorization Request, store it with its current session, and pass it in the `nonce` Authorization Request Parameter to the Wallet. See (#preventing-replay) for details. Values MUST only contain ASCII URL safe characters (uppercase and lowercase letters, decimal digits, hyphen, period, underscore, and tilde).
```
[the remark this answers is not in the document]
danielfett: To be discussed - I left it REQUIRED for now. Technically not required if only asking for presentations, but then again, it is not hard to always add a nonce.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -674,6 +710,14 @@ Note that Relying Parties must verify that the issuer of a received presentation
 trusted on their own and this feature mainly aims to help data minimization by not
 revealing information that would likely be rejected.
 
+`allow_replay`:
+: OPTIONAL. A boolean which indicates whether the Verifier requires a Holder Binding
+proof. The default value is `false`, i.e., a Verifiable Presentation with Holder Binding
+is requested. If set to `true`, the Verifier accepts a Credential without Holder Binding
+proof. The Wallet MUST NOT include a Holder Binding proof in this case.
+
+Depending on the Credential format, a different Credential Format Identifier may be required to request Presentations without Holder Binding. Details are defined in (#format_specific_parameters).
```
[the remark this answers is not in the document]
danielfett: That would work, but if we only use `ldp_vc`, for example, it means another breaking change (the format parameter is not only used in the DCQL query). If we only use `ldp_vp`, the identifier would not be correct in case no holder binding is requested.

openid-4-verifiable-presentations-1_0.md
danielfett: `state` is not in the VP Token that is validated here. We don't have a section on general response validation yet, I think.

There is a clause in the text I added further above that `state` must be validated, if used.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1336,11 +1380,15 @@ This document also defines the following additional error codes and error descri
...
-1. Perform the checks on the Credential(s) specific to the Credential Format (i.e., validation of the signature(s) on each VC).
-1. Confirm that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
-1. Perform the checks required by the Verifier's policy based on the set of trust requirements such as trust frameworks it belongs to (i.e., revocation checks), if applicable.
+1. Check the individual Presentations according to the specific Credential or Presentation format requested:
+   1. Validate the integrity and authenticity of the Presentation and Credential.
+   1. Validate that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
+   1. Validate that all Presentations contain a cryptographic proof of Holder Binding (i.e., that they are Verifiable Presentations), unless specifically requested otherwise.
+   1. For Verifiable Presentations, validate the Holder Binding. See (#preventing-replay) for the checks required to prevent replay of a Verifiable Presentation.
```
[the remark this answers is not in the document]
danielfett: This is in practice the same step. I'll change the wording a bit to clarify that.
```suggestion
   1. Validate that all Presentations contain a cryptographic proof of Holder Binding (i.e., that they are Verifiable Presentations), unless specifically requested otherwise.
   1. For Verifiable Presentations, validate the Holder Binding, including the checks required to present replay described in (#preventing-replay).
```

openid-4-verifiable-presentations-1_0.md
danielfett: From my experience with similar discussions in SD-JWT, this is a bit of a rathole and there are a lot of opinions on this topic :-) I suggest to file an issue and add text later.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2565,7 +2624,7 @@ This section defines how Credentials complying with [@!I-D.ietf-oauth-sd-jwt-vc]
 
 ### Format Identifier
 
-The Credential format identifier is `dc+sd-jwt`.
+The Credential Format Identifier is `dc+sd-jwt`. By default, an SD-JWT [@!I-D.ietf-oauth-selective-disclosure-jwt] with a Key Binding JWT (SD-JWT+KB) is returned as the Verifiable Presentation. If `allow_replay` is set to `true` in the Credential Query, an SD-JWT without the Key Binding JWT is returned.
```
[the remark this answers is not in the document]
danielfett: Agree with moving the text further above.

I think `allow_replay` must be set to `true` any time an SD-JWT is presented without key binding. No matter if it supports key binding or not.

openid-4-verifiable-presentations-1_0.md
danielfett: Making this sound slightly weird and not like something you want to use was exactly the intention. I fear that Verifiers may just use it because "I don't need this weird proof thing anyway".

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: The identifiers are both used in `vp_formats_supported` to distinguish the algorithm for the VC and the one for the VP (see [here](https://github.com/openid/OpenID4VP/pull/500/files#diff-63c80bfdcc58989ea965cff9db251b78ceb99b68a7afeb03066f440535863bf1)). @TimoGlastra As you recently looked into this - do you have an idea how to solve this?

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
danielfett: Text was changed to reflect that.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
TimoGlastra: We're already make breaking changes to `vp_formats_supported`, so I don't see much problem in making more breaking changes.

I do think it's important to stay consistent with OpenID4VCI. The same format identifiers are used there (and making the same changes to OpenID4VCI might be a bigger concern now)

Some options I see:
1. Update to a new combined format identifier, e.g. `w3c_di`
  - this is better aligned with how W3C (JSON-LD) credentials are secured today (it's now called data integrity proofs rather than linked data proofs).
  - we merge the `vp_formats_supported` into one object under `w3c_di`, similar to e.g. `mso_mdoc` which has `issuer_signed_alg_values` and `device_signed_alg_values`. I didn't want to introduce this change as part of #500 as I didn't think it was worth the breaking change, but if we're already merging the format identifiers I think using a different identifier does not really anymore anymore.
  - the format identifier for dcql always becomes `w3c_di`.
2.  We keep using `ldp_vc`, since that is already used in the DCQL query
  - We still remove the `ldp_vp` format from the `vp_formats_supported` and merge it under one object.
  - SD-JWT also uses `dc+sd-jwt` (previously `vc+sd-jwt`) to refer to both the VC and VP, so one could argue that using `vc` in the format identifier is okay for both VC and VP. Although I think the concern here would be that `ldp_vc` used to have a different meaning.

One thing with merging the format identifiers is that with W3C there's a completely different structure returned whether a VC or VP is returned, so you'd have to look into the `type` value to see if `VerifiableCredential` or `VerifiablePresentation` is used (and check based on the query `allow_replay` value what it should be). This seems reasonable and simple to me.

DCQL with W3C support is new and wasn't really defined up until like a week ago, so there's not any adoption of DCQL + W3C yet. So the main thing we should be concerned about is the `vp_formats_supported` structure breaking, but that is already breaking in #500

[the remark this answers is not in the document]
danielfett: The second solution is what I had in mind - it seems to be the smaller change from what we have right now.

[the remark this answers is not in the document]
danielfett: I went with the second solution, as it seems to be @paulbastian's preferred solution as well, i.e., there are only `jwt_vc_json` and `ldp_vc` now. The text is changed accordingly, please adapt #500, thanks!

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2236,17 +2291,17 @@ OpenID for Verifiable Presentations is Credential Format agnostic, i.e., it is d
...
+The `claims_path` parameter in the Credential Query as defined in (#credential_query) is used to specify the claims that the Verifier wants to receive in the Presentation. When used in the context of W3C Verifiable Credentials, the `claims_path` parameter always matches on the root of Verifiable Credential (not the Verifiable Presentation). Examples are shown in the following subsections.
 
 ### Verifier Metadata
 
-The `vp_formats` parameter of the Verifier metadata MUST have the key `jwt_vp_json` or `ldp_vp` (according to the format, as defined in (#jwt_vc) and (#ldp_vc)), and the value MUST be an object consisting of the following name/value pair:
+The `vp_formats` parameter of the Verifier metadata MUST have the keys `jwt_vp_json`, `jwt_vc_json`, `ldp_vp` or `ldp_vc` (according to the format, as defined in (#jwt_vc) and (#ldp_vc)), and for `jwt_vp_json` and `ldp_vp`, the value MUST be an object consisting of the following name/value pair:
```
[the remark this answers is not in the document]
danielfett: Done, see discussion above.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
paulbastian: That's my preferred solution as well, as it makes things more consistent with OpenID4VCI.

paulbastian (commented):
From my view, two open questions to the WG:

- [x] nonce REQUIRED for requests without key binding? 
- [x] rename `allow_replay` to `key_binding_required` ?

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
c2bo: We could add similar text to the one in state (required if you are requesting anything that has a key binding)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -674,6 +709,12 @@ Note that Relying Parties must verify that the issuer of a received presentation
 trusted on their own and this feature mainly aims to help data minimization by not
 revealing information that would likely be rejected.
 
+`allow_replay`:
+: OPTIONAL. A boolean which indicates whether the Verifier requires a Holder Binding
+proof. The default value is `false`, i.e., a Verifiable Presentation with Holder Binding
+is requested. If set to `true`, the Verifier accepts a Credential without Holder Binding
+proof. The Wallet MUST NOT include a Holder Binding proof in this case.
```
Sakurann, line 716:
```suggestion
`require_cryptographic_holder_binding`:
: OPTIONAL. A boolean which indicates whether the Verifier requires a Cryptographic Holder Binding
proof. The default value is `true`, i.e., a Verifiable Presentation with Cryptographic Holder Binding
is required. If set to `false`, the Verifier accepts a Credential without Cryptographic Holder Binding
proof.
```

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
Sakurann: WG discussion, agreed to keep it required

openid-4-verifiable-presentations-1_0.md
Sakurann: renamed to `require_cryptographic_holder_binding`

paulbastian (approved):

openid-4-verifiable-presentations-1_0.md
tlodderstedt: https://github.com/openid/OpenID4VP/issues/533

tlodderstedt (approved):

Sakurann (approved):

c2bo (approved):

jogu (approved):

PR #524, opened 2025-04-11
fix credential format specific metadata

Sakurann:
for AnonCreds, one of the schema_id_values or cred_def_id_values must be present.
for W3C, type_values must be present


Sakurann:
closing in favor or #526

Issue #520, opened 2025-04-09
Example that requests encrypted response should be added


jogu:
We don't appear to have an example that sets authorization_encrypted_response_alg and authorization_encrypted_response_enc, which I think means we don't have an example that actually requests an encrypted response.

bc-pi:
authorization_encrypted_response_alg was removed in #523 and a few hours later I did #536 to add an example for this

PR #523, opened 2025-04-11
Changing JARM references to JWT/JWE directly

bc-pi:
Inspired by the work of @GarethCOliver in #477 and borrowing heavily from it (thanks @GarethCOliver!), this Removes the JARM option for response signing, referencing JWT/JWE encryption directly.

This PR endeavors to accomplish the same general things as #477 and incorporate additional items agreed to https://github.com/openid/OpenID4VP/pull/477#discussion_r2031613035 during the pre-IIW XL hybrid DCP WG Meeting a few days ago. I got kinda stuck trying to overlay those updates onto #477 and eventually gave up and started fresh with this. 

The JARM client_metadata parameter authorization_encrypted_response_enc is still used, for compatibility etc. as is the jwks parameter. How they are used for response encryption alson is updated/clarified and the option for signing the authorization response has been removed.

closes https://github.com/openid/OpenID4VP/issues/463, closes https://github.com/openid/OpenID4VP/issues/438

also closes #477 (can it do that?)

PR #536, opened 2025-04-11
add example of response encryption

bc-pi:
Add some examples showing an encrypted response and a corresponding request with jwks. For #520


charsleysa (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1232,11 +1232,77 @@ To obtain the Verifier's public key to which to encrypt the Authorization Respon
...
 The JWE `alg` algorithm used MUST the `alg` value of the chosen `jwk`, if present, or otherwise make sense to use with the selected key.
 If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that used to encrypt the response.
-The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128CBC-HS256` when not explictiy set.
+The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the Wallet to use a content encryption algorithm of its choosing when not explicitly set.
```
jogu, line 1235: I'd say the practical effect of this is probably that verifiers have to set this value as they can't risk the wallet returning one they don't support?

bc-pi: Practically there are only 6 defined content encryption algorithms defined https://www.iana.org/assignments/jose/jose.xhtml and I am not aware of libraries/platforms that don't support most or all of them (thought of course my rough observation is not an objective measure). 

The current JOSE HPKE work is making some IMO very questionable choices with treatment of algorithms. The current draft has what amounts to a null content encryption algorithm that only can/should be used in conjunction with a particular HPKE based key management alg. 

With this wording, inspired by on informal questions you'd asked about the PR that proceeded this, I was trying to make room to better allow agility with respect to wherever the shenanigans with JOSE HPKE might land. That was probably a mistake. And definitely shouldn't have tied to adding a (hopefully) helpful example. 

This would undo that but fix a spelling error:

```suggestion
The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the for the default value of `A128CBC-HS256` when not explicitly set.
```

Sakurann: we agreed on A128GCM during post-IIW mtg

Sakurann (approved):
CBC -> GCM

Sakurann (changes requested):
CBC -> GCM

openid-4-verifiable-presentations-1_0.md
bc-pi: > we agreed on A128GCM during post-IIW mtg

That's what I thought but then the merged #523 had `A128CBC-HS256` in this part of the text. I *think* maybe the #523 had it in two places but then only one was changed during the post-IIW mtg. Anyway, how's about:

```suggestion
The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the for the default value of `A128GCM` when not explicitly set.
```

Sakurann: oh, then we probably missed it. thanks for catching! lgtm

selfissued: Yes, this was discussed on Friday, but to the extent that we're using the same parameter name as JARM, it should have the same meeting.  [JARM says](https://openid.net/specs/oauth-v2-jarm-final.html#name-client-metadata) "the default for this value is A128CBC-HS256."  We should do the same.

Related to this, a recorded action for me in the minutes from Friday was:
> Mike will share a link to text that says gcm default support is required for JARM.

In fact, A128CBC-HS256 is what's required, both for JARM and in the [IANA JSON Web Signature and Encryption Algorithms registry](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms).  I'll correct that on-list.  Whereas support for A128GCM is recommended in the registry, not required.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1232,11 +1232,92 @@ To obtain the Verifier's public key to which to encrypt the Authorization Respon
...
+(with added line breaks for display purposes only):
+
+```
+{
+ "response" : "eyJhbGciOiJFQ0RILUVTIiwiZW5jIjoiQTEyOEdDTSIsImtpZCI6ImFjIiwiZXBrIjp7Imt
+    0eSI6IkVDIiwieCI6Im5ubVZwbTNWM2piaGNhZlFhUkJrU1ZOSGx3Wkh3dC05ck9wSnVmeVlJdWsiLCJ5I
+    joicjRmakRxd0p5czlxVU9QLV9iM21SNVNaRy0tQ3dPMm1pYzVWU05UWU45ZyIsImNydiI6IlAtMjU2In1
+    9..uAYcHRUSSn2X0WPX.yVzlGSYG4qbg0bq18JcUiDRw56yVnbKR8E7S7YlEtzT00RqE3Pw5oTpUG3hdLN
+    4taHZ9gC1kwak8JOnJgQ.1wR024_3-qtAlx1oFIUpQQ"
+}
```
bc-pi, line 1301: I meant to, but apparently forgot, to include the private key too so folks playing along at home could decrypt this.

bc-pi: now added

bc-pi:
This has been updated to reconcile with changes in #555

selfissued (approved):
Reviewing this from my phone on a plane, so I'm going to trust that the encrypted content is correct.

bc-pi:
> Reviewing this from my phone on a plane, so I'm going to trust that the encrypted content is correct.

I did create and check the encrypted content but certainly having an independent check would be good too.

c2bo:
> > Reviewing this from my phone on a plane, so I'm going to trust that the encrypted content is correct.
> 
> I did create and check the encrypted content but certainly having an independent check would be good too.

can you paste the full JWE here? 

> > Reviewing this from my phone on a plane, so I'm going to trust that the encrypted content is correct.
> 
> I did create and check the encrypted content but certainly having an independent check would be good too.

Looks good to me - this is what I get when decoding and decrypting:

Header:
```json
{
  "alg": "ECDH-ES",
  "enc": "A128GCM",
  "kid": "ac",
  "epk": {
    "kty": "EC",
    "x": "nnmVpm3V3jbhcafQaRBkSVNHlwZHwt-9rOpJufyYIuk",
    "y": "r4fjDqwJys9qUOP-_b3mR5SZG--CwO2mic5VSNTYN9g",
    "crv": "P-256"
  }
}
```
Payload:
```json
{
  "vp_token": {
    "example_credential_id": [
      "eyJhb...YMetA"
    ]
  }
}
```

c2bo (approved):

jogu (approved):

Issue #535, opened 2025-04-11
Use cbor-to-json conversion (6.1) to provide the value matching


GarethCOliver:
For mdocs, perform value matching by converting the underlying field to json, and then matching on that.

GarethCOliver:
validate vs existing usecases + bytestrings (probably will work for other but out of scope to validate(

PR #538, opened 2025-04-11
Specifies value matching for mdocs via a reference to cbor-to-json

GarethCOliver:
Replacement for https://github.com/openid/OpenID4VP/pull/481 as discussed in the working group call.

Resolves https://github.com/openid/OpenID4VP/issues/535


davidz25 (approved):
LGTM, much cleaner than my PR - thanks Gareth!

charsleysa (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -896,6 +896,10 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+If a Wallet implements value matching and the Credential to match against is
+an ISO mdoc-based credential, the the CBOR value to match against MUST first be converted to JSON, following the advice 
+given in Section 6.1 of [@RFC8949]. The resulting JSON value is then be used to match against the values property as specified above.
```
charsleysa, line 901:
```suggestion
If a Wallet implements value matching and the Credential to match against is
an ISO mdoc-based credential, the CBOR value to match against MUST first be converted to JSON, following the advice 
given in Section 6.1 of [@RFC8949]. The resulting JSON value is then to be used to match against the values property as specified above.
```

c2bo (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -896,6 +896,10 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+If a Wallet implements value matching and the Credential to match against is
+an ISO mdoc-based credential, the CBOR value to match against MUST first be converted to JSON, following the advice 
+given in Section 6.1 of [@RFC8949]. The resulting JSON value is then to be used to match against the values property as specified above.
```
Sakurann, line 901:
```suggestion
If a Wallet implements value matching and the credential being matched is
an ISO mdoc-based credential, the CBOR value used for matching MUST first be converted to JSON, following the advice 
given in Section 6.1 of [@RFC8949]. The resulting JSON value is then used to match against the `values` property as specified above.
```
small tweak

openid-4-verifiable-presentations-1_0.md
```diff
@@ -896,6 +896,10 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
 rules are defined in (#selecting_claims).
 
+If a Wallet implements value matching and the credential being matched is
+an ISO mdoc-based credential, the CBOR value used for matching MUST first be converted to JSON, following the advice 
+given in Section 6.1 of [@RFC8949]. The resulting JSON value is then used to match against the `values` property as specified above.
```
jogu, line 901: It's really weird having a 'MUST' say to follow the non-normative advice in https://datatracker.ietf.org/doc/html/rfc8949#section-6.1

The example of the 'substitute value' being JSON null is even less normative (if being less normative than non-normative is actually possible). I suspect that case should perhaps explicitly fail rather than doing a match against null? (I think you get the same outcome either way but I'm not 100% sure!)

martijnharing (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -896,6 +896,10 @@ If the `values` property is present, the Wallet SHOULD return the claim only if
 type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing
```
martijnharing, line 896: To check whether the conversion works, the allowed values in values are "strings, integers or boolean values".
Does the "integers" in this context refer to the integer definition in RFC 8259, which is
`int = zero / ( digit1-9 *DIGIT )` i.e. negative values are not allowed.

c2bo: That RFC also uses integers for the "normal" definition of integers. A few lines below your quote it says:
> Note that when such software is used, numbers that are integers and
   are in the range [-(2**53)+1, (2**53)-1] are interoperable in the
   sense that implementations will agree exactly on their numeric
   values.
   
 Only that specific ABNF seems to use that weird definition of int.

tplooker (approved):

openid-4-verifiable-presentations-1_0.md
GarethCOliver: Could make it less strict, something like: 

...'the CBOR value used for matching MUST first be converted to JSON. Advice on how to do so is given in Section 6.1 of [@RFC8949]. It is RECOMMENDED that Wallets this advice, to provide consistent matching.'

To me, consistency is the goal, as long as the Verifiers and Wallet expect the same conversion then the matching logic will work pretty well (even if the logic they follow is weird). Problems only arise when there are relevant differences. We currently only care about equality on simple types so seems fine.

Sakurann:
```suggestion
given in Section 6.1 of [@RFC8949]. The resulting JSON value is then used to match against the `values` property as specified above.
When conversion according to these rules is not clearly defined, behavior is out of scope of this specification.
```

Sakurann:
wg discussion: ok to merge once @jogu re-reviews

jogu (approved):

Issue #519, opened 2025-04-09
Referencing ISO 18013-7 in OpenID4VP is problematic


TimoGlastra:
OpenID4VP latest draft refers to ISO 18013-7, which in turn refers to OpenID4VP ID2

As an implementer I think it's really problematic that OpenID4VP 1.0 will point to a spec (ISO 18013-7), that requires usage of OpenID4VP second implementers draft, which also mandates PEX and a lot of features that have changed since that draft was published. 

It will make mDOC over OpenID4VP without DC-API basically impossible to support for implementers targeting OpenID4VP 1.0 (if you're not using the DC-API). 

The text currently reads that you need to use 18013-7 to support mDOC without DC-API, which requires supporting an older draft version of OpenID4VP (with PEX!). Changing this behaviour post 1.0 will mean it's a breaking change and thus not possible.

I propose any references to ISO 18013-7 are removed before 1.0 is published. Ideally we add the missing piece for doing mDOC without 18013-7 before 1.0 (i think fixing https://github.com/openid/OpenID4VP/issues/402 should be enough), but otherwise leaving it unspecified is better than referencing 18013-7.

Sakurann:
WG discussion 
- remove references to 18013-7 in Final 1.0
- in 1.1, define how to do mdocs over openid4vp

TimoGlastra:
Thank you for considering this issue for 1.0

Issue #402, opened 2025-01-30
Define session transcript for OpenID4VP (without DC API) in OpenID4VP


TimoGlastra:
There has been a lot of discussions about the session transcript for the DC API, but I couldn't find anything related to the non-dc api.

#374  introduced text that no-dc-api should follow ISO 18013-7. However with the addition of the Session transcript for DC API to OID4VP, and with the PR to define session encryption details (https://github.com/openid/OpenID4VP/pull/380) i think it makes sense to also define the session transcript options for OID4VP without dc-api in OID4VP. 

Then the specific requirements around encryption algs can be defined in HAIP, and there's no need anymore for ISO 18013-7 when doing OID4VP and mDOC, especially when using DCQL and the PD restrictions don't apply.

Sakurann:
session transcript for OpenID4VP (without DC API) is currently defined in ISO 18013-7 annex B, which is to be followed for now. 
Issue with that Annex is that it relies on Second implementers draft of openid4vp, while the third one has been recently published. there is an agreement between ISO SC17 WG10 and OIDF that OIDF can define mdoc profile for openid4vp without browser api when it decides the time is right. the issue to track that in HAIP is here (https://github.com/openid/oid4vc-haip/issues/139), but i am ok keeping this one as its counterpart in openid4vp repo.

c2bo:
Apparently github understood
> We could specify a session transcript in Invocation via other methods {#non-dc-api-invocation} to resolve https://github.com/openid/OpenID4VP/issues/402

as "resolves 402" 😅

andprian:
If I understand correctly, there is consensus that OID4VP 1.0 (standalone) will not work for mDOC in modes direct_post and direct_post.jwt, is that right?

Sakurann:
WG call: after an alignment call with ETSI and EC, agreed to proceed with a PR

jogu:
> If I understand correctly, there is consensus that OID4VP 1.0 (standalone) will not work for mDOC in modes direct_post and direct_post.jwt, is that right?

I don't believe so. If such a restriction was made I think it would be made in HAIP. (ISO 18013-7 Annex B does prohibit use of direct_post/direct_post.jwt for mDL but VP would not be the correct place to make that restriction, nor would it be right to apply that to all mdoc cases.).

andprian:
@jogu , I was not talking about making a restriction. This is about defining SessionTranscript, otherwise vanilla OID4VP would not work without some kind of profiling for mDOC

[...]

PR #546, opened 2025-04-14
Remove references to 18013-7 from OpenId4VP

GarethCOliver:
This removes the references to 18013-7, per WG discussion as it is not compatible with the current document. 

resolves https://github.com/openid/OpenID4VP/issues/519

This approach to be brought up with WG10 this week.

We could specify a session transcript in  Invocation via other methods {#non-dc-api-invocation} to resolve https://github.com/openid/OpenID4VP/issues/402

as follows: 

OpenID4VPVanillaHandover = [
  "OpenID4VPVanillaHandover", ; A fixed identifier for this handover type
  OpenID4VPVanillaHandoverInfoHash ; A cryptographic hash of OpenID4VPVanillaHandoverInfo
]

; Contains the sha-256 hash of OpenID4VPVanillaHandoverInfoBytes
OpenID4VPVanillaHandoverInfoHash = bstr

; Contains the bytes of OpenID4VPVanillaHandoverInfo encoded as CBOR
OpenID4VPVanillaHandoverInfoBytes = bstr .cbor OpenID4VPVanillaHandoverInfo

OpenID4VPDCAPIHandoverInfo = [
  clientId,
  responseUri,
  nonce,
  jwk_thumbprint
] ; Array containing handover parameters

(nonce/jwk_thumbprint as in DC API, clientId/responseUri from request). 

This would ~match what is done today for 18013-7, while being inline with what we do in the DC API. However I think it leaves some questions (how do we handle multi-rp auth, do we really need clientId and responseUri) so would rather re-introduce in a follow-up PR, or in 1.1.


Sakurann:
makes me think if we should also remove references to 23220-4..?

GarethCOliver:
> makes me think if we should also remove references to 23220-4..?

The remaining 23220-4 are just to it as the 'building blocks' version of 18013-5. I think this still makes sense (it's only the remote presentation part that is in conflict).

Sakurann (approved):

jogu (approved):

awoie (approved):
LGTM, `SessionTranscript` for vanilla OID4VP could be done in 1.1.

tplooker (approved):
LGTM

Sakurann:
because some people asked.. i think what this PR means is.. 
- 18013-7 annex B continues to exist until ISO decides otherwise
- openid4vp only defines how to do mdocs over DC API
- opendi4vp is silent on how to do mdocs over vanilla openid4vp until dcp wg decides otherwise

Sakurann:
wg discussion: ok to merge once @martijnharing forms an opinion

PR #553, opened 2025-04-16
clarify issuere_signed_alg_values and device_signed_alg_values

### LINKED ISSUE #548: mdoc device signed can be digitally signed or HMAC-ed. for HMAC-ed device signed, agreed to define identifiers in openid4vp specification
mdoc device signed can be digitally signed or HMAC-ed. for HMAC-ed device signed, agreed to define identifiers in openid4vp specification

_Originally posted by @Sakurann in https://github.com/openid/OpenID4VP/pull/500#discussion_r2040063305_

Sakurann:
resolves #548 which was remaining after we merged #500

accommodates identifiers defined in 
- 18013-5 (including device mac auth)
- a fully-specified cryptographic algorithm spec
- else


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `issuer_signed_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithms identifiers. The `alg` COSE header (as defined in [@!RFC8152]) of the `IssuerSigned` structure of the presented mdoc MUST match one of the array values, when the `IssuerSigned` structure is signed with one of a fully-specified cryptographic algorithm defined in [@!I-D.ietf-jose-fully-specified-algorithms] or with  signature algorithms defined in section 9.1.2.4 of [@ISO.18013-5]. In all other cases (commonly known as a polymoprhic cryptographic algorithm identifier), the fully-specified algorithm derived from the combination of the `alg` value from the COSE header and the `crv` parameter from the signing key MUST match one of the array values.
```
martijnharing, line 2541: This definition says that one of the following two items is true for the "alg" header in the MSO:
a. IssuerSigned contains a fully specified cryptographic algorithm
b. A signature algorithm defined in section 9.1.2.4 of 18013-5, which are (ES256, ES384, ES512, EdDSA)
The "alg" header in the MSO MUST match one of the array values.

Since 18013-5 only specifies ES256, ES384, ES512 and EdDSA, (b) is almost always true. This means that if someone uses the fully specified cryptographic algorithms, the transaction will almost always fail, since they won't match.

If we want to support fully-specified algorithms in the issuer_signed_alg_values values, we should say that the equivalent combinations of 'alg' and the curve used are also considered to be a match.
We can then add an example saying that "ES256" when used with "P-256" matched the equivalent fully specified algorithm alg identifier.

Also note that there is no 'crv' parameter of the signing key, since the signing key is (typically) contained in an X.509 certificate, which does not have a 'crv' parameter.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `device_signed_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithms identifiers. The `alg` COSE header (as defined in [@!RFC8152]) of the `DeviceSigned` structure of the presented mdoc MUST match one of the array values, when the `DeviceSigned` structure is signed with one of a fully-specified cryptographic algorithm defined in [@!I-D.ietf-jose-fully-specified-algorithms] or with  signature algorithms defined in Section 9.1.3.6 of [@ISO.18013-5]. The `alg` COSE header MUST be `HMAC 256/256` when mdoc MAC Authentication defined in Section 9.1.3.5 of [@ISO.18013-5] is used. In all other cases (commonly known as a polymoprhic cryptographic algorithm identifier), the fully-specified algorithm derived from the combination of the `alg` value from the COSE header and the `crv` parameter from the signing key MUST match one of the array values.
```
martijnharing, line 2542: Same comment as for issuer_signed_alg_values applies here as well.

"The `alg` COSE header MUST be `HMAC 256/256` when mdoc MAC Authentication defined in Section 9.1.3.5 of [@ISO.18013-5] is used." What is the purpose of this sentence? This seems to just limit the maccing algorithm that can be used by an mdoc. While it aligns with the requirement in 18013-5, I don't see a reason why we need this requirement here.

If the goal of the device_signed_alg_values is to indicate which algorithms and curves the Verifier can support, shouldn't there also be a way for the Verifier to say which curves it supports for Maccing verification? If we do, I'm not sure if there is a clean way to use the device_signed_alg_values for that, since there currently is no identifier that combines HMAC 256/256 with a particular key.

martijnharing (changes requested):

openid-4-verifiable-presentations-1_0.md
Sakurann: yeah, i think i could have been clear enough. i think what i meant to say that when 18013-5 is used, use -5 identifiers. otherwise, use fully-specified, if neither is true, use polymorphic. 

> The issuing authority infrastructure shall use one of the signature algorithms for calculating the signature
over the MSO: “ES256” (ECDSA with SHA-256), “ES384” (ECDSA with SHA-384), “ES512” (ECDSA with
SHA-512) or “EdDSA” (EdDSA) shall be used. ”ES256” shall be used with curves P-256 and
brainpoolP256r1. “ES384” shall be used with curves P-384, brainpoolP320r1 and brainpoolP384r1.
“ES512” shall be used with curves P-521 and brainpoolP512r1. “EdDSA” shall be used with curves
Ed25519 and Ed448. For verifying the signature, the mdoc reader shall support all of these signature
algorithms and curves.

openid-4-verifiable-presentations-1_0.md
Sakurann:
```suggestion
* `device_signed_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If the DeviceSigned structure of the presented mdoc is signed using an algorithm defined in Section 9.1.3.6 of [@ISO.18013-5], the alg COSE header (as defined in [@!RFC8152]) MUST match one of the -5-style identifiers in the array. If mdoc MAC Authentication (as defined in Section 9.1.3.5 of [@ISO.18013-5]) is used, the alg COSE header MUST be HMAC 256/256. If 18013-5 is not used, but the signature uses a fully-specified algorithm defined in [@!I-D.ietf-jose-fully-specified-algorithms], the alg value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the alg value from the COSE header and the crv parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `issuer_signed_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If the `IssuerSigned` structure of the presented mdoc is signed using an algorithm defined in section 9.1.2.4 of [@ISO.18013-5], the `alg` COSE header (as defined in [@!RFC8152]) MUST match one of the -5-style identifiers in the array. If 18013-5 is not used, but the signature uses a fully-specified algorithm as defined in [@!I-D.ietf-jose-fully-specified-algorithms], the `alg` value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the `alg` value from the COSE header and the `crv `parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
```
martijnharing, line 2541: I don't think this updated text addresses the issue:

My interpretation of the proposed text is that it says:
1. If the 'alg' COSE header is one of the algorithms defined in 9.1.2.4 (ES256, ES384, ES512 or EdDSA), the 'alg' COSE header must match one  of the elements in the array.
2. If the 'alg' COSE header is one of the algorithms defined in [Jose-fully-specified] (ESP256, ESP384, ESP512, ESB256, ESB320, ESB384 or ESB512), the 'alg' COSE header must match one of the elements in the array.
3. If the 'alg' COSE header is anything else, the combination of the 'alg' value of the COSE header and the 'crv' parameter from the signing key must resolve to a fully-specified algorithm which must match one of the array values.

Since an mdoc not using one of the alg values as specified in 18013-5 will be uncommon at best (since it's not compliant to 18013-5), effectively 1. is the most common scenario. This means there is no way for the verifier to say "I support P-256, but not Brainpool256", or the other way round. Reading #500 , it seems that a (the?) reason to add this request parameter was to have a way for the verifier to indicate which curves it supports by having a 'crv' parameter. The 'crv' parameter was then removed noting that there is the [Jose-fully-specified] spec, but the currently proposed text doesn't actually allow the [Jose-fully-specified] spec to be used with the existing 'alg' parameters.

To support crv selection, either we need a crv_values parameter, or we need to allow the use of the [Jose-fully-specified] spec to be used when the 'alg' parameter in the header does not use an 'alg' from the [Jose-fully-specified] spec.

I think the intention is to do something like:
`issuer_signed_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If present, the credential must match one of the elements in the array. The credential matches a value in the array if one of the following is true:
- The value in the array matches the  'alg' value in the `IssuerAuth` COSE header. 
- The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the 'alg' value in the 'IssuerAuth' COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm.

As an example, if the 'IssuerAuth' structure contains an `alg` header with value `-7` (which stands for ECDSA with SHA-256) and is signed by a P-256 key, then it matches an `issuer_signed_alg_values` element of `-7` and `-9`.

Noting that the value of `-9` is the requested value for ECDSA using P-256 curve and SHA-256.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `device_signed_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If the DeviceSigned structure of the presented mdoc is signed using an algorithm defined in Section 9.1.3.6 of [@ISO.18013-5], the alg COSE header (as defined in [@!RFC8152]) MUST match one of the -5-style identifiers in the array. If mdoc MAC Authentication (as defined in Section 9.1.3.5 of [@ISO.18013-5]) is used, the alg COSE header MUST be HMAC 256/256. If 18013-5 is not used, but the signature uses a fully-specified algorithm defined in [@!I-D.ietf-jose-fully-specified-algorithms], the alg value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the alg value from the COSE header and the crv parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
```
martijnharing, line 2542: I think this has the same issues as the issuer_signed_alg_values for the signing operation, see the other comment on how I think it could be addressed.
However this would not yet address the issue for maccing mentioned in the previous comment. Since 18013-5 always uses HMAC 256/256, specifying that algorithm just says that maccing is supported. It doesn't say which curves. So including `device_signed_alg_values` is not possible if you only support a subset of curves for maccing. 
A possible solution could be to similarly define codes for 'fully' specified maccing algorithms to be used with this specification.

martijnharing (changes requested):

openid-4-verifiable-presentations-1_0.md
Sakurann: Asia wg call discussion

> To support crv selection, either we need a crv_values parameter, or we need to allow the use of the [Jose-fully-specified] spec to be used when the 'alg' parameter in the header does not use an 'alg' from the [Jose-fully-specified] spec.

- these are the options we have at hand. defining issuersigned_crv_values from IANA is probably a simpler and more future proof (from the perspective of new PQC algs being registered, for example).

openid-4-verifiable-presentations-1_0.md
Sakurann: same as above, add `devicesigned_crv_values`

openid-4-verifiable-presentations-1_0.md
awoie, line 2541: ISO does not use `crv` values but the signing algorithm can be represented by a COSE curve identifier. The curve used for the key in the signing algorithm is derived from the `x5chain`, i.e., the X.509 doc signer certificate. I'd also rename issuer signed -> issuer auth since more precise. Perhaps the following suggestion works:

```suggestion
* `issuer_auth_alg_values`: OPTIONAL. A JSON array containing cryptographic COSE algorithm identifiers. If present, the credential MUST be signed with one of the elements in the array. If the COSE algorithm identifier is polymorphic, it is RECOMMENDED to include `issuer_auth_crv_values` to clarify the ambiguity.
* `issuer_auth_crv_values`: OPTIONAL. A JSON array containing cryptographic COSE curve identifiers. If present, the cryptographic curve of the key that signed the credential MUST be  one of the curve identifiers included in the array. 
```

openid-4-verifiable-presentations-1_0.md
awoie, line 2542: I would also rename device signed -> device auth (since MAC and ECDSA is supported) and it is also more precise. Same as above but it should be noted that HMAC with DH key exchange is supported:
```suggestion
* `device_auth_alg_values`: OPTIONAL. A JSON array containing cryptographic COSE algorithm identifiers. If present, the credential MUST be signed or MACed with one of the elements in the array. If the COSE algorithm identifier is polymorphic, it is RECOMMENDED to include `device_auth_crv_values` to clarify the ambiguity.
* `device_auth_crv_values`: OPTIONAL. A JSON array containing cryptographic COSE curve identifiers. If present, the cryptographic curve of the key that signed or was used to derive the MAC key to sign or compute the MAC of the credential MUST be one of the curve identifiers included in the array. 
```

awoie (changes requested):
requested some changes

openid-4-verifiable-presentations-1_0.md
Sakurann, line 2542:
```suggestion
* `issuersigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If the `IssuerSigned` structure of the presented mdoc is signed using an algorithm and a curve defined in section 9.1.2.4 of [@ISO.18013-5], the Verifier and the Wallet MUST map the combination of these values to a fully specified algorithm identifier, which MUST match one of the array values. If 18013-5 is not used, but the signature uses a fully-specified algorithm as defined in [@!I-D.ietf-jose-fully-specified-algorithms], the `alg` value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the `alg` value from the COSE header and the `crv `parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
* `devicesigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers.If the `DeviceSigned` structure of the presented mdoc is signed using an algorithm and a curve defined in section 9.1.3.6 of [@ISO.18013-5], the Verifier and the Wallet MUST map the combination of these values to a fully specified algorithm identifier, which MUST match one of the array values. If mdoc MAC Authentication (as defined in Section 9.1.3.5 of [@ISO.18013-5]) is used, the `alg` COSE header MUST be `HMAC 256/256`. If 18013-5 is not used, but the signature uses a fully-specified algorithm defined in [@!I-D.ietf-jose-fully-specified-algorithms], the `alg` value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the `alg` value from the COSE header and the `crv` parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `issuersigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If the `IssuerSigned` structure of the presented mdoc is signed using an algorithm and a curve defined in section 9.1.2.4 of [@ISO.18013-5], the Verifier and the Wallet MUST map the combination of these values to a fully specified algorithm identifier, which MUST match one of the array values. If 18013-5 is not used, but the signature uses a fully-specified algorithm as defined in [@!I-D.ietf-jose-fully-specified-algorithms], the `alg` value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the `alg` value from the COSE header and the `crv `parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
```
awoie, line 2541:
```suggestion
* `issuer_auth_alg_values`: OPTIONAL. A JSON array containing cryptographic COSE algorithm identifiers. If present, the `IssuerAuth` structure as defined in [@ISO.18012-5] MUST be signed with an algorithm that matches one of the algorithms in the array. If the COSE algorithm identifier is polymorphic, it is RECOMMENDED to include `issuer_auth_crv_values` to clarify the ambiguity.
* `issuer_auth_crv_values`: OPTIONAL. A JSON array containing cryptographic COSE curve identifiers. If present, the cryptographic curve of the key that signed the `IssuerAuth` structure as defined in [@ISO.18013-5] MUST match one of the curve identifiers included in the array. 
```

paulbastian (commented):
As much as I love fully-specified algorithms, it may be easier to implement to use the versions using `crv` as defined in -5

openid-4-verifiable-presentations-1_0.md
Sakurann: WG discussion. going fully specified way

```suggestion
* `issuersigned_alg_values` : OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If present, the credential must match one of the elements in the array. The credential matches a value in the array if one of the following is true: 1) The value in the array matches the 'alg' value in the IssuerAuth COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the 'alg' value in the 'IssuerAuth' COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. As an example, if the 'IssuerAuth' structure contains an alg header with value -7 (which stands for ECDSA with SHA-256) and is signed by a P-256 key, then it matches an issuer_signed_alg_values element of -7 and -9.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `devicesigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers.If the `DeviceSigned` structure of the presented mdoc is signed using an algorithm and a curve defined in section 9.1.3.6 of [@ISO.18013-5], the Verifier and the Wallet MUST map the combination of these values to a fully specified algorithm identifier, which MUST match one of the array values. If mdoc MAC Authentication (as defined in Section 9.1.3.5 of [@ISO.18013-5]) is used, the `alg` COSE header MUST be `HMAC 256/256`. If 18013-5 is not used, but the signature uses a fully-specified algorithm defined in [@!I-D.ietf-jose-fully-specified-algorithms], the `alg` value MUST match one of the array values. In all other cases — typically involving polymorphic algorithm identifiers — the combination of the `alg` value from the COSE header and the `crv` parameter from the signing key MUST resolve to a fully-specified algorithm, which MUST match one of the array values.
```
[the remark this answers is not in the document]
Sakurann: @martijnharing to make a suggestion on a text that defines hmac identifier used with mdoc mac authn for -5 in openid4vp

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,8 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `issuersigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. If present, the Credential must match one of the elements in the array. The Credential matches a value in the array if one of the following is true: 1) The value in the array matches the 'alg' value in the IssuerAuth COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the `alg` value in the `IssuerAuth` COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. As an example, if the `IssuerAuth` structure contains an `alg` header with value `-7` (which stands for ECDSA with SHA-256) and is signed by a P-256 key, then it matches an `issuersigned_alg_values` element of `-7` and `-9`.
```
Sakurann, line 2541:
```suggestion
* `issuersigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. The Credential MUST be considered to fulfill requirement(s) expressed in this parameter if one of the following is true: 1) The value in the array matches the 'alg' value in the IssuerAuth COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the `alg` value in the `IssuerAuth` COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. As an example, if the `IssuerAuth` structure contains an `alg` header with value `-7` (which stands for ECDSA with SHA-256) and is signed by a P-256 key, then it matches an `issuersigned_alg_values` element of `-7` and `-9`.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2538,8 +2538,22 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `issuersigned_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. The Credential MUST be considered to fulfill requirement(s) expressed in this parameter if one of the following is true: 1) The value in the array matches the 'alg' value in the IssuerAuth COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the `alg` value in the `IssuerAuth` COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. As an example, if the `IssuerAuth` structure contains an `alg` header with value `-7` (which stands for ECDSA with SHA-256 in [@IANA.COSE]) and is signed by a P-256 key, then it matches an `issuersigned_alg_values` element of `-7` and `-9` (which stands for ECDSA using P-256 curve and SHA-256 in [@!I-D.ietf-jose-fully-specified-algorithms]).
```
Sakurann, line 2541:
```suggestion
* `issuerauth_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. The Credential MUST be considered to fulfill requirement(s) expressed in this parameter if one of the following is true: 1) The value in the array matches the 'alg' value in the IssuerAuth COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the `alg` value in the `IssuerAuth` COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. As an example, if the `IssuerAuth` structure contains an `alg` header with value `-7` (which stands for ECDSA with SHA-256 in [@IANA.COSE]) and is signed by a P-256 key, then it matches an `issuerauth_alg_values` element of `-7` and `-9` (which stands for ECDSA using P-256 curve and SHA-256 in [@!I-D.ietf-jose-fully-specified-algorithms]).
```

awoie (approved):
Doesn't seem to be simpler than defining an additional _crv param but still approving this.

davidz25 (approved):
Looked through, seemed to match what we talked about in the DCP call earlier, thanks!

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2579,8 +2588,23 @@ Some document types support some transaction data ((#transaction_data)) to be pr
...
+* `deviceauth_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. The Credential MUST be considered to fulfill requirement(s) expressed in this parameter if one of the following is true: 1) The value in the array matches the 'alg' value in the `DeviceAuth` COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the `alg` value in the `DeviceAuth` COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. 3) The value in the array is the `alg` of the COSE header is `HMAC 256/256` (as described in Section 9.1.3.5 of [@ISO.18013-5]) and the curve of the device key (from Table 22 of [@ISO.18013-5]) matches one of the values defined in the following table:
```
Sakurann, line 2592:
```suggestion
```suggestion
* `deviceauth_alg_values`: OPTIONAL. A JSON array containing cryptographic algorithm identifiers. The Credential MUST be considered to fulfill requirement(s) expressed in this parameter if one of the following is true: 1) The value in the array matches the 'alg' value in the `DeviceSignature` or `DeviceMac` COSE header. 2) The value in the array is a fully specified algorithm according to [@!I-D.ietf-jose-fully-specified-algorithms] and the combination of the `alg` value in the `DeviceSignature` COSE header and the curve used by the signing key of the COSE structure matches the combination of the algorithm and curve identified by the fully specified algorithm. 3) The value in the array is the `alg` of the `DeviceMac` COSE header is `HMAC 256/256` (as described in Section 9.1.3.5 of [@ISO.18013-5]) and the curve of the device key (from Table 22 of [@ISO.18013-5]) matches one of the values defined in the following table:
```
```

martijnharing (approved):

tplooker:
Very confused to have seen this PR merged in this fashion considering the conversation that was had on the Asia call earlier in the day, this is not what was discussed or agreed. What was discussed and I believe there was agreement on, including multiple people in this issue referencing the same solution was the preference to have a second metadata element which communicated the curve.

The approach decided here limits our ability to be forward compatible with new algorithms that might get registered in the COSE IANA registry as we are limiting ourselves to only a subset of algorithms supported in COSE by pointing at a specific IETF document rather than the underlying IANA registry. That means if we want to support PQ algs in future we have to revise this text. I also question whether it is even going to work in practise as the key the verifier uses could technically be a different curve to the device key that is encapsulated in the issuer auth structure.

Sakurann:
I think what was discussed was that since we don't think we expect fully specified algs to be used, let's optimize for the solution for 18013-5 implementations. ( I forgot to ask for the note taker so we unfortunately do not have detailed notes, sorry about that.. )

PR #554, opened 2025-04-16
clarify client id prefix processing

### LINKED ISSUE #484: Client_id requirements and clarifications
With the recent changes to the client_id and the multi-rp PR merged, there are some clarifications necessary for the client_id definition and processing.

Some of this could also be handled by https://github.com/openid/OpenID4VP/pull/465 but I'm not sure whether it's clearer to split these comments or not.

Some of the ambiguity comes from the definition. The client_id is currently defined through a reference to RFC 6749. However RFC 6749 defines the client_id as being issued by the authorization server as part of a registration process. client_ids aren't always part of a registration process (e.g. they can be derived from other information like a certificate) and they aren't always issued by the authorization server. 
Can we change the reference, or add language that these requirements from RFC 6749 do not apply?

For some of the requirements we have to state more explicitly who has to verify those:

- For x509_san_dns: "The Wallet MUST validate the signature and the trust chain of the X.509 certificate." Other schemes have similar requirements.
Can we add a general statement along the lines of "Parsing and verification requirements for client identifiers are only applicable to the Wallet when the Wallet uses the information from that client identifier."

- For x509_hash: "When the Client Identifier Scheme is x509_hash, the Client Identifier MUST be a hash and match the hash of the leaf certificate passed with the request." Other schemes have similar requirements.
Is it required, recommended or optional for the Wallet to verify this? For example when using the x509_hash and the DC api, the client_id is not included in the response, and all the information used to verify the signature is in the certificate itself. Therefore the hash is effectively duplicating information of the certificate. Since both are signed over, I don't think there is a security benefit from actually matching the x509_hash against the certificate itself. 
To make sure that there is agreement on verification requirements, we have to state explicitly whether and when these checks must be done or not by the Wallet.

Sakurann:
resolves #484


openid-4-verifiable-presentations-1_0.md
```diff
@@ -261,7 +261,7 @@ when it receives `request_uri_method` parameter with the value `post` but does n
...
+The Verifier communicates a Client Identifier Prefix that indicate how the Wallet is supposed to interpret the Client Identifier and associated data in the process of Client identification, authentication, and authorization as a prefix in the `client_id` parameter. This enables deployments of this specification to use different mechanisms to obtain and validate Client metadata beyond the scope of [@!RFC6749]. "A certain Client Identifier Prefix sets the requirements whether the Verifier needs to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
```
selfissued, line 264:
```suggestion
The Verifier communicates a Client Identifier Prefix that indicate how the Wallet is supposed to interpret the Client Identifier and associated data in the process of Client identification, authentication, and authorization as a prefix in the `client_id` parameter. This enables deployments of this specification to use different mechanisms to obtain and validate Client metadata beyond the scope of [@!RFC6749]. A certain Client Identifier Prefix sets the requirements whether the Verifier needs to sign the Authorization Request as means of authentication and/or pass additional parameters and require the Wallet to process them.
```

charsleysa (approved):

c2bo (approved):

bc-pi (approved):

PR #555, opened 2025-04-16
Rename authorization_encrypted_response_enc parameter

### LINKED ISSUE #552: Treatment of encrypted responses inconsistent with JARM while reusing JARM parameter
#523 removed the reference to [JARM](https://openid.net/specs/oauth-v2-jarm.html), kept the JARM parameter `authorization_encrypted_response_enc`, but deleted the `authorization_encrypted_response_alg` parameter that it is paired with.  Throughout OpenID and OAuth specifications, pairs of *_alg and *_enc parameters are used to specify what encryption algorithms are used.  The deletion of `authorization_encrypted_response_alg` breaks that pattern.

Also, our current use of `authorization_encrypted_response_enc` is inconsistent with its definition in JARM.  In JARM, when omitted, the default content encryption algorithm is `A128CBC-HS256`.  [In OpenID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0-26.html#name-new-parameters), it's currently `A128GCM`.  Since we're using the same parameter, we should be using it in the same way as already specified.

In my view, we shouldn't be creating bespoke means of communicating algorithms when there was already a standard way of doing so that we were using.  Continuing to use the `authorization_encrypted_response_alg` would better align us with what's come before.  Please restore it.

I'll also observe that pulling the algorithm out of the key goes in the opposite direction that OAuth and other standards are going.  [Fully-Specified Algorithms](https://datatracker.ietf.org/doc/draft-ietf-jose-fully-specified-algorithms/) is there exactly to enable signing and encryption parameters to be declared in metadata, without having to inspect keys to learn them.  (Also, if we're going to pull the algorithms from keys, we need to require the optional `alg` JWK parameter, which the spec doesn't currently do, but again, I think this has taken us in the wrong direction, so I'm not advocating that.)

I'll note that an alternative to using the two metadata parameters as specified in JARM that is acceptable to me if we truly need to be able to specify multiple algorithms is to use plural versions of them, as is done for other parameters in [RP Metadata Choices](https://openid.net/specs/openid-connect-rp-metadata-choices-1_0.html).  These plural versions would be along the lines of:

- `authorization_encrypted_response_alg_values_supported`: JSON array containing a list of the JWE `alg` values supported when encrypting responses.
- `authorization_encrypted_response_enc_values_supported`: JSON array containing a list of the JWE `enc` values supported when encrypting responses.

jogu:
We can potentially look at aligning with a "values_supported" style approach that aligns with https://openid.net/specs/openid-connect-rp-metadata-choices-1_0.html in the future.

closes #552

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -283,7 +283,8 @@ In the context of an authorization request according to [@RFC6749], parameters c
 : OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
     * `jwks`: OPTIONAL. A JSON Web Key Set, as defined in [@!RFC7591], that contains one or more public keys, such as those used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#response_encryption)), or where the Wallet will require the public key of the Verifier to generate a Verifiable Presentation. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests. Each JWK in the set MUST have a `kid` (Key ID) parameter that uniquely identifies the key within the context of the request.
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] `enc` algorithm is used to convey the requested content encryption algorithm for encrypting the authorization response. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present for anything other than the default value of `A128GCM`. Otherwise this SHOULD be absent. Note that because the response can be encrypted (see (#response_encryption)) but not signed, most of the general mechanisms of JARM do not apply. However, the  `authorization_encrypted_response_enc` metadata parameter from JARM is reused to avoid redefining it.
+    * `authorization_encrypted_response_alg`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] algorithm is used to convey the requested content encryption algorithm for encrypting the authorization response. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present. Otherwise this SHOULD be absent. Note that because the response can be encrypted (see (#response_encryption)) but not signed, most of the general mechanisms of JARM do not apply. However, the  `authorization_encrypted_response_alg` metadata parameter from JARM is reused to avoid redefining it.
+    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] `enc` algorithm is used to convey the requested content encryption algorithm for encrypting the authorization response. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present for anything other than the default value of `A128CBC-HS256`. Otherwise this SHOULD be absent. Note that because the response can be encrypted (see (#response_encryption)) but not signed, most of the general mechanisms of JARM do not apply. However, the  `authorization_encrypted_response_enc` metadata parameter from JARM is reused to avoid redefining it.
     * `vp_formats_supported`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
 
     Authoritative data the Wallet is able to obtain about the Client from other sources, for example those from an OpenID Federation Entity Statement, take precedence over the values passed in `client_metadata`.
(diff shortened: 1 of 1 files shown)


selfissued (approved):
Thanks, Joseph

peppelinux (approved):
Yes.

bc-pi:
I believe this would be a mistake that moves things in the wrong direction for the wrong reasons that will not be easy to undo or move away from. See also https://github.com/openid/OpenID4VP/issues/552#issuecomment-2821805454. But I honestly don't think I have the energy to engage on this.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3131,7 +3132,8 @@ The technology described in this specification was made available from contribut
 
    * rename `vp_formats` to `vp_format_supported` in Verifier Metadata
    * update the `vp_formats_supported` metadata to always be format specific, and explicitly define the structure for `mso_mdoc`, `jwt_vc_json` and `ldp_vc`.
-   * require fully-specified COSE and JOSE algoriths for `mso_mdoc` and `dc+sd-jwt` formats in `vp_formats_supported` metadata
+   * revert to usage of JARM's authorization_encrypted_response_alg for picking JWE encryption algorithm for now
+   * require fully-specified COSE and JOSE algorithms for `mso_mdoc` and `dc+sd-jwt` formats in `vp_formats_supported` metadata
```
Sakurann, line 3136:
```suggestion
   * require fully-specified COSE and JOSE algorithms for `mso_mdoc` and `dc+sd-jwt` formats in `vp_formats_supported` metadata
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1293,10 +1294,10 @@ This section defines how an Authorization Response containing a VP Token (such a
...
+The JWE `alg` algorithm used is obtained from the `authorization_encrypted_response_alg` parameter of client metadata, such as the `client_metadata` request parameter.
+The Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters.
+If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that was used to encrypt the response.
+The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128CBC-HS256` when not explicitly set.
```
Sakurann, line 1300:
```suggestion
Using what it supports and its preferences, the Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters.
The JWE `alg` algorithm used MUST the `alg` value of the chosen `jwk`, if present, or otherwise make sense to use with the selected key.
If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that used to encrypt the response.
The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128CBC-HS256` when not explictiy set.
```

tplooker:
Note - Disregard the prior approvals and comments on this PR, the scope of it has changed based on the WG discussion had on the 23rd of April, see changes from @Sakurann's commit for the new scope of change for this PR.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -283,7 +283,7 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
-    * `authorization_encrypted_response_enc`: OPTIONAL. As defined in [@!JARM]. The JWE [@!RFC7516] `enc` algorithm is used to convey the requested content encryption algorithm for encrypting the authorization response. When a `response_mode` requiring encryption of the Authorization Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present for anything other than the default value of `A128GCM`. Otherwise this SHOULD be absent. Note that because the response can be encrypted (see (#response_encryption)) but not signed, most of the general mechanisms of JARM do not apply. However, the  `authorization_encrypted_response_enc` metadata parameter from JARM is reused to avoid redefining it.
+    * `encrypted_response_enc`: OPTIONAL. Array of strings, where each string is the JWE [@!RFC7516] `enc` algorithm that is used to convey the requested content encryption algorithm for encrypting the Response. When a `response_mode` requiring encryption of the Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified this MUST be present for anything other than the default value of `A128GCM`. Otherwise this SHOULD be absent.
```
selfissued, line 286:
```suggestion
    * `encrypted_response_enc_values_supported`: OPTIONAL. Array of strings, where each string is a JWE [@!RFC7516] `enc` algorithm that can be used as the content encryption algorithm for encrypting the Response. When a `response_mode` requiring encryption of the Response (such as `dc_api.jwt` or `direct_post.jwt`) is specified, this MUST be present for anything other than the default single value of `A128GCM`. Otherwise, this SHOULD be absent.
```

The parameter name needs to be plural.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1293,10 +1293,11 @@ This section defines how an Authorization Response containing a VP Token (such a
 To encrypt the Authorization Response, implementations MUST use an unsigned, encrypted JWT as described in [@!RFC7519].
 
 To obtain the Verifier's public key to which to encrypt the Authorization Response, the Wallet uses keys from client metadata, such as the `jwks` member within the `client_metadata` request parameter, the metadata defined in the Entity Configuration if OpenID Federation is used, or other mechanisms.
-Using what it supports and its preferences, the Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters.
+Using what it supports and its preferences, the Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. 
+`alg` parameter MUST be present in the JWK.
```
selfissued, line 1297:
```suggestion
The `alg` parameter MUST be present in the JWK.
```

selfissued (changes requested):
Please apply my suggestions.  Then I should be able to approve.

openid-4-verifiable-presentations-1_0.md
bc-pi: Agree with @selfissued that the name should suggest that it's plural. 

Also acknowledging that naming is hard.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1293,10 +1293,11 @@ This section defines how an Authorization Response containing a VP Token (such a
...
-If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that used to encrypt the response.
-The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128CBC-HS256` when not explictiy set.
+If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that was used to encrypt the response.
+The JWE `enc` content encryption algorithm used is obtained from the `authorization_encrypted_response_enc` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128CBC-HS256` when not explicitly set.
```
bc-pi, line 1300: This sentence needs to be reconciled with the changes or suggested changes above around ~ https://github.com/openid/OpenID4VP/pull/555/files#r2054743816 
```suggestion
The JWE `enc` content encryption algorithm used is obtained from the `encrypted_response_enc_values_supported` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128GCM` when not explicitly set.
```

openid-4-verifiable-presentations-1_0.md
tplooker: Slight preference for `encrypted_response_enc_values_supported` over `encrypted_response_enc`.

Sakurann (approved):

c2bo (approved):
Only thing that I wondered about: If the Verifier provides multiple keys and the algorithm doesn't uniquely identify the selected key, then the Verifier should include a kid in the JWKs.

Seems somewhat obvious and unrealistic tbh and we mandate `kid` to be in the JWE if present, so should be fine as is?

selfissued (approved):

tplooker (approved):

Sakurann:
+ changelog needs to be moved to -28

PR #558, opened 2025-04-17
Replace uses of 'Relying Party' with 'Verifier'

### LINKED ISSUE #545: Inconsistencies in terminology
There are some inconsistencies in the usage of defined terminology, for example we introduce "Verifier", but use "Relying Party" in some parts of the spec.

jogu:
closes #545


c2bo (approved):
You were faster! 😄

danielfett (approved):

PR #563, opened 2025-04-17
add language for non-compliant JSON-LD processors

### LINKED ISSUE #527: Add DCQL type_values example for JSON-based W3C VCs
Current DCQL `type_values` query only has example for LD creds. We should also add a `type_values` example for the following:

```
{
    "@context": [
      "https://www.w3.org/2018/credentials/v1",
      "https://www.w3.org/2018/credentials/examples/v1"
    ],
    "type": [
      "VerifiableCredential",
      "IDCredential"
    ],
    "credentialSubject": {
      "given_name": "Max",
      "family_name": "Mustermann",
      "birthdate": "1998-01-11",
      "address": {
        "street_address": "Sandanger 25",
        "locality": "Musterstadt",
        "postal_code": "123456",
        "country": "DE"
      }
    }
  }
```

Note that the following `type_values` would match the credential above:

```
"type_values": [ "IDCredential" ]
```

Note that JSON-LD expansion would result in `IDCredential`, so the example in [B.1.3.1](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-B.1.3.1) is correct.

### LINKED ISSUE #547: Clarify JWT-VC JSON Text to Allow Alternatives to JSON-LD Expansion
For JWT-VC JSON, we should clarify the text to state that either JSON-LD expansion should be used (as currently specified), or that an equivalent mechanism that achieves the same result is also acceptable.

awoie:
This PR adds language for non-JSON-LD implementations that implement JWT VC (1.1). It should say that either do JSON-LD expansion and get the expanded type, or do something equivalent. Note that for JWT VC (1.1) the expanded type is typically a relative IRI since I assume that JWT VCs don't define their terms in the `@context`.

Note, if you run JSON-LD expansion on a JSON object where some values are not defined, the values would stay unchanged (still be relative IRIs) but some JSON-LD processors might return error since after expansion all of these values have to be absolute IRIs.

resolves #547 
resolves #527


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2362,7 +2362,7 @@ If `require_cryptographic_holder_binding` is set to `true` in the Credential Que
...
+: REQUIRED. An array of string arrays that specifies the fully expanded types (IRIs) after the `@context` was applied that the Verifier accepts to be presented in the Presentation. If a type value was not defined by any `@context`, its value stays the same. This works fine in practice, even if it's not strictly JSON-LD compliant. Implementations can use alternative mechanisms that produce the same outcome as applying the `@context`, provided that the result is equivalent. Each of the top-level arrays specifies one alternative to match the `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types.
```
c2bo, line 2365: Maybe I am missing something but the later part feels redundant?
```suggestion
: REQUIRED. An array of string arrays that specifies the fully expanded types (IRIs) after the `@context` was applied that the Verifier accepts to be presented in the Presentation. If a type value was not defined by any `@context`, its value stays the same. This works fine in practice, even if it's not strictly JSON-LD compliant. Implementations can use alternative mechanisms that produce the same outcome as applying the `@context`. Each of the top-level arrays specifies one alternative to match the `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

Sakurann, line 2365: few clarification questions:

> If a type value was not defined by any `@context`, its value stays the same. 

what value?

> This works fine in practice, even if it's not strictly JSON-LD compliant. 

i would just say `This works in practice` to make it a bit more formal and avoid talking about JSON-LD compliance..?

> Each of the top-level arrays specifies one alternative to match the `type` values of the Verifiable Credential against. 

which top level arrays..?

Sakurann:
@or13 would appreciate if you take a look at this one

awoie:
> @OR13 would appreciate if you take a look at this one

@OR13 @tplooker  This relates to W3C VCDM 1.1 credentials that don't define IRIs for type values. Those processors shouldn't do JSON-LD expansion to match the type value. I believe JSON-LD processors might fail since you would still have relative IRIs after expansion but that is ok for W3C VCDM 1.1 credentials that use only the default context URIs.

openid-4-verifiable-presentations-1_0.md
tplooker: > what value?

I believe what @awoie is referring to is the original type value in its non-expanded JSON-LD form

> i would just say This works in practice to make it a bit more formal and avoid talking about JSON-LD compliance..?

I would go one step further and not even say "works in practise" the fact we have defined this in the spec means implicitly it works.

OR13:
`@type` is aliased to `type`.

its valid for `@type` to contain both expanded and context defined IRIs:

[example](https://v.jsld.org/3xq5Wb8rWpZZheC2shyXCqdm5kRsE72wPNXn6typPtjic6NK7RLfZNqxSDrA7MLSkECVY758sRkaLSHZB9g1wD6piVy1wtzJ7kdAPfKgVf77V6KHQLkvgXvfnJZScvYCTRSt6WGsUi3itDkktmJo4gTYrW7mUyvttHCimNpRigCgXd21M3ZbmVNFSJ13rvADa6v1vkSjPqqEgJu182U9qGYYPDwU2285rhLyWiGvqvDoxsMuUAjqx)

openid-4-verifiable-presentations-1_0.md
tplooker:
```suggestion
: REQUIRED. An array of string arrays that specifies the fully expanded types (IRIs) after the `@context` was applied that the Verifier accepts to be presented in the Presentation. If a type value was not defined by any `@context`, its value stays the same as it is effectively a relative IRI. Implementations can use alternative mechanisms that produce the same outcome as applying the `@context`, provided that the result is equivalent. Each of the top-level arrays specifies one alternative to match the `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

Sakurann (approved):

tplooker (approved):

awoie:
I updated the PR. @OR13 @tplooker @Sakurann , could you please check if the latest language makes sense? I also provided an example for W3D VCDM 1.1 example that don't define the type value in a context hence this PR.

OR13:
A better solution is to check both the context values and the type, and never do JSON-LD processing unless you plan to reject the data when it fails.

When type is not an IRI, the full IRI is only acceptable via JSON-LD processing, which is dependent on the context, and can fail for many reasons.

JSON-LD processing is out of scope for this document, but a savvy verifier can use schema validation to ensure that the context and type are recognized before proceeding.

Also not sure why you are focusing on 1.1 when 2.0 is not the recommendation.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2362,7 +2362,7 @@ If `require_cryptographic_holder_binding` is set to `true` in the Credential Que
...
+: REQUIRED. An array of string arrays specifying the fully expanded types (IRIs) that the Verifier accepts in a Presentation, after applying the `@context` to the Verifiable Credential. If a `type` value in a Verifiable Credential is not defined in any `@context`, it remains unchanged, i.e., remains a relative IRI after JSON-LD processing. For this reason, JSON-LD processing MAY be skipped in such cases and the relative IRI is considered to be the fully expanded type, as applying the `@context` would not alter the value. Implementations MAY use alternative mechanisms to obtain the fully expanded types, as long as the results are equivalent to those produced by JSON-LD processing. Each of the top-level arrays specifies one alternative to match the fully expanded `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the fully expanded types in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types.
```
Sakurann, line 2365:
```suggestion
: REQUIRED. An array of strings, where each string specifies a fully expanded type (IRI) that the Verifier accepts in a Presentation, after applying the `@context` to the Verifiable Credential. If a `type` value in a Verifiable Credential is not defined in any `@context`, it remains unchanged, i.e., remains a relative IRI after JSON-LD processing. For this reason, JSON-LD processing MAY be skipped in such cases and the relative IRI is considered to be the fully expanded type, as applying the `@context` would not alter the value. Implementations MAY use alternative mechanisms to obtain the fully expanded types, as long as the results are equivalent to those produced by JSON-LD processing. Each of the top-level arrays specifies one alternative to match the fully expanded `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the fully expanded types in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

Sakurann (approved):
much clearer, thank you. just a nit

Sakurann:
@peacekeeper @alenhorvat @Or13 would appreciate your (re-)review.

OR13 (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2404,6 +2407,17 @@ The following is another non-normative example of a W3C Verifiable Credential th
...
 ```
 
+The following is another non-normative example of a W3C Verifiable Credential that would match the `type_values` DCQL query above (other claims omitted for readability):
+
+```json
+{
+  "@context": [
+    "https://www.w3.org/2018/credentials/v1"
+  ],
+  "type": ["VerifiableCredential", "IdentityCredential"]
```
OR13, line 2417: In this case, the "IdentityCredential" is silently dropped by JSON-LD Processors, such as the one used on json-ld playground:

https://json-ld.org/playground/

```
_:b0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> .
```

Try pasting the JSON-LD above in, and note the errors and silent dropping when changing tabs.

PR #566, opened 2025-04-17
Only reject the presentations that fail checks

### LINKED ISSUE #562: vp_token validation when it contains multiple presentations / credentials
Section 8.6 states that if any of the check fail, the full vp_token must be rejected. However this means that if there are multiple credentials in the response, one of which cannot be validated by the Verifier (e.g. because it doesn't have the trust chain), it also needs to reject the credentials that it can validate.

We should state that when an entry in the vp_token fails it checks, that entry must be rejected.

jogu:
Also replace an 'i.e.' with an 'e.g.' as I'm pretty sure this was meant to be an example of one possible check.

closes #562

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1409,10 +1409,10 @@ Verifiers MUST validate the VP Token in the following manner:
    1. Validate that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
    1. Validate that all Presentations contain a cryptographic proof of Holder Binding (i.e., that they are Verifiable Presentations), unless specifically requested otherwise.
    1. For Verifiable Presentations, validate the Holder Binding, including the checks required to present replay described in (#preventing-replay).
-   1. Perform the checks required by the Verifier's policy based on the set of trust requirements such as trust frameworks it belongs to (i.e., revocation checks), if applicable.
+   1. Perform the checks required by the Verifier's policy based on the set of trust requirements such as trust frameworks it belongs to (e.g., revocation checks), if applicable.
 1. Check that the set of Presentations returned satisfies all requirements defined in the Verifier's request as described in (#dcql_query_lang_processing_rules).
 
-If any of these checks fails, the VP Token MUST be rejected.
+If any of these checks fails, the relevant Presentations MUST be rejected.
 
 # Wallet Invocation {#wallet-invocation}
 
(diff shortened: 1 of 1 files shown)


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1409,10 +1409,10 @@ Verifiers MUST validate the VP Token in the following manner:
    1. Validate that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
    1. Validate that all Presentations contain a cryptographic proof of Holder Binding (i.e., that they are Verifiable Presentations), unless specifically requested otherwise.
    1. For Verifiable Presentations, validate the Holder Binding, including the checks required to present replay described in (#preventing-replay).
-   1. Perform the checks required by the Verifier's policy based on the set of trust requirements such as trust frameworks it belongs to (i.e., revocation checks), if applicable.
+   1. Perform the checks required by the Verifier's policy based on the set of trust requirements such as trust frameworks it belongs to (e.g., revocation checks), if applicable.
 1. Check that the set of Presentations returned satisfies all requirements defined in the Verifier's request as described in (#dcql_query_lang_processing_rules).
 
-If any of these checks fails, the VP Token MUST be rejected.
+If any of these checks fails, the relevant Presentations MUST be rejected.
```
c2bo, line 1415: Should we leave that part as is and move the logic to reject individual Presentations (if validation of a presentation fails) to the nested part? Since this whole section is about vp token validation that would feel cleaner imho.

Sakurann: i kind of agree with Christian's point

tplooker:
```suggestion
If any of the checks related to an individual Presentation fail, the effected Presentation MUST be discarded. If any of the checks pertaining to the VP Token, or the overall response fails, the VP Token MUST be rejected.  
```

What about this @c2bo @Sakurann @jogu ?

tplooker: Im trying to distinguish "discarding a presentation" from "rejection of the response/VP Token" as rejecting a presentation seems a bit confusing to me.

jogu: I've applied Tobias's suggestion - thanks!

c2bo (approved):

tplooker (approved):

Sakurann (approved):

GarethCOliver (approved):

martijnharing (approved):

paulbastian (approved):

PR #583, opened 2025-05-01
Add IANA considerations for encrypted_response_enc_values_supported

### LINKED ISSUE #573: Add IANA Considerations for encrypted_response_enc_values_supported
We need a client metadata IANA registration entry for the new client metadata added in https://github.com/openid/OpenID4VP/pull/555

jogu:
closes #573

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -3155,6 +3155,13 @@ This specification registers the following client metadata parameters
 in the IANA "OAuth Dynamic Client Registration Metadata" registry [@IANA.OAuth.Parameters]
 established by [@!RFC7591].
 
+### encrypted_response_enc_values_supported
+
+* Client Metadata Name: `encrypted_response_enc_values_supported`
+* Client Metadata Description: Array of strings, where each string is a JWE [@!RFC7516] `enc` algorithm that can be used as the content encryption algorithm for encrypting the Response
+* Change Controller: OpenID Foundation Digital Credentials Protocols Working Group - openid-specs-digital-credentials-protocols@lists.openid.net
+* Reference: (#new_parameters) of this specification
+
 ### vp_formats_supported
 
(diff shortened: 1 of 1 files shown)


selfissued (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -3155,6 +3155,13 @@ This specification registers the following client metadata parameters
 in the IANA "OAuth Dynamic Client Registration Metadata" registry [@IANA.OAuth.Parameters]
 established by [@!RFC7591].
 
+### encrypted_response_enc_values_supported
+
+* Client Metadata Name: `encrypted_response_enc_values_supported`
+* Client Metadata Description: Array of strings, where each string is a JWE [@!RFC7516] `enc` algorithm that can be used as the content encryption algorithm for encrypting the Response
```
c2bo, line 3161:
```suggestion
* Client Metadata Description: An array of strings, where each string is a JWE [@!RFC7516] `enc` algorithm that can be used as the content encryption algorithm for encrypting the Response
```

c2bo (approved):
minor nitpick

Not in scope of this PR, but I just realised that the `verifier_attestations` entry looks wrong for this section and I think should be moved to the OAuth Parameters Registry?

jogu:
> minor nitpick
> 
> Not in scope of this PR, but I just realised that the `verifier_attestations` entry looks wrong for this section and I think should be moved to the OAuth Parameters Registry?

That's a rather good point :) Separate issue opened.

openid-4-verifiable-presentations-1_0.md
jogu: Thanks for the suggestion - I thought about this and decided to leave it as it is for consistency with other entries in https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#client-metadata which all seem to omit a particle.

PR #589, opened 2025-05-02
Explicitly state that various arrays need to be non-empty

### LINKED ISSUE #572: Clarify meaning of empty Array in Section 6.3 (DCQL `values`)
In Section 6.3 https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-claims-query it says

> If the values property is present, the Wallet SHOULD return the claim only if the type and value of the claim both match exactly for at least one of the elements in the array

However this does not really specify, what the wallet should do, if the array is present but empty. 

In my opinion, as it is written now, it basically means that it can never match, because the presence in the claims query asks for the disclosure of that property, but it can never match to nothing. 

So I would propose to clarify 
>  An array of strings

 to

> A _non_ empty array of strings

 or something similiar.

jogu:
Also remove 'JSON' from various instances of 'JSON array' as in the DC API case it's not JSON and the JSON qualifier doesn't really add anything anyway.

closes #572


openid-4-verifiable-presentations-1_0.md
```diff
@@ -2424,7 +2424,7 @@ If `require_cryptographic_holder_binding` is set to `true` in the Credential Que
...
+: REQUIRED. A non-empty array of string arrays specifying the fully expanded types (IRIs) that the Verifier accepts in a Presentation, after applying the `@context` to the Verifiable Credential. If a `type` value in a Verifiable Credential is not defined in any `@context`, it remains unchanged, i.e., remains a relative IRI after JSON-LD processing. For this reason, JSON-LD processing MAY be skipped in such cases and the relative IRI is considered to be the fully expanded type, as applying the `@context` would not alter the value. Implementations MAY use alternative mechanisms to obtain the fully expanded types, as long as the results are equivalent to those produced by JSON-LD processing. Each of the top-level arrays specifies one alternative to match the fully expanded `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the fully expanded types in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types.
```
charsleysa, line 2427: Should this be more explicit like `options` in the DCQL Credential Set Query? e.g. Will there be ambiguity that this is allowed? `[ [] ]`

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1451,7 +1451,7 @@ The following is a non-normative example of a `vp_formats_supported` parameter:
 ```
 
 `client_id_prefixes_supported`:
-: OPTIONAL. Array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `openid_federation`, `verifier_attestation`, `decentralized_identifier`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles or extensions of this specification.
+: OPTIONAL. Non-empty array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `openid_federation`, `verifier_attestation`, `decentralized_identifier`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles or extensions of this specification.
```
TimoGlastra, line 1454:
```suggestion
: OPTIONAL. A non-empty array of strings containing the values of the Client Identifier Prefixes that the Wallet supports. The values defined by this specification are `pre-registered` (which represents the behavior when no Client Identifier Prefix is used), `redirect_uri`, `openid_federation`, `verifier_attestation`, `decentralized_identifier`, `x509_san_dns` and `x509_hash`. If omitted, the default value is `pre-registered`. Other values may be used when defined in the profiles or extensions of this specification.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2839,17 +2839,17 @@ The transaction data mechanism requires use of an SD-JWT VC with Cryptographic H
...
 * The Key Binding JWT in the response includes the following top level parameters:
-  * `transaction_data_hashes`: Array of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
+  * `transaction_data_hashes`: Non-empty array of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
```
TimoGlastra, line 2844:
```suggestion
  * `transaction_data_hashes`: A non-empty array of hashes, where each hash is calculated using a hash function over the data in the strings received in the `transaction_data` request parameter. Each hash value ensures the integrity of, and maps to, the respective transaction data object. If `transaction_data_hashes_alg` was specified in the request, the hash function MUST be one of its values. If `transaction_data_hashes_alg` was not specified in the request, the hash function MUST be `sha-256`.
```

TimoGlastra (approved):

openid-4-verifiable-presentations-1_0.md
jogu: @charsleysa Good point. Would adding an extra 'non-empty' solve this:

```suggestion
: REQUIRED. A non-empty array of non-empty string arrays specifying the fully expanded types (IRIs) that the Verifier accepts in a Presentation, after applying the `@context` to the Verifiable Credential. If a `type` value in a Verifiable Credential is not defined in any `@context`, it remains unchanged, i.e., remains a relative IRI after JSON-LD processing. For this reason, JSON-LD processing MAY be skipped in such cases and the relative IRI is considered to be the fully expanded type, as applying the `@context` would not alter the value. Implementations MAY use alternative mechanisms to obtain the fully expanded types, as long as the results are equivalent to those produced by JSON-LD processing. Each of the top-level arrays specifies one alternative to match the fully expanded `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the fully expanded types in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

charsleysa: I thought about that, but it still feels a little ambiguous. Would it be better to have something like this (which is similar to what's done for `options` in the DCQL Credential Set Query)?
```suggestion
: REQUIRED. A non-empty array of string arrays. The value of each element in the `type_values` array is a non-empty array specifying the fully expanded types (IRIs) that the Verifier accepts in a Presentation, after applying the `@context` to the Verifiable Credential. If a `type` value in a Verifiable Credential is not defined in any `@context`, it remains unchanged, i.e., remains a relative IRI after JSON-LD processing. For this reason, JSON-LD processing MAY be skipped in such cases and the relative IRI is considered to be the fully expanded type, as applying the `@context` would not alter the value. Implementations MAY use alternative mechanisms to obtain the fully expanded types, as long as the results are equivalent to those produced by JSON-LD processing. Each of the top-level arrays specifies one alternative to match the fully expanded `type` values of the Verifiable Credential against. Each inner array specifies a set of fully expanded types that MUST be present in the fully expanded types in the `type` property of the Verifiable Credential, regardless of order or the presence of additional types. 
```

jogu: That looks good - thanks!

charsleysa (approved):

c2bo (approved):

paulbastian (approved):

awoie (approved):

PR #596, opened 2025-05-12
ed: fix title in 23220-2 reference

awoie:
23220-2 became a TS, so we can reference the TS instead of the DTS version


Sakurann (approved):
funny that URL does not change

PR #597, opened 2025-05-13
Add encryption details parameter

### LINKED ISSUE #347: Using 'detached' session information as part of encryption key derivation to solve 'lazy verifier' problem
The current encryption methods supported in OID4VP require that any information that is used as part of the key derivation is included in the JWE header (e.g. `apu` and `apv`). This is basically how the JWE specs and libraries currently work and it's not really possible to not send this information and be compliant with the current JWE RFCs.

It seems it be advantageous to not send some of this information in the JWE and to instead have the verifier provide it from it's view of the world - in particular if information like the web origin, keys, etc, where included in the key derivation and not transmitted, then the verifier would hopefully fail to decrypt a response that has been injected into the wrong session.

Given we have seen a long history of OAuth clients not implementing security checks like session binding correctly, a mechanism where a response from a different session simply fails to decrypt is highly advantageous from an "overall security of an ecosystem" point of view.

martijnharing:
Resolves #347


c2bo:
We would need to add IANA registration requests for both parameters in their respective registries I assume?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1301,6 +1302,8 @@ The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen `jwk
 If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that was used to encrypt the response.
 The JWE `enc` content encryption algorithm used is obtained from the `encrypted_response_enc_values_supported` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128GCM` when not explicitly set.
 
+When the `encryption_details_supported parameter` `client_metadata` parameter is used in the request (see #new_parameters) and requires the Wallet to include a referenced profile in the response, the selected profile can be included in the JWE protected header using the `encryption_details` parameter. When present, the `encryption_details` parameter MUST contain a string of the selected profile.
```
awoie, line 1305:
```suggestion
When the `encryption_details_supported` parameter `client_metadata` parameter is used in the request (see #new_parameters) and requires the Wallet to include a referenced profile in the response, the selected profile can be included in the JWE protected header using the `encryption_details` parameter. When present, the `encryption_details` parameter MUST contain a string of the selected profile.
```

awoie: Is this a Verifier metadata parameter that is only used in the request, or is it a new Verifier metadata? In the latter case (e.g., OpenID Federation potentially may want that but may not), we would need to add this to section [11. ](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-11)[Verifier Metadata (Client Metadata)](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-verifier-metadata-client-me).

awoie:
> We would need to add IANA registration requests for both parameters in their respective registries I assume?

This needs registration in two IANA registries:
- `encryption_details_supported_parameter` -> [E.4. ](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#appendix-E.4)[OAuth Authorization Server Metadata Registry](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#name-oauth-authorization-server-)
- `encryption_details` -> if I read the PR correctly, then this has to go into [JSON Web Signature and Encryption Header Parameters
](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-header-parameters)

openid-4-verifiable-presentations-1_0.md
```diff
@@ -287,6 +287,7 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats_supported`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
+    * `encryption_details_supported`: OPTIONAL. A non-empty array of strings, where each string references a profile of encryption details that the Verifier supports. The meaning and expected behavior of the Verifier and Wallet is profile specific and out of scope of this specification. When the Verifier includes this parameter, the Wallet MUST either reject the transaction or include one of the profile referenced in the response and implement all of the requirements defined by the referenced profile (see #response_encryption for details of how to include it in the JWE in the response). When the Verifier includes this parameter, it MUST verify that one of the referenced profiles is included in the response and it MUST implement all of the requirements required by the profile included in the response.
```
awoie, line 290:
```suggestion
    * `encryption_details_supported`: OPTIONAL. A non-empty array of strings, where each string references a profile of encryption details that the Verifier supports. The meaning and expected behavior of the Verifier and Wallet is profile-specific and out of scope of this specification. When the Verifier includes this parameter, the Wallet MUST either reject the transaction or include one of the profile referenced in the response and implement all of the requirements defined by the referenced profile (see #response_encryption for details of how to include it in the JWE in the response). When the Verifier includes this parameter, it MUST verify that one of the referenced profiles is included in the response and it MUST implement all of the requirements required by the profile included in the response.
```

openid-4-verifiable-presentations-1_0.md
Sakurann: > Is this a Verifier metadata parameter that is only used in the request, or is it a new Verifier metadata?

what's the difference to be honest? should we move vp_formats_supported to client-metadata section and just point to client_metadata section from section 11.1?

openid-4-verifiable-presentations-1_0.md
scvenema: Suggest the following rewording (words to **add* and ~~delete~~):
"When the Verifier includes this parameter **in a request**, the Wallet MUST either reject the ~~transaction~~ **request** or include one of the ~~profile~~ referenced **profiles** in the response...

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1301,6 +1302,8 @@ The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen `jwk
 If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that was used to encrypt the response.
 The JWE `enc` content encryption algorithm used is obtained from the `encrypted_response_enc_values_supported` parameter of client metadata, such as the `client_metadata` request parameter, allowing for the default value of `A128GCM` when not explicitly set.
 
+When the `encryption_details_supported` parameter is present in the `client_metadata` parameter of the request (see #new_parameters) and requires the Wallet to include a referenced profile in the response, the selected profile MAY be included in the JWE protected header using the `encryption_details` header. When present, the `encryption_details` header MUST contain a string of the selected profile.
```
awoie, line 1305: Any reason this is a MAY and not a SHOULD?
```suggestion
When the `encryption_details_supported` parameter is present in the `client_metadata` parameter of the request (see #new_parameters) and requires the Wallet to include a referenced profile in the response, the selected profile SHOULD be included in the JWE protected header using the `encryption_details` header. When present, the `encryption_details` header MUST contain a string of the selected profile.
```

awoie (changes requested):
The following two statements allow for a gap in implementations that lead to broken flows, right?

> the Verifier MUST verify that one of the referenced profiles is included in the response 

AND

> the selected profile MAY be included in the JWE protected header using the `encryption_details` header

Why would the latter be a MAY/SHOULD, if it would break the verifier?

selfissued (changes requested):
This is not specified in sufficient detail to enable interoperable implementations.

What are the parameters intended to be passed with this parameter and what are the processing rules for them?

awoie:
@martijnharing Generally fine with this PR but just had a question/suggestion above.

c2bo:
Would it make sense to define a "profile" name for the current/default behaviour?

martijnharing:
> The following two statements allow for a gap in implementations that lead to broken flows, right?
> 
> > the Verifier MUST verify that one of the referenced profiles is included in the response
> 
> AND
> 
> > the selected profile MAY be included in the JWE protected header using the `encryption_details` header
> 
> Why would the latter be a MAY/SHOULD, if it would break the verifier?

How about:
"The selected profile **MUST** be included in the JWE protected header using the `encryption_details` header **unless the profile defines differently**."

openid-4-verifiable-presentations-1_0.md
```diff
@@ -287,6 +287,7 @@ In the context of an authorization request according to [@RFC6749], parameters c
...
     * `vp_formats_supported`: REQUIRED when not available to the Wallet via another mechanism. As defined in (#client_metadata_parameters).
+    * `encryption_details_supported`: OPTIONAL. A non-empty array of strings, where each string references a profile of encryption details that the Verifier supports. The meaning and expected behavior of the Verifier and Wallet is profile-specific and out of scope of this specification. When the Verifier includes this parameter in a request, the Wallet MUST either reject the request or include one of the referenced profiles in the response and implement all of the requirements defined by the referenced profile (see (#response_encryption) for details of how to include it in the JWE in the response). When the Verifier includes this parameter in a request, the Verifier MUST verify that one of the referenced profiles is included in the response and MUST implement all of the requirements required by the profile included in the response.
```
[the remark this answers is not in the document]
bc-pi: I would very much like to understand what value is actually added by introducing an under-defined but interoperability hampering extension point?

bc-pi (approved):
As others have noted, precedent and consistency would require that these new things have IANA registration requests for both in their respective registries. Writing those IANA requests and taking them to the designated experts and IANA might underscore the rather tenuous nature of the problem and solution here.

bc-pi (changes requested):
As others have noted, precedent and consistency would require that these new things have IANA registration requests for both in their respective registries. Writing those IANA requests and taking them to the designated experts and IANA might underscore the rather tenuous nature of the problem and solution here.

openid-4-verifiable-presentations-1_0.md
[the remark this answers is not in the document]
bc-pi: > * @c2bo to make a suggestion for an exact wording on how we describe "default behavior"

I'd furthermore argue that there has to be a way to indicate that the "default behavior" is also an acceptable referenced profile. Otherwise there's going to be a rather sharp interoperability edge.

Sakurann:
WG discussion:
- we cannot define apu apv values in the profiles of openid4vp spec because then we need a parameter in openid4vp that indicates encryption parameter, which adds somewhat unnecessary complexity/optionality + less encryption profiles the better.
- apu - do not define it; apv - base64urlencoded hash string of following string: nonce**origin OR client_id**jwkthumbprint**response_uri** in that order 
    - JWE already says apv and apu is base64url-encoded string, so need to make sure it's not double encoded
- noting concerns of @bc-pi regarding the value of doing this + error proneness of multiple value as apv

Sakurann:
WG discussion about defining apu/apv values for ECDH-ES JWE..
- on one hand...
    - adds marginal security benefit (majority agrees with @selfissued and @bc-pi disagreeing)
    - would not be the most complex part of the OpenID4VP implementation and can be implemented using existing JOSE libraries
    - there is implementation experience of similar mechanisms in 18013-7 annex B 
- on the other hand...
    - there are strong concerns that adding this mechanism last minute would result in it being underspecified leading to interoperability problems
    - extension mechanism to define apu/apv values can be added in 1.1 (verifiers that support only that extension will potentially reject responses from wallets that do not support it)
    - wg seems to agree to define the content of a detached payload for JOSE-HPKE in OpenID4VP 1.1 or HAIP

wg agreed to discuss and tackle #347 during wglc/review period and as an outcome this PR and PR #628 will be closed for the above reasons since there is no wg rough consensus to proceed with either of them.

PR #598, opened 2025-05-13
Implement signature verification requirements

### LINKED ISSUE #561: Clarify when signature verification is not required
We discussed verification of signed requests here: https://github.com/openid/OpenID4VP/issues/395 where we agreed that it's not required by the OpenID4VP specification (but of course can be by a profile) to validate signed requests. We made changes to the specification, but I don't think we actually added the text that says verification is optional.
There are also other cases where signatures are used, where we don't want to always require the wallet to validate a signature and otherwise reset.

Looking at the current text, I think it does actually require verification of certain signatures. While these signatures should definitely be verified in certain scenarios, we shouldn't always require validation of those signatures.

These include:

section 5.9.3
verifier_attestation: "he Wallet MUST validate the signature on the Verifier attestation JWT." ... "If the Wallet cannot establish trust, it MUST refuse the request."
x509_san_dns: "The Wallet MUST validate the signature and the trust chain of the X.509 certificate."
x509_hash: "The Wallet MUST validate the signature and the trust chain of the X.509 leaf certificate."

I think it's sufficient to state that these requirements only apply when the corresponding signature is also verified stating that verification of the request signatures is optional, but we should make that explicit.

In verifier_attestations, data: "The Wallet MUST validate this signature and ensure binding."

In this case we should state that verification is required if the information is used.

martijnharing:
Resolves #561


Sakurann:
WG discussion:
- add a generic text that applies to all client identifier schemes that clarifies the behavior

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,7 +320,7 @@ The following is a non-normative example of a transaction data content, after ba
...
 
     * `format`: REQUIRED. A string that identifies the format of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers. Further processing of the attestation is determined by the type of the attestation, which is specified in a format-specific way. 
-    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. The Wallet MUST validate this signature and ensure binding.
+    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. Whether the Wallet uses the information from the verifier_attestation is out of scope of this specification, but if the Wallet uses the information from the verifier attestation the Wallet MUST validate this signature and ensure binding.
```
tlodderstedt, line 323:
```suggestion
    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. In case of using OpenID4VP over DC API it is at the discretion of the wallet whether it uses the information from the verifier_attestation, but if the Wallet uses the information from the verifier attestation the Wallet MUST validate this signature and ensure binding.
```

tlodderstedt (changes requested):
The currently proposed text does not distinguish between DC API and vanilla OpenID4VP. Made a suggestion how it could be modified to differentiate.

openid-4-verifiable-presentations-1_0.md
martijnharing: Clarified the text to bring it in line with the client identifier text, but did not make it only applicable in the context of the DC API.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -320,7 +320,7 @@ The following is a non-normative example of a transaction data content, after ba
...
 
     * `format`: REQUIRED. A string that identifies the format of the attestation and how it is encoded. Ecosystems SHOULD use collision-resistant identifiers. Further processing of the attestation is determined by the type of the attestation, which is specified in a format-specific way. 
-    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. The Wallet MUST validate this signature and ensure binding.
+    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. It is at the discretion of the Wallet (for example based on policies from ecosystems, profiles, issuers or Holders) whether it uses the information from `verifier_attestations`. If the Wallet uses information from `verifier_attestations`, the Wallet must MUST validate the signature and ensure binding.
```
c2bo, line 323:
```suggestion
    * `data`: REQUIRED. An object or string containing an attestation (e.g. a JWT). The payload structure is defined on a per format level. It is at the discretion of the Wallet (for example based on policies from ecosystems, profiles, issuers or Holders) whether it uses the information from `verifier_attestations`. If the Wallet uses information from `verifier_attestations`, the Wallet MUST validate the signature and ensure binding.
```

openid-4-verifiable-presentations-1_0.md
```diff
@@ -559,7 +559,9 @@ From this definition, it follows that pre-registered clients MUST NOT contain a
 
 ### Defined Client Identifier Prefixes {#client_identifier_prefixes}
 
-This specification defines the following Client Identifier Prefixes, followed by the examples where applicable: 
+This specification defines the following Client Identifier Prefixes, followed by the examples where applicable. 
+
+In case of using OpenID4VP over DC API as defined in #dc_api it is at the discretion of the Wallet (for example based on policies from ecosystems, profiles, issuers or Holders) whether it uses the information from a client identifier. If the Wallet uses information from a client identifier, the Wallet must perform all the verification requirements for that client identifier as defined below. If it doesn't use the information from a client identifier, the Wallet does not need to implement the verification requirements for that client identifier as defined below.
```
Sakurann, line 564:
```suggestion
In case of using OpenID4VP over DC API as defined in (#dc_api) it is at the discretion of the Wallet (for example based on policies from ecosystems, profiles, issuers or Holders) whether it uses the information from a client identifier. If the Wallet uses information from a client identifier, the Wallet must perform all the verification requirements for that client identifier as defined below. If it doesn't use the information from a client identifier, the Wallet does not need to implement the verification requirements for that client identifier as defined below.
```

awoie (commented):
I'm generally ok with this direction. Editorially, I think the DC API exception should go into the DC API Annex since the Annex is supposed to be self-contained. 

Something like this:

> The client_id parameter MUST be omitted in unsigned requests defined in [Appendix A.3.1](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#unsigned_request). The Wallet MUST ignore any client_id parameter that is present in an unsigned request.

> Parameters defined by a specific Client Identifier Prefix (such as the trust_chain parameter for the OpenID Federation Client Identifier Prefix) are also supported over the W3C Digital Credentials API.

> The client_id parameter MUST be present in signed requests defined in [Appendix A.3.2](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#signed_request), as it communicates to the wallet which Client Identifier Prefix and Client Identifier to use when authenticating the client through verification of the request signature or retrieving client metadata.

NEW:
> In case of using OpenID4VP over DC API as defined in (#dc_api) it is at the discretion of the Wallet (for example based on policies from ecosystems, profiles, issuers or Holders) whether it uses the information from a client identifier. If the Wallet uses information from a client identifier, the Wallet must perform all the verification requirements for that client identifier as defined below. If it doesn't use the information from a client identifier, the Wallet does not need to implement the verification requirements for that client identifier as defined below.

tlodderstedt (commented):
Will approve pending Kristina's proposed change is incorporated.

c2bo (commented):
+1 for Kristina's suggestions. I don't think it's super important but I'd also agree with Oliver that having the DC API specific parts in the DCP section makes sense.

Sakurann:
wg discussion
@martijnharing please incorporate Kristina's suggestions and we can merge

Sakurann (approved):

c2bo (approved):

tlodderstedt (approved):

tplooker (approved):

PR #605, opened 2025-05-14
Adjusted reference for single/multi sign requests and streamlined protocol identifier text

### LINKED ISSUE #604: Editorial: protocol identifier references for multi-signed need to be adjusted
This text needs to be adjusted slightly and the following bolded text should be added:

The value 1 MUST be used for the <version> field to indicate the request and response are compatible with this version of the specification. For <request-type>, unsigned requests, as defined in Appendix A.3.1, MUST use unsigned, signed requests, as defined in **Appendix A.3.2.1**, MUST use signed, **and multi-signed requests, as defined in Appendix A.3.2.2, MUST use multisigned**.

awoie:
Fixes #604


jogu (approved):

bc-pi (approved):

jogu:
Non-normative, 3 approvals and discussed yesterday's WG call - merging.

PR #610, opened 2025-05-19
Add session transcript for redirect-based oid4vp flow

### LINKED ISSUE #402: Define session transcript for OpenID4VP (without DC API) in OpenID4VP

awoie:
This PR fixes #402 

Also beautified CDDL (editorial)

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -531,7 +531,7 @@ The Client Identifier Prefix is a string that MAY be communicated by the Verifie
 
 ### Syntax
 
-In the `client_id` Authorization Request parameter and other places where the Client Identifier is used, the Client Identifiers are prefixed to the usual Client Identifier, separated by a `:` (colon) character:
+In the `client_id` Authorization Request parameter and other places where the Client Identifier is used, the Client Identifier Prefixes are prefixed to the usual Client Identifier, separated by a `:` (colon) character:
 
 ```
 <client_id_prefix>:<orig_client_id>
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2756,7 +2756,7 @@ OpenID4VPHandoverInfoBytes = bstr .cbor OpenID4VPHandoverInfo
 OpenID4VPHandoverInfo = [
   clientId,
   nonce,
-  jwk_thumbprint,
+  jwkThumbprint,
   responseUri  
 ] ; Array containing handover parameters
 
@@ -2764,7 +2764,7 @@ clientId = tstr
 
 nonce = tstr
 
-jwk_thumbprint = bstr
+jwkThumbprint = bstr
 
 responseUri = tstr
 ```
@@ -2886,14 +2886,14 @@ OpenID4VPDCAPIHandoverInfoBytes = bstr .cbor OpenID4VPDCAPIHandoverInfo
 OpenID4VPDCAPIHandoverInfo = [
   origin,
   nonce,
-  jwk_thumbprint
+  jwkThumbprint
 ] ; Array containing handover parameters
 
 origin = tstr
 
 nonce = tstr
 
-jwk_thumbprint = bstr
+jwkThumbprint = bstr
 ```
 
 The `OpenID4VPDCAPIHandover` structure has the following elements:
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -2733,14 +2733,15 @@ The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as d
 
 ### `Handover` and `SessionTranscript` Definitions
 
-#### Invocation via Redirects and HTTP POST
+#### Invocation via Redirects
 
-If the presentation request is invoked using the Digital Credentials API, the `SessionTranscript` CBOR structure as defined in Section 9.1.5.1 in [@ISO.18013-5] MUST be used with the following changes:
+If the presentation request is invoked using redirects, the `SessionTranscript` CBOR structure as defined in Section 9.1.5.1 in [@ISO.18013-5] MUST be used with the following changes:
 
 * `DeviceEngagementBytes` MUST be `null`.
 * `EReaderKeyBytes` MUST be `null`.
 * `Handover` MUST be the `OpenID4VPHandover` CBOR structure as defined below.
 
+```cddl
 OpenID4VPHandover = [
   "OpenID4VPHandover", ; A fixed identifier for this handover type
   OpenID4VPHandoverInfoHash ; A cryptographic hash of OpenID4VPHandoverInfo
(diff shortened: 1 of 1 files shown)


jogu (approved):
For info for others, this is different to the handover structure in 18013-7 annex b, but it's quite closely aligned with what we have for DC API. To me the difference looks like origin is replaced with client_id + response_uri, which seems reasonable.

TimoGlastra (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2733,7 +2734,135 @@ The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as d
...
+
+responseUri = tstr
+```
+
+The `OpenID4VPHandover` structure has the following elements:
+
+* The first element MUST be the string `OpenID4VPHandover`. This serves as a unique identifier for the handover structure to prevent misinterpretation or confusion.
+* The second element MUST be a Byte String which contains the sha-256 hash of the bytes of `OpenID4VPHandoverInfo` when encoded as CBOR.
+* The `OpenID4VPHandoverInfo` has the following elements:
+  * The first element MUST be the `client_id` request parameter. If the request is signed, the `client_id` MUST be obtained from the signed Request Object.
```
andprian, line 2778: And if the request is not signed?

awoie: not signed case:
 The first element MUST be the `client_id` request parameter

signed case:
 If the request is signed, the `client_id` MUST be obtained from the signed Request Object.

Should we make this clearer?

andprian: I think it would be very useful to state that the first sentence only applies to the unsigned case. We could even say the client_id parameter in the Authorization Request if we want to be very precise.

Sakurann:
```suggestion
  * The first element MUST be the `client_id` request parameter. When the request is signed, the `client_id` MUST be obtained from the signed Request Object.
```
not sure if this helps

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2733,7 +2734,135 @@ The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as d
...
+  responseUri  
+] ; Array containing handover parameters
+
+clientId = tstr
+
+nonce = tstr
+
+jwkThumbprint = bstr
+
+responseUri = tstr
```
Sakurann, line 2770: why do we need to put responseUri in the handover? is this an ISO requirement? we don't have this for other credential formats, do we?

tplooker (approved):
LGTM

openid-4-verifiable-presentations-1_0.md
awoie: @andprian How about this? This uses the same terminology as other places in the spec:
```suggestion
  * The first element MUST be the Client Identifier. If applicable, this includes the Client Identifier Prefix. If the request is unsigned, the Client Identifier is taken from the `client_id` Authorization Request parameter. If the request is signed, the `client_id` MUST be obtained from the signed Request Object
```

andprian: @awoie, it looks great

awoie:
@andprian I updated the PR again to clarify that all request parameters (also note that JAR says the Request Object contains authorization request parameters hence it is ok to refer to them as request parameters) have to be obtained from the authz request query params unless the request is signed and then the params have to obtained from the signed request object.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2733,7 +2734,137 @@ The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as d
...
+; Contains the sha-256 hash of OpenID4VPHandoverInfoBytes
+OpenID4VPHandoverInfoHash = bstr
+
+; Contains the bytes of OpenID4VPHandoverInfo encoded as CBOR
+OpenID4VPHandoverInfoBytes = bstr .cbor OpenID4VPHandoverInfo
+
+OpenID4VPHandoverInfo = [
+  clientId,
+  nonce,
+  jwkThumbprint,
```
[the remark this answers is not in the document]
awoie: What are the cases where the encryption key is not provided as JWK? My assumption is it would be always the JWK from the `client_metadata`.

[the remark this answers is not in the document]
Sakurann: jwk thumbprint is for the cases where the encryption key is ~~not~ provided as JWK, right?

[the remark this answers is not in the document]
c2bo: > What are the cases where the encryption key is not provided as JWK? My assumption is it would be always the JWK from the `client_metadata`. 

I would agree: within the flows defined here, I would expect the public key to come in JWK form from issuer metadata

c2bo (commented):
PR looks fine and I quickly skimmed over the CBOR content and that looks good to me.

One thing that feels a bit weird is the asymmetry between the 2 session transcripts (DC API and non-DC API).

andprian (approved):

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2733,7 +2734,137 @@ The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as d
 
 ### `Handover` and `SessionTranscript` Definitions
 
-#### Invocation via the Digital Credentials API
+#### Invocation via Redirects
```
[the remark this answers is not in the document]
awoie: Any objections to using the term "invocation via redirects"?

Possible alternatives:
- Invocation via Authorization Endpoints
- Invocation via URLs
- Invocation via Vanilla OpenID4VP

openid-4-verifiable-presentations-1_0.md
```diff
@@ -2733,7 +2734,137 @@ The VP Token contains the base64url-encoded `DeviceResponse` CBOR structure as d
...
+]
+
+; Contains the sha-256 hash of OpenID4VPHandoverInfoBytes
+OpenID4VPHandoverInfoHash = bstr
+
+; Contains the bytes of OpenID4VPHandoverInfo encoded as CBOR
+OpenID4VPHandoverInfoBytes = bstr .cbor OpenID4VPHandoverInfo
+
+OpenID4VPHandoverInfo = [
+  clientId,
```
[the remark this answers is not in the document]
c2bo: multi-auth is currently not supported for the vanilla flows (only DC API), but yes it would break them.
That is one of the things that feel odd between DC API and vanilla flow right now to me.

Sakurann:
WG discussion:
- merge this PR as-is, but...
- in a separate PR, add implementation considerations section that describes that in non-DC API flew, multi RP in one request is not supported and that RP might have to send multiple requests if it does not know which trust framework the wallet belongs to. @GarethCOliver will do a PR

PR #614, opened 2025-05-27
Clarify 'successfully processed' for direct_post response mode (#137)

### LINKED ISSUE #137: Clarify what successfully processing
This assumes response mode direct_post and the following clause in the OpenID 4 VP spec (Section 6.2)

> If the Response Endpoint has successfully processed the request, it MUST respond with HTTPS status code 200.

The word _successfully_ does a lot of the heavy lifting and is open for interpretation

Imagine the following: A wallet posts a response, the response is structurally correct and the submission requirements meet the definition. But, the verifier notices that the credential is expired, or claim values do not match the functional needs, etc, and for this (imaginary) case this is not acceptable.

I believe that there are two options here

1. Return 200 (w. redirect_uri)
2. Return 400 (w. an error response and redirect_uri (and potentially other attr to allow resubmit))

I believe it is reasonable to assume that the Wallet receives 400 from the Response Endpoint, allowing for the possibility to correct and re-submit with different credentials (or whatever the case may be). If 200 is always returned, it is more difficult to make the distinction between a success state and one that is not from the perspective of the Wallet. At least, this is what we have experienced during our implementation.

subhra-io:
This PR addresses issue #137 by adding/updating a paragraph in the Response Endpoint section (or [specify section, e.g., new subsection under 6. Response]) to clarify the definition of "successfully processed" for the direct_post response mode. The text specifies that HTTP 200 indicates protocol compliance, while HTTP 400 with error details is used for invalid credentials or functional mismatches, supporting wallet resubmission. Note: Issue #137 references Section 6.2, but the current draft has Section 6.2 as "Credential Set Query," so the change was applied to [specify section or note if new].
Closes #137

I’m working on this issue but noticed Section 6.2 is "Credential Set Query" in the current draft, not related to the Response Endpoint or direct_post. Could you clarify the correct section? I propose adding the clarification under [specify section, e.g., Response Endpoint or new subsection].

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1201,6 +1201,8 @@ The Response Mode is defined in accordance with [@!OAuth.Responses] as follows:
 
 The following new Authorization Request parameter is defined to be used in conjunction with Response Mode `direct_post`: 
 
+If the Response Endpoint receives a request in the direct_post response mode that is structurally valid and complies with the submission requirements of the presentation definition, it MUST respond with an HTTP status code 200, indicating that the request was successfully processed in terms of protocol compliance. If the submitted credential is invalid (e.g., expired, revoked, or does not meet the verifier’s functional requirements), the Response Endpoint SHOULD respond with an HTTP status code 400 (Bad Request). The 400 response MUST include an `error` parameter with a value such as `invalid_credential` or `functional_requirements_not_met`, and MAY include a `redirect_uri` to allow the wallet to resubmit with different credentials. The `error_description` parameter MAY provide additional details to assist the wallet in correcting the submission.
+
 `response_uri`:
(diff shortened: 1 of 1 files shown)


jogu (changes requested):
@subhra-io We don't seem to have a contribution agreement from you - can you (or your company if applicable) complete the docusign form at https://openid.net/intellectual-property/openid-foundation-contribution-agreements/ please?

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1201,6 +1201,8 @@ The Response Mode is defined in accordance with [@!OAuth.Responses] as follows:
 
 The following new Authorization Request parameter is defined to be used in conjunction with Response Mode `direct_post`: 
 
+If the Response Endpoint receives a request in the direct_post response mode that is structurally valid and complies with the submission requirements of the presentation definition, it MUST respond with an HTTP status code 200, indicating that the request was successfully processed in terms of protocol compliance. If the submitted credential is invalid (e.g., expired, revoked, or does not meet the verifier’s functional requirements), the Response Endpoint SHOULD respond with an HTTP status code 400 (Bad Request). The 400 response MUST include an `error` parameter with a value such as `invalid_credential` or `functional_requirements_not_met`, and MAY include a `redirect_uri` to allow the wallet to resubmit with different credentials. The `error_description` parameter MAY provide additional details to assist the wallet in correcting the submission.
```
Sakurann, line 1204: there is no more submission requirements, since there is only DCQL. 

also this text should be in section 8.5 https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-8.5

Issue #457, opened 2025-03-14
Clarify transaction_data_hashes input and output encoding in SD-JWT


sander:
Context: https://github.com/openid/OpenID4VP/pull/421#discussion_r1993894939

Proposal:

> Array of **base64url-encoded** hashes, where each hash is calculated using a hash function over the data in the **base64url-decoded** strings received in the `transaction_data` request parameter.

ubamrein:
I would highly discourage using decoded data as an argument for the hash function, if there is not a very solid argument for it. Basically applying any transformation to data (e.g. base64decoding) will potentially change the underlaying data which then ultimately will be hashed (to a different hash). 

Especially in the case when transaction data is a JSON string, one might be tempted to decode it directly to a string and parse it (maybe trimming because, you know the parser fails otherwise), which then obviously leads to problems.

Also I think it needs immediate attention, as the comment here https://github.com/openid/OpenID4VP/pull/421#discussion_r1969356942 was in the end ignored, and the very unfortunate phrasing (which is open to interpretations) was merged

> Array of hashes, where each hash is calculated using a hash function over the data in the strings received in the transaction_data request parameter

@Sakurann we are already aware of different implementations e.g. niscy uses the base64ENCODED data whereas the AT wallet team uses the base64DECODED data.

bc-pi:
> I would highly discourage using decoded data as an argument for the hash function

I would agree with this

tlenz:
> I would highly discourage using decoded data as an argument for the hash function, if there is not a very solid argument for it. Basically applying any transformation to data (e.g. base64decoding) will potentially change the underlaying data which then ultimately will be hashed (to a different hash).

We do not see a problem with that because the Base64Url transformation is only needed to transport arbitrary data, which must not have a String representation on its own as a `transaction_data` element. Therefore, calculating the hash directly over the arbitrary data requires a to decode it. 

We fully agree that we need clarification because the current version is still confusing and has room for interpretation that breaks interoperability. Regarding the clarification, we support @sander original proposal to:
> Array of base64url-encoded hashes, where each hash is calculated using a hash function over the data in the base64url-decoded strings received in the transaction_data request parameter.

because it is fully clear and allows arbitrary data to be `transaction_data` elements. However, there is no strong opinion on that.

ubamrein:
[...]

PR #615, opened 2025-05-27
Issue 457 transaction data hashes

subhra-io:
This PR resolves issue #457 by updating the `transaction_data_hashes` definition in Appendix B.3.1.0 (SD-JWT Profile) to specify that each hash is a base64url-encoded SHA-256 hash of the UTF-8 JSON data obtained by base64url-decoding each `transaction_data` element, per @sander’s proposal. A non-normative example demonstrates the decoding, hashing, and encoding steps. The text addresses @ubamrein’s concerns by requiring valid UTF-8 decoding to prevent errors, ensuring interoperability.


jogu:
I've commented on the issue and believe there is a need for a clarification, I've opened https://github.com/openid/OpenID4VP/pull/621 to make the minimal clarification I think is needed.

@subhra-io We don't seem to have a contribution agreement from you - can you (or your company if applicable) complete the docusign form at https://openid.net/intellectual-property/openid-foundation-contribution-agreements/ please? I think adding an example would still be good but if you're going to do that we would need the contribution agreement to be signed first.

Sakurann (changes requested):
there is absolutely no WG agreement to hash base64url decoded data

PR #619, opened 2025-05-28
clarify text about how encryption keys are obtained

### LINKED ISSUE #616: clarify how encryption key is obtained
> To obtain the Verifier's public key to which to encrypt the Authorization Response, the Wallet uses keys from client metadata, such as the jwks member within the client_metadata request parameter, the metadata defined in the Entity Configuration if OpenID Federation is used, or other mechanisms. 

WG discussion. clarify the text that encryption key must be obtained from jwks in the client metadata. that client metadata can be obtained from client_metadata or entity identifier, etc.

https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-8.3-3

fix in the same direction in section 5.10

https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#section-5.10-5

bc-pi:
hopefully fixes #616

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -618,7 +618,7 @@ The following parameters are defined to be included in the request to the Reques
 `wallet_nonce`:
 : OPTIONAL. A String value used to mitigate replay attacks of the Authorization Request. When received, the Verifier MUST use it as the `wallet_nonce` value in the signed authorization request object. Value can be a base64url-encoded, fresh, cryptographically random number with sufficient entropy.  
 
-If the Wallet requires the Verifier to encrypt the Request Object, it uses the `jwks` parameter within the `wallet_metadata` parameter to pass public encryption keys. If the Wallet requires an encrypted Authorization Response, it SHOULD specify supported encryption algorithms using the `authorization_encryption_alg_values_supported` and `authorization_encryption_enc_values_supported` parameters.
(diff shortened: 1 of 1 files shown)


c2bo (approved):

awoie (commented):
Generally looks fine but what is the reason the SHOULDs were removed? Is the intention to make it effectively a MUST which is ok but just wanted to clarify.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -618,7 +618,7 @@ The following parameters are defined to be included in the request to the Reques
...
 
-If the Wallet requires the Verifier to encrypt the Request Object, it SHOULD use the `jwks` or `jwks_uri` parameter within the `wallet_metadata` parameter to pass the public key for the input to the key agreement. Other mechanisms to pass the encryption key can be used as well. If the Wallet requires an encrypted Authorization Response, it SHOULD specify supported encryption algorithms using the `authorization_encryption_alg_values_supported` and `authorization_encryption_enc_values_supported` parameters. 
+If the Wallet requires the Verifier to encrypt the Request Object, it uses the `jwks` parameter within the `wallet_metadata` parameter to pass public encryption keys. If the Wallet requires an encrypted Authorization Response, it specifies supported encryption algorithms using the `authorization_encryption_alg_values_supported` and `authorization_encryption_enc_values_supported` parameters.
```
Sakurann, line 621:
```suggestion
If the Wallet requires the Verifier to encrypt the Request Object, it SHOULD use the `jwks` parameter within the `wallet_metadata` parameter to pass public encryption keys. If the Wallet requires an encrypted Authorization Response, it SHOULD specify supported encryption algorithms using the `authorization_encryption_alg_values_supported` and `authorization_encryption_enc_values_supported` parameters.
```

PR #626, opened 2025-06-02
Update non-final specs

### LINKED ISSUE #530: Update non-final note and re-check references
After other PRs for 1.0 are merged, check that the section is up-to-date

Currently aware of the fully specified algorithms one

c2bo:
Closes #530

Added Fully-specified Algorithms and updated references.

I went through the normative references and we currently have 2 other specs not mentioned here:
- SD-JWT which I think is fine since it is implied by sd-jwt vc and is getting published soon?
- DC API which is intentional as far as I understand and the section is written in a way that we do not normatively depend on the concrete constructions inside that reference?

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1742,6 +1742,7 @@ Implementers should be aware that this specification uses several specifications
 
 * OpenID Federation 1.0 draft -42 [@!OpenID.Federation]
 * SIOPv2 draft -13 [@!SIOPv2]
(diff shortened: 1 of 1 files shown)


tplooker (approved):

c2bo:
> * for SD-JWT, once its published, we would need to change how we reference that spec in opeind4vp, right? to [@!RFCxxxx]
> * for DC API, then we need to move this to informative section [W3C.Digital_Credentials_API]
>   Caceres, M., Cappalli, T., and S. Goto, "Digital Credentials API", https://w3c-fedid.github.io/digital-credentials/.

Yeah I was wondering about that part for DC API as well - whether that should be a normative reference or informal

Sakurann:
@jogu why keep it in the normative section?

Sakurann (approved):

PR #628, opened 2025-06-04
Defined APV values for ECDH-ES

GarethCOliver:
Solves topic from #597

This defines the value for ECDH-ES, as the base64url encoding of the sha-256 hash of concatted values. 

The values to be concatted are defined based on mode per the request, with reasoning AIUI. 

TODO to add two illustrative examples. examples (and update the above examples.

### DIFF
--- a/openid-4-verifiable-presentations-1_0.md
+++ b/openid-4-verifiable-presentations-1_0.md
@@ -1410,6 +1410,22 @@ into the key derivation process that is used to derive the content encryption ke
 
 Note: For encryption, implementers have a variety of options available through JOSE, including the use of Hybrid Public Key Encryption (HPKE) as detailed in [@I-D.ietf-jose-hpke-encrypt]. 
 
+### ECDH-ES APU/APV Values
+When performing Reponse Encryption where the `alg` is ECDH-ES the `apv` MUST be set by the Wallet and validated by the Verifier. The value to set is is the base64url encodig of the sha-256 hash of the bytes of several fields concatanated. The fields to be concatanated is based on the response mode:
+
+  - When response mode is `direct_post.jwt` the clientId, nonce, jwkThumbprint and responseUri
+  - When response mode is `dc_api.jwt` the origin, nonce and jwkThumbprint
+
+(TODO: Add two examples of the apv values, one for `direct_post.jwt` and another for `dc_api.jwt`).
+
+The fields must be concatonated in the order specified here before being hashed, and base64url encoded. To validate the `apv` the Verifier MUST construct the same value and compare it to the `apv` included in the protected headed. If the `apv` value fails validation the Verifier MUST stop processing and discard the response. 
+
+These values are the same as those required to perform verification of the credential presentations. This binds the encryption to the presentation and allows the verifier to terminate processing early when the presentation is not meant for them.
+
+Note that because the `apv` is attached, this does not allow the encryption to fail closed (i.e. be unable to decrypt when the presentation is not meant for them), but the same approach could be taken when such an algorithm is supported.
+
+There are no requirements on the `apu` value so the Wallet may use any value (including leaving it empty).
+
 ### Response Mode "direct_post.jwt" {#direct_post_jwt}
 
 This specification also defines a new Response Mode `direct_post.jwt`, which allows for encryption to be used on top of the Response Mode `direct_post` defined in (#response_mode_post). The mechanisms described in (#response_mode_post) apply unless specified otherwise in this section.


openid-4-verifiable-presentations-1_0.md
```diff
@@ -1410,6 +1410,22 @@ into the key derivation process that is used to derive the content encryption ke
 
 Note: For encryption, implementers have a variety of options available through JOSE, including the use of Hybrid Public Key Encryption (HPKE) as detailed in [@I-D.ietf-jose-hpke-encrypt]. 
 
+### ECDH-ES APU/APV Values
+When performing Reponse Encryption where the `alg` is ECDH-ES the `apv` MUST be set by the Wallet and validated by the Verifier. The value to set is is the base64url encodig of the sha-256 hash of the bytes of several fields concatanated. The fields to be concatanated is based on the response mode:
+
+  - When response mode is `direct_post.jwt` the clientId, nonce, jwkThumbprint and responseUri
+  - When response mode is `dc_api.jwt` the origin, nonce and jwkThumbprint
```
selfissued, line 1417: Per https://www.rfc-editor.org/rfc/rfc7638.html#section-3, the JWK Thumbprint is the hash of the bytes of a canonical representation of the JWK.  When SHA-256 is the hash function, this is 32 bytes of binary data.  Is that what you intend to include in the concatenation?  Or do you intend to include the base64url-encoding of the JWK Thumbprint in the concatenation?

selfissued (changes requested):
When this was proposed on the last call, I had asked that a clear description of the problem being solved with this proposed addition be provided in the PR description.  Specifically, what is the attack that is possible without this addition and how does this addition prevent it?  How realizable is the attack is in practice and what are its consequences?  What parties must perform what malicious actions for the attack to succeed, and what are the preconditions for them to be able to do so?

This information isn't provided in the PR or in the predecessor PR or issue.

Given we're 6 weeks into the [60-day review](https://openid.net/public-review-period-for-proposed-openid-verifiable-presentations-final-specification/) for the specification to become final, I'm quite skeptical of making this breaking and complicating change without a clear and compelling description of why we MUST do this now.  That description should include an analysis of why including these specific parameters in the computation will prevent the attack and why this is the right set of parameters to do so.

openid-4-verifiable-presentations-1_0.md
```diff
@@ -1410,6 +1410,22 @@ into the key derivation process that is used to derive the content encryption ke
...
+
+(TODO: Add two examples of the apv values, one for `direct_post.jwt` and another for `dc_api.jwt`).
+
+The fields must be concatonated in the order specified here before being hashed, and base64url encoded. To validate the `apv` the Verifier MUST construct the same value and compare it to the `apv` included in the protected headed. If the `apv` value fails validation the Verifier MUST stop processing and discard the response.
```
c2bo, line 1421:
```suggestion
The fields MUST be concatenated in the order specified here before being hashed, and base64url encoded. To validate the `apv` the Verifier MUST construct the same value and compare it to the `apv` included in the protected header. If the `apv` value fails validation the Verifier MUST stop processing and discard the response. 
```

openid-4-verifiable-presentations-1_0.md
c2bo, line 1417: I think something like this would help?
```suggestion
  - When response mode is `direct_post.jwt` the `client_id`, `nonce`, `jwkThumbprint` and `responseUri`
  - When response mode is `dc_api.jwt` the `origin`, `nonce` and `jwkThumbprint`
  
  where 
  
  - `client_id` is the `client_id` request parameter
  - `nonce` is the `nonce` request parameter
  - `jwkThumbprint` is the base64url encoded JWK SHA-256 Thumbprint as defined in [@!RFC7638]
  - `responseUri` is either the `redirect_uri` or the `response_uri` request parameter,
  - `origin` is the string representing the Origin of the request as described in (#dc_api). MUST NOT be prefixed with `origin:`
```

martijnharing: And then add "of the Verifier's public key used to encrypt the response." for the jwkThumbprint?

Sakurann:
WG discussion about defining apu/apv values for ECDH-ES JWE..
- on one hand...
    - adds marginal security benefit (majority agrees with @selfissued and @bc-pi disagreeing)
    - would not be the most complex part of the OpenID4VP implementation and can be implemented using existing JOSE libraries
    - there is implementation experience of similar mechanisms in 18013-7 annex B 
- on the other hand...
    - there are strong concerns that adding this mechanism last minute would result in it being underspecified leading to interoperability problems
    - extension mechanism to define apu/apv values can be added in 1.1 (verifiers that support only that extension will potentially reject responses from wallets that do not support it)
    - wg seems to agree to define the content of a detached payload for JOSE-HPKE in OpenID4VP 1.1 or HAIP

wg agreed to discuss and tackle #347 during wglc/review period and as an outcome this PR and PR #597 will be closed for the above reasons since there is no wg rough consensus to proceed with either of them.

PR #630, opened 2025-06-09
Add missing entries to document history

jogu:
Went through commits & check everything has a history entry (veering more towards the "list absolutely everything that changed even if it isn't normative" side of things given the spec is already under public review).

Reordered them slightly so the most normative looking ones are first in the list.

Includes history entry for privacy considerations PR for speed, even though it's not merged yet.


selfissued (approved):

Sakurann (approved):
looks great - thank you!

PR #633, opened 2025-06-10
Rough shell script partly automating publishing to openid.net/specs

jogu:
This makes it easier to prepare the files correctly.

### DIFF
--- a/publish.sh
+++ b/publish.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+# build the OID4VP spec and copy it into publication repo
+#
+# assumes you've done:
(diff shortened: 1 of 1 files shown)


c2bo:
If that helps, we could add a github action for this that runs on something like a tag or a github release? That way you could just trigger a release in the github repo and get the prepared spec as an artifact.

jogu:
> If that helps, we could add a github action for this that runs on something like a tag or a github release? That way you could just trigger a release in the github repo and get the prepared spec as an artifact.

I think this is where we want to end up... the action could I think actually create the PR in the publication repo. Any help would be very welcome!

PR #636, opened 2025-06-13
Fix three typos in Draft-29.

dastoikov:
An attempt to fix three typos in openid-4-verifiable-presentations-1_0.md. No systematic reading of the spec was performed, so there may be more of this kind lurking.


dastoikov:
> @dastoikov do you have OIDF IPR signed? we would need it signed to be able to accept the PR

@Sakurann I have just signed it.

PR #638, opened 2025-06-16
Clarify that QR code include both client_id and request_uri

jogu:
RFC9101 requires client_id to be in the authorization endpoint parameters so fix editorial text that implies only request_uri is included.

(and add Michael Z to the acknowledgements as he reported this and some
other things previously, and also add Andres Olave)


sloops77 (approved):

sloops77:
@jogu while you are at it, pls add me to the helpers list :)

jogu:
@sloops77 just to check, have you signed a contribution agreement? If not could you do so please, you can do it very easily on docusign here: https://openid.net/intellectual-property/openid-foundation-contribution-agreements/

c2bo (approved):

PR #651, opened 2025-07-03
Various editorial nits

jogu:


dastoikov:
@jogu :

> I'm not sure if I'm fully following what you meant, but I found and fixed an incorrect "fulfilled" so thank you and hopefully that was what you meant :)

Yes, I meant the one that you fixed. Apologies for not being clear.

> (I added you to the acknowledgments too)

Thank you!

> If there's something else that still needs fixing please shout!

I will review the document this evening (UTC+3), so if you are in a hurry, please merge. Most likely, there will not be anything else that needs fixing.

dastoikov (approved):

dastoikov:
@jogu :

Attached is a second patch, this time consisting solely of typo fixes (along with two instances of adding a comma before the last item in a list).

For example, replaced "intend" with "intent" in "issued by a trusted authority, to prove that the Verifier has publicly registered its intent to request certain credentials.".
[0001-Typo-fixes.patch.zip](https://github.com/user-attachments/files/21108109/0001-Typo-fixes.patch.zip)

I reviewed the document in full.

jogu:
Thanks @dastoikov - I've applied most of both patches - I didn't take a couple of changes:

- I've retained "interacting with the end-user" rather than simplifying to "interaction" as I think the extra words aid the reader in understanding
- I didn't take the ASCII URL safe change as it felt a little too close to a normative change (I think it's probably okay but I'd rather veer on the side of caution at this late stage - we could definitely put it into a separate PR and consider it for an errata)

PR #657, opened 2025-07-10
Add 1.0 to document title

jogu:
Many (but not all) OIDF specs have version numbers in the title; given we are likely to have 1.1 etc versions of OID4VP it seems good to give the version number a bit more prominence.


jogu:
> We should also update the title of OpenID4VCI, etc. to add 1.0 to the titles in the references.

I've made that change.

PR #663, opened 2025-08-06
Consistently use given_name and family_name in examples

### LINKED ISSUE #662: Use given_name instead of first_name in Appendix D mdoc examples
The DCQL examples in appendix D use either `first_name` or `given_name` seemingly without much consistency. While I would find it more satisfying to always use `given_name`, that might be a matter of personal preference. However, when referencing specified mdoc namespaces such as `org.iso.18013.5.1`, I do think we need to use the defined values (`given_name` and not `first_name`). See example below for the incorrect use imo.

```json
{
  "credentials": [
    {
      "id": "my_credential",
      "format": "mso_mdoc",
      "meta": {
        "doctype_value": "org.iso.7367.1.mVRC"
      },
      "claims": [
        {"path": ["org.iso.7367.1", "vehicle_holder"]},
        {"path": ["org.iso.18013.5.1", "first_name"]}
      ]
    }
  ]
}
```

I'll create a quick PR

javereec:
closes #662 

I did opt to change all attribute names as I noticed that SD-JWT ones were also not consistent.

Noticed that the main text is also not consistent (but left that as is)


Sakurann (approved):

Sakurann:
editors/chairs need to create 1.1.md file

Issue #646, opened 2025-06-25
Handling of `origin` in Native Mobile Platform App to App flows for DC API


tplooker:
When using OpenID4VP to accomplish an App to App flow (e.g where the verification application is a native mobile application rather than a web application), the `origin` used in the DC API flow takes a different form to when using the W3C DC API on the web. 

In Appendix A currently all the examples are focused around using the W3C DC API via the web, so the `origin` is always a web origin, e.g `origin1.example.com`.

However, when performing an App to App flow, the origin is mobile platform specific.

For example on Android using the [Android Credential Manager](https://developer.android.com/jetpack/androidx/releases/credentials), the origin that wallets are using, such as [CMWallet](https://github.com/digitalcredentialsdev/CMWallet/tree/main) follows the following syntax

```
android:apk-key-hash-sha256:<base64_encoded_sha256_hash-of-apk-signing-cert>
```

See [here](https://github.com/digitalcredentialsdev/CMWallet/blob/ebab30b5ed329ac4ebc7d682fc8d5d906b8e9696/app/src/main/java/com/credman/cmwallet/mdoc/MDoc.kt#L20-L28) for a reference.

Which in turn looks to be based on the [FIDO standard](https://fidoalliance.org/specs/common-specs/fido-appid-and-facets-v2.1-ps-20220523.html) which is used for solving a similar problem for FIDO credentials.

Currently in VP we have text that notes the origin is platform specific in these scenarios, however we could improve guidance here either pointing externally or adding some text describing how to handle the origin in these cases.

javereec:
Is this still relevant? In the latest draft "Origin" already includes clarification on this

> Origin:
> : An identifier for the calling website or native application, asserted by the web or app platform. A web origin is the combination of a scheme/protocol, host, and port, with port being omitted when it matches the default port of the scheme. An app platform may use a linked web origin, or use a platform-specific URI for the app origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The web origin is `https://verify.example.com` with `https` being the scheme, `verify.example.com` being the host, and the port is not explicitly included as `443` is the default port for the protocol `https`. The native applications origin on some platforms will also be `https://verify.example.com` and on other platforms, may be `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`.

brentzundel:
[...]

PR #719, opened 2026-04-16
Refine processing expected_origins and processing steps

### LINKED ISSUE #224: Clarifications on processing expected_origins and processing steps
The spec says:

> expected_origins: REQUIRED when signed requests defined in [Appendix A.3.2](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#signed_request) are used with the W3C Digital Credentials API [[w3c.digital_credentials_api](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-wg-draft.html#w3c.digital_credentials_api)]. An array of strings, each string representing an origin of the Verifier that is making the request. The Wallet can detect replay of the request from a malicious Verifier by comparing values in this parameter to the origin asserted by the Web Platform.

However, the spec seems to be missing details of how `expected_origins` is to be processed.

What should happen when:

1. **`expected_origins` is empty**
   - **Issue:** It's unclear whether the implementation should accept any origin or reject all requests when `expected_origins` is empty.
   - **Example Case:** If `expected_origins` is an empty array, should the implementation treat this as allowing any origin or blocking all origins?
   - WG meeting 11.12.25: final text: empty array must be rejected

2. **`expected_origins` contains an invalid origin or URLs**
   - **Issue:** Should the implementation reject the entire list or simply ignore invalid entries if it contains invalid origins?
   - **Example Case:** If `expected_origins` contains entries like `["1://2", "https://foo.com"]`, should the presence of `1://2` cause the whole list to be rejected, or should it just ignore `1://2` and drop it on the floor?
   -   WG meeting 11.12.25:  Note: expected origins could be web origins or other kind of origins, such as Android package names. The wallet is supposed to treat it as a string, not as a URL. 
   - Question: How do we ensure that the string provided by the platform to the wallet is formatted the same way on all platform?

3. **`expected_origins` contains a non-https origin**
   - **Issue:** It is unclear whether non-https origins should be allowed.
   - **Example Case:** If `expected_origins` includes `["gopher://foo:123"]`, should this entry be accepted or rejected?
   - WG meeting 11.12.25:  does not matter

4. **`expected_origins` contains a relative or absolute path, or an empty string**
   - **Issue:** Should paths be resolved against a base URL or be considered invalid? How should empty strings be handled?
   - **Example Case:** If `expected_origins` includes relative or absolute paths like `[".", "/"]` or an empty string like `[""]`, should these be resolved against a base URL or rejected?
   -  WG meeting 11.12.25:  does not matter, as the origin is not being resolved

5. **`expected_origins` contains URLs with paths, query, and/or fragments**
   - **Issue:** It is unclear whether origins with paths, query parameters, or fragments should be normalized (e.g., removing paths, queries, and fragments) or considered invalid.
   - **Example Case:** If `expected_origins` contains URLs such as `["https://foo.com/path?query=1#fragment"]`, should these be normalized to just the origin or be rejected?
   -   WG meeting 11.12.25: does not matter

6. **`expected_origins` contains duplicates**
   - **Issue:** Should duplicates be removed or cause the entire list to be rejected?
   - **Example Case:** If `expected_origins` includes `["https://foo.com", "https://foo.com"]`, should the duplicates be removed or should this result in an error?
   - WG meeting 11.12.25: just checking the match, no need to detect duplicates

7. **`expected_origins` contains non-string entries**
   - **Issue:** Should all entries in `expected_origins` be converted to strings? By implication, `expected_origins` should be defined as a WebIDL `USVString` to ensure proper handling of Unicode and type conversions.
   - **Example Case:** If `expected_origins` includes entries like `["https://foo.com", null, "https://bar.com"]`, should non-string entries be converted to strings or cause an error?
   - WG meeting 11.12.25: would not match, non string would not match definition, must be rejected
   
   **Explanation:** Defining `expected_origins` as a WebIDL `USVString` ensures that all entries are properly handled as strings, accommodating Unicode characters correctly. This is important because URLs can contain Unicode characters, and using `USVString` helps to normalize these characters to their Unicode Scalar Value (USV) form, ensuring consistency and preventing issues related to encoding and type mismatches.

8. **Handling of Localhost and Intranet URLs**
   - **Issue:** Should URLs pointing to `localhost` or intranet addresses be treated with special consideration, especially regarding security implications?
   - **Example Case:** If `expected_origins` contains `["https://localhost"]` or `["https://192.168.1.1"]`, should these be allowed given they are local resources?

9. **Normalization of URLs**
   - **Issue:** During normalization, the URLs should be stripped of their query parameters, fragments, and paths. The primary concern is ensuring that only the origin part (scheme, host, port) is used.
   - **Example Case:** If `expected_origins` includes `["https://foo.com/path?query=1#fragment"]`, should it be normalized to `["https://foo.com"]` before comparison?

   **Recommendation:** Each URL should be parsed using the WHATWG URL parser, and only the origin should be added to a set to avoid duplicates. This ensures that the origins are consistently and correctly normalized, enhancing security and preventing ambiguity.

To ensure proper handling of `expected_origins`, we recommend defining it as follows in WebIDL:

```webidl
dictionary OpenID4VPRequest {
  sequence<USVString> expected_origins;
};
```

Further, the spec is missing validation step. Here's a start:

### Validator and Processor for `OpenID4VPRequest`

This section defines an algorithm to validate and process the `expected_origins` array in an `OpenID4VPRequest` dictionary. The output will be a set of normalized origins, safely discarding query, fragments, paths, etc.

#### Algorithm

1. **Input:**
    - An `OpenID4VPRequest` dictionary |request|, containing:
      - `expected_origins`: A sequence of `USVString`.

2. **Output:**
    - A set of normalized origins.

3. **Steps:**
  
    1. **Check if `expected_origins` is empty:**
        1. If `expected_origins` is empty, throw a `TypeError` indicating that `expected_origins` must not be empty.

    3. Let `origins_set` be an empty set.

    4. For each `origin` in `expected_origins`:
        1. **Parse the URL:**
            1. Let `url` be the result of parsing `origin` using the [URL parser](https://url.spec.whatwg.org/#concept-url-parser) with `null` as the URL base.
            2. If parsing `origin` fails, throw a `TypeError` indicating an invalid URL.
            
        2. **Check scheme:**
            1. If `url`'s scheme is not "https", throw a `TypeError` indicating an unsupported URL scheme.
            2. **ISSUE:** Should `localhost` and intranet addresses be allowed? (e.g., `https://localhost` or `https://192.168.1.1`). Probably not. 
        3. **Extract origin:**
            1. Let `normalized_origin` be the result of running the [origin](https://html.spec.whatwg.org/multipage/origin.html#ascii-serialisation-of-an-origin) serialization steps on `url`.

        5. **Check for duplicates:**
            1. If `origins_set` contains `normalized_origin`, continue.

        6. **Add to set:**
            1. Add `normalized_origin` to `origins_set`.

    4. **Check if `origins_set` is empty:**
        1. If `origins_set` is empty, throw a `TypeError` indicating that no valid origins were found.

    5. **Return the set:**
        1. Return `origins_set`.

awoie:
Fixes #224


1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2439,7 +2439,7 @@ The value of the `response_mode` parameter MUST be `dc_api` when the response is
...
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as an opaque string, using simple string comparison, without any prior processing of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
awoie:
```suggestion
* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string without interpretation, using simple string comparison, without any prior processing of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3567,6 +3567,8 @@ The technology described in this specification was made available from contribut
 -31
 
    * Clarify that `encrypted_response_enc_values_supported` applies only if JWE content encryption algorithm is used
+   * Clarify provided Origin is an opaque string and validation is based on simple string comparison
```
awoie:
```suggestion
   * Clarify provided Origin is a string without prior interpretation and validation is based on simple string comparison
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2439,7 +2439,7 @@ The value of the `response_mode` parameter MUST be `dc_api` when the response is
...
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as an opaque string, using simple string comparison, without any prior processing of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
awoie:
```suggestion
* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string without interpretation, using simple string comparison, without any prior processing of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3569,3 +3569,5 @@ The technology described in this specification was made available from contribut
 
    * Add security consideration not to use VP Token as Access Token
    * Clarify that `encrypted_response_enc_values_supported` applies only if JWE content encryption algorithm is used; e.g., it does not apply to JOSE HPKE
+   * Clarify provided Origin is an opaque string and validation is based on simple string comparison
```
awoie:
```suggestion
   * Clarify provided Origin is a string without prior interpretation and validation is based on simple string comparison
```

awoie:
I removed the term "opaque" as per WG consensus.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2439,7 +2439,7 @@ The value of the `response_mode` parameter MUST be `dc_api` when the response is
...
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string without interpretation, using simple string comparison, without any prior processing of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
awoie:
```suggestion
* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string, using simple string comparison, without any prior processing or interpretation of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2439,7 +2439,7 @@ The value of the `response_mode` parameter MUST be `dc_api` when the response is
...
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string without interpretation, using simple string comparison, without any prior processing of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
awoie:
```suggestion
* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string, using simple string comparison, without any prior processing or interpretation of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2388,7 +2388,7 @@ Secondly, the session with the End-User will always continue in the initial cont
 
 Thirdly, cross-device requests benefit from the use of secure transports with proximity checks, which are handled by the OS platform, e.g., using FIDO CTAP 2.2 with hybrid transports.
 
-And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the user agent, which is important for phishing resistance.
+And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the trusted platform (e.g., user agent), which is important for phishing resistance.
```
fkj: Are there other trusted platforms which are not an user agent?
```suggestion
And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the trusted platform (i.e. user agent), which is important for phishing resistance.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2388,7 +2388,7 @@ Secondly, the session with the End-User will always continue in the initial cont
 
 Thirdly, cross-device requests benefit from the use of secure transports with proximity checks, which are handled by the OS platform, e.g., using FIDO CTAP 2.2 with hybrid transports.
 
-And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the user agent, which is important for phishing resistance.
+And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the trusted platform (e.g., user agent), which is important for phishing resistance.
```
fkj: Same comment as above regarding other trusted platforms than user agents.

fkj (commented):
I think the clarity could be improved by also adding text to note that an Origin can really be any string in the definition of Origin.

Sakurann:
APAC DCP WG discussion:
- to resolve #224, 
  - current text looks good to clarify origin is a string and must not be parsed
  - suggestion is to add to this PR clarification of origin definition for non-native app RPs. @fkj will make a suggestion
- keep discussion of native app origin value out of scope of this PR and continue that discussion in #646

1.0/openid-4-verifiable-presentations-1_0.md
fkj:
```suggestion
Origin:
: An identifier for the calling website or native application, asserted by the web or app platform. The Origin MAY be any string, but will typically follow a platform-specific convention. A web origin is the combination of a scheme/protocol, host, and port, with port being omitted when it matches the default port of the scheme. An app platform may use a linked web origin, or use a platform-specific URI for the app origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The web origin is `https://verify.example.com` with `https` being the scheme, `verify.example.com` being the host, and the port is not explicitly included as `443` is the default port for the protocol `https`. The native applications origin on some platforms will also be `https://verify.example.com` and on other platforms, may be `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`.
```

awoie: @fkj Just editorially, I would like to avoid having normative verbs in the terminology section for some reasons (e.g. defining test cases). Does the following work?

```suggestion
Origin:
: An identifier for the calling website or native application, asserted by the web or app platform. The Origin is any string, but will typically follow a platform-specific convention. A web origin is the combination of a scheme/protocol, host, and port, with port being omitted when it matches the default port of the scheme. An app platform may use a linked web origin, or use a platform-specific URI for the app origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The web origin is `https://verify.example.com` with `https` being the scheme, `verify.example.com` being the host, and the port is not explicitly included as `443` is the default port for the protocol `https`. The native applications origin on some platforms will also be `https://verify.example.com` and on other platforms, may be `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`.
```

1.1/openid-4-verifiable-presentations-1_1.md
awoie: see above

1.0/openid-4-verifiable-presentations-1_0.md
awoie: @fkj Technically different from DC API, but we are using the same method for third-party-app verifier app sending a request via the app platform to the wallet. In this case, the trusted platform is the app platform or something in between the two apps. That is why I used "e.g.". Not sure if user agent applies in this case. Does this make sense?

1.1/openid-4-verifiable-presentations-1_1.md
awoie:
```suggestion
And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the trusted platform (i.e., user agent), which is important for phishing resistance.
```

1.0/openid-4-verifiable-presentations-1_0.md
fkj: Yeah, I have no strong opinion on the exact wording here.

1.0/openid-4-verifiable-presentations-1_0.md
fkj: I'm not sure whether that qualifies as an user agent either to be honest.

GarethCOliver (approved):

paulbastian (approved):

deshmukhrajvardhan (approved):

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -121,7 +121,8 @@ Issuer-Holder-Verifier Model:
...
+Origin:
+: An identifier for the calling website or native application, asserted by the web or app platform. The Origin is any string, but will typically follow a platform-specific convention. A web origin is the combination of a scheme/protocol, host, and port, with port being omitted when it matches the default port of the scheme. An app platform may use a linked web origin, or use a platform-specific URI for the app origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The web origin is `https://verify.example.com` with `https` being the scheme, `verify.example.com` being the host, and the port is not explicitly included as `443` is the default port for the protocol `https`. The native applications origin on some platforms will also be `https://verify.example.com` and on other platforms, may be `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`.
```
lj-raidiam: We use the terms "app origin" and "web origin" here, where "app origin" can be a URI. Could we reference the relevant RFCs for clarity? WDYT @awoie @fkj?
If this suggestion is added, RFC 6454 should be added to the references at the bottom.

```suggestion
: An identifier for the calling website or native application, asserted by the web or app platform. The Origin is any string, but will typically follow a platform-specific convention. A web origin follows scheme, host, and port convention defined in [@!RFC6454], with port being omitted when it matches the default port of the scheme. An app platform may use a linked web origin, or use a platform-specific URI syntax as defined by [@!RFC3986] for the app origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The web origin is `https://verify.example.com` with `https` being the scheme, `verify.example.com` being the host, and the port is not explicitly included as `443` is the default port for the protocol `https`. The native applications origin on some platforms will also be `https://verify.example.com` and on other platforms, may be `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`.
```

fkj: I don't see why not, it might add some clarity.

c2bo (approved):

1.0/openid-4-verifiable-presentations-1_0.md
marcoscaceres: I don’t think that’s the right spec. Browser and most modern URL implementations use WHATWG URL. 

Similarly, origin you probably want :
https://html.spec.whatwg.org/multipage/browsers.html#concept-origin

Generally, please  avoid redefining what’s in other specs. Subtle wording differences can lead to confusion.

marcoscaceres: I think this is all you want here, but check: 

```suggestion
: an [origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin) or an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque) [[HTML]]. See [7.1.1 Origins](https://html.spec.whatwg.org/multipage/browsers.html#origin) of HTML for details. 
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2388,7 +2389,7 @@ Secondly, the session with the End-User will always continue in the initial cont
 
 Thirdly, cross-device requests benefit from the use of secure transports with proximity checks, which are handled by the OS platform, e.g., using FIDO CTAP 2.2 with hybrid transports.
 
-And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the user agent, which is important for phishing resistance.
+And lastly, as part of the request, the Wallet is provided with information about the Verifier's Origin as authenticated by the trusted platform (i.e. user agent), which is important for phishing resistance.
```
marcoscaceres, line 2424: Origin should link to definition

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2439,7 +2440,7 @@ The value of the `response_mode` parameter MUST be `dc_api` when the response is
...
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string, using simple string comparison, without any prior processing or interpretation of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
marcoscaceres: As above.

1.0/openid-4-verifiable-presentations-1_0.md
fkj: Good point. I think we were not previously aware that WHATWG actually defined the opaque origin as a concept. I still think it might be relevant to say that the opaque origin is platform-specific.

c2bo: We should still preserve text with examples about platform-provided origin (app-to-app) imho. Even if we re-use other, existing terminology. I believe we should be more explicit here, especially with an example what a platform-provided origin could look like

marcoscaceres: Yes, a non-normative example would be great. If there’s a common one that’s already in use, all the better. However, it’s important to think about any that we know absolutely must not be allowed (e.g., file:, http:)… they don’t need to be defined here, but should be somewhere. 

something like: 
> If the url's [scheme](https://url.spec.whatwg.org/#concept-url-scheme) is a [local scheme](https://fetch.spec.whatwg.org/#local-scheme), or http:, or file:, or javascript:, or ws:, or wss:, return error.

brentzundel:
Discussed during call. In general Marco's proposed change seems fine, but it should have correct references and retain some example text.

1.0/openid-4-verifiable-presentations-1_0.md
c2bo: for what it's worth, OpenID specs usually don't like to Terminology like the w3c specs do

brentzundel:
@marcoscaceres could you respond to comments made so far?

1.0/openid-4-verifiable-presentations-1_0.md
awoie: OIDF specs don't link to terminology.

1.0/openid-4-verifiable-presentations-1_0.md
awoie: This OIDF spec doesn't link to terminology.

1.0/openid-4-verifiable-presentations-1_0.md
awoie: Since `expected_origins` is part of the OID4VP protocol payload, OID4VP can define constraints on the values a verifier is allowed to provide there. The platform-provided origin itself is outside the scope of OID4VP and is determined by the W3C Digital Credentials API / platform layer. OID4VP can therefore only specify how the wallet compares `expected_origins` with the origin provided by the platform, and how the wallet behaves when no match is found, or the value of `expected_origins` uses an unsafe / unsupported URI scheme.

We could say something like this (wdyt @marcoscaceres @fkj @c2bo ?) :
> Values using unsafe or unsupported URI schemes, including `ftp`, `javascript`, `data`, `ws`, and `wss`, MUST NOT be used. Values using the `http` scheme MUST NOT be used unless they are explicitly allowed for constrained scenarios such as local development or equivalent non-production environments.

awoie: How about the following which combines @marcoscaceres reference while including native applications as well as retaining the examples?

> Origin: An identifier for the calling website or native application, asserted by the underlying web or application platform.
> For Web-based callers, the Origin is an opaque origin or tuple origin as defined in HTML Section 7.1.1 "Origins"[HTML].
> For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin.
> For example, the Verifier for the organization MyExampleOrg is served from `https://verify.example.com`. The corresponding web Origin is `https://verify.example.com`, where `https` is the scheme, `verify.example.com` is the host, and the port is not explicitly included because `443` is the default port for `https`. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as `platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0`.
> The Origin provided by the underlying platform is outside the scope of this specification.

We could then also add a constraint on `expected_origins` as follows:
> Values using unsafe or unsupported URI schemes, including `ftp`, `javascript`, `data`, `ws`, and `wss`, MUST NOT be used. Values using the `http` scheme MUST NOT be used unless they are explicitly allowed for constrained scenarios such as local development or equivalent non-production environments.

awoie: Wdyt about the above proposal @fkj @marcoscaceres @c2bo @lj-raidiam ?

awoie: See proposal below.

c2bo: That sounds good to me!

fkj: I think we should add that the platform-specific application origins MUST be treated as opaque origins (cf. the HTML spec).

c2bo: > The Origin provided by the underlying platform is outside the scope of this specification.

I guess we could modify the last sentence along those lines?

> The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string.

paulbastian:
wg call:
- suggestion from @c2bo and @fkj to further iterate on @awoie latest proposal, taking reference from Marcos suggesting and merging them with ideas from @awoie 
- @c2bo is syncing with @awoie to make a new proposal, as it seems to complicated in a suggestion

c2bo:
I added the changes discussed in comments with the reference to the html origin terminology.

I wasn't sure about the link to whatwg - in IETF we had to pin to a specific commit instead of the living standard , so I did the same here.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2389,6 +2389,16 @@ Ecosystems intending to use trusted authority mechanisms SHOULD ensure that the
   </front>
 </reference>
 
+<reference anchor="whatwg.html" target="https://html.spec.whatwg.org/commit-snapshots/c560a56704e7a31887ab79299d30e6f68d696cf4/#origin">
```
marcoscaceres: Yeah, definitely don't do that. We (WHATWG Members) made them living for a reason.

marcoscaceres: Also, I think we want the origin concept, not the section: 
```suggestion
<reference anchor="whatwg.html" target="https://html.spec.whatwg.org#concept-origin">
```

c2bo: Yeah, that is why i wasn't sure / commented on it -a lot of SDOs don't allow living standards and require some kind of pinned version

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2461,7 +2471,7 @@ The value of the `response_mode` parameter MUST be `dc_api` when the response is
...
+* `expected_origins`: REQUIRED when signed requests defined in (#signed_request) are used with the Digital Credentials API (DC API). A non-empty array of strings, each string representing an Origin of the Verifier that is making the request. The Wallet MUST compare values in this parameter to the provided Origin, treated as a string, using simple string comparison, without any prior processing or interpretation of either the `expected_origins` values or the provided Origin, to detect replay of the request from a malicious Verifier. Values using unsafe or unsupported URI schemes, including `ftp`, `javascript`, `data`, `ws`, and `wss`, MUST NOT be used. Values using the http scheme MUST NOT be used unless they are explicitly allowed for constrained scenarios such as local development or equivalent non-production environments. If the Origin does not match any of the entries in `expected_origins`, the Wallet MUST return an error. This error SHOULD be an `invalid_request` error. This parameter is not for use in unsigned requests and therefore a Wallet MUST ignore this parameter if it is present in an unsigned request.
```
marcoscaceres, line 2475: It's not clear what "Values using unsafe or unsupported URI schemes" are, and the current list leaves it open ended. Would it be worthwhile to just restrict to `http` and `https` for now? 

Similarly, the "Values using the http scheme MUST NOT be used unless they are explicitly allowed for constrained scenarios such as local development or equivalent non-production environments." 

That's ambiguous. I think you want to go through the "[is potentially trustworthy?](https://w3c.github.io/webappsec-secure-contexts/#is-url-trustworthy)" checks of the Secure Context spec. That covers the "local development or equivalent non-production environments" case unambiguously and safely.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -117,7 +117,7 @@ Issuer-Holder-Verifier Model:
...
+: An identifier for the calling website or native application, asserted by the underlying web or application platform. For Web-based callers, the Origin is an opaque `origin` or `tuple origin` as defined in Section 7.1.1 "Origins" of [@!whatwg.html]. For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The corresponding web Origin is https://verify.example.com, where https is the scheme, verify.example.com is the host, and the port is not explicitly included because 443 is the default port for https. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0. The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string.
```
marcoscaceres: Note: it's safer not to cite sections in HTML. Stuff tends to move around, but the concepts remain fairly stable.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2453,6 +2453,16 @@ Ecosystems intending to use trusted authority mechanisms SHOULD ensure that the
   </front>
 </reference>
 
+<reference anchor="whatwg.html" target="https://html.spec.whatwg.org/commit-snapshots/c560a56704e7a31887ab79299d30e6f68d696cf4/#origin">
```
marcoscaceres: as above...

1.0/openid-4-verifiable-presentations-1_0.md
c2bo: I looked at a few more examples to reference whatwg - i think something like this would be a decent compromise? https://datatracker.ietf.org/doc/html/rfc10008#section-6.2 

> WHATWG, "FETCH", WHATWG Living Standard, <https://fetch.spec.whatwg.org/>. Commit snapshot: <https://fetch.spec.whatwg.org/commit-snapshots/3bab31a55154bda73f25b45a23df718616f2f64e/>.

1.0/openid-4-verifiable-presentations-1_0.md
c2bo: Since we also allow want to allow calls from other apps with an expected_origins value , we shouldn't restrict to http/https imho.

Sakurann: WG discussion. given that there are use-cases for app-to-app, preference is to not limit it to http/https only.

we could quote https://www.w3.org/TR/secure-contexts/#is-url-trustworthy for the cases when values are http/https. @marcoscaceres is Secure context document published? it says `[W3C Candidate Recommendation Draft](https://www.w3.org/standards/types#CRD), 10 November 2023` which means it's not a TR but URL says TR? if it's CR draft, then "These documents MUST NOT be cited as W3C standards and may or may not become W3C standards." ...

1.1/openid-4-verifiable-presentations-1_1.md
Sakurann:
```suggestion
: An identifier for the calling website or native application, asserted by the underlying web or application platform. For Web-based callers, the Origin is an opaque `origin` or `tuple origin` as defined in the "Origins" section of [@!whatwg.html]. For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The corresponding web Origin is https://verify.example.com, where https is the scheme, verify.example.com is the host, and the port is not explicitly included because 443 is the default port for https. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0. The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string.
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -121,7 +121,7 @@ Issuer-Holder-Verifier Model:
...
+: An identifier for the calling website or native application, asserted by the underlying web or application platform. For Web-based callers, the Origin is an opaque `origin` or `tuple origin` as defined in Section 7.1.1 "Origins" of [@!whatwg.html]. For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The corresponding web Origin is https://verify.example.com, where https is the scheme, verify.example.com is the host, and the port is not explicitly included because 443 is the default port for https. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0. The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string.
```
Sakurann:
```suggestion
: An identifier for the calling website or native application, asserted by the underlying web or application platform. For Web-based callers, the Origin is an opaque `origin` or `tuple origin` as defined in the "Origins" section of [@!whatwg.html]. For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The corresponding web Origin is https://verify.example.com, where https is the scheme, verify.example.com is the host, and the port is not explicitly included because 443 is the default port for https. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0. The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string.
```

Sakurann:
WG discussion: ready to merge once the last comment from marcos is resolved

c2bo:
@marcoscaceres does the current state of the PR work for you? Given that we don't want to deny the app-based origins, it feels like this is a decent state?

1.0/openid-4-verifiable-presentations-1_0.md
marcoscaceres: CR(Draft) are stable documents. They are Candidate Recommendations, which is the last step before Recommendation and for all intents and purposes considered stable.

marcoscaceres:
@c2bo as long as app origins are actually origins and can be treated purely and strictly as origins (scheme, host, port) then it should be fine. But note that paths and other parts will be discarded. At the same time, some origins should be disallowed, like file://, sms://, mailto:, etc.

brentzundel:
from what I can determine, this PR is ready to merge. Requested changes and comments have been addressed, there has been sufficient review time, there are no outstanding requests for changes, and there are multiple positive approvals

PR #726, opened 2026-04-26
Normative version inconsistency for SD-JWT VC in 1.0 final

### LINKED ISSUE #659: Normative version inconsistency for SD-JWT VC in 1.0 final
Whilst working on the implementation, one of my colleagues noted an inconsistence In the normative references section of the spec for SD-JWT VC.

It points to [draft 10](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-16) in the normative references, but then in section [13.4](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-13.4) it points to version 9. To make matters worse, that last section mentions to always use the version listed in that section in case the respective draft would be updated and breaking changes would be introduced. 

So that section basically says to use draft 9, whilst the formal normative references section mentions draft 10. Section 13.4 is a bit vague anyway. It tries to convey something that it should only be the case if breaking changes are introduced. But it is not a burden to put on an implementor of this spec IMO, and simply leaves room for interpretation and thus interop problems.

Happy to make a PR, but then would need to know which one prevails. Assuming 1.0 Final is out, I guess it would become draft 9 actually and 10 for future versions of this spec. However 13.4 does not really make clear these are normative statements, so you could also say that the only normative version is 10

Sakurann:
fixes #659 by changing typo in sd-jwt-vc version

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -1762,7 +1762,7 @@ Implementers should be aware that this specification uses several specifications
 * OpenID Federation 1.0 draft -43 [@!OpenID.Federation]
 * SIOPv2 draft -13 [@!SIOPv2]
 * Selective Disclosure for JWTs (SD-JWT) draft -22 [@!I-D.ietf-oauth-selective-disclosure-jwt]
-* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -09 [@!I-D.ietf-oauth-sd-jwt-vc]
+* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -10 [@!I-D.ietf-oauth-sd-jwt-vc]
 * Fully-Specified Algorithms for JOSE and COSE draft -13 [@!I-D.ietf-jose-fully-specified-algorithms]
 
 While breaking changes to the specifications referenced in this specification are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced versions above in preference to the final versions, unless updated by a profile or new version of this specification.
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -1762,7 +1762,7 @@ Implementers should be aware that this specification uses several specifications
 * OpenID Federation 1.0 draft -43 [@!OpenID.Federation]
 * SIOPv2 draft -13 [@!SIOPv2]
 * Selective Disclosure for JWTs (SD-JWT) draft -22 [@!I-D.ietf-oauth-selective-disclosure-jwt]
-* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -10 [@!I-D.ietf-oauth-sd-jwt-vc]
+* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -16 [@!I-D.ietf-oauth-sd-jwt-vc]
 * Fully-Specified Algorithms for JOSE and COSE draft -13 [@!I-D.ietf-jose-fully-specified-algorithms]
 
 While breaking changes to the specifications referenced in this specification are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced versions above in preference to the final versions, unless updated by a profile or new version of this specification.
@@ -3211,7 +3211,7 @@ The following outlines a suggested non-normative set of processing steps for SD-
 
 ##### Step 1: SD-JWT VC Processing
 
-- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 4 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
+- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
 - If the `vct` value is associated with any SD-JWT VC Type Metadata, schema validation of the entire SD-JWT VCLD is performed, including the nested `ld` claim.
 - Additionally, trust framework rules are applied, such as ensuring the Credential Issuer is authorized to issue SD-JWT VCLDs for the specified `vct` value.
 
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -1822,7 +1822,7 @@ Implementers should be aware that this specification uses several specifications
 * OpenID Federation 1.0 draft -43 [@!OpenID.Federation]
 * SIOPv2 draft -13 [@!SIOPv2]
 * Selective Disclosure for JWTs (SD-JWT) draft -22 [@!I-D.ietf-oauth-selective-disclosure-jwt]
-* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -09 [@!I-D.ietf-oauth-sd-jwt-vc]
+* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -10 [@!I-D.ietf-oauth-sd-jwt-vc]
 * Fully-Specified Algorithms for JOSE and COSE draft -13 [@!I-D.ietf-jose-fully-specified-algorithms]
 
 While breaking changes to the specifications referenced in this specification are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced versions above in preference to the final versions, unless updated by a profile or new version of this specification.
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -1822,7 +1822,7 @@ Implementers should be aware that this specification uses several specifications
 * OpenID Federation 1.0 draft -43 [@!OpenID.Federation]
 * SIOPv2 draft -13 [@!SIOPv2]
 * Selective Disclosure for JWTs (SD-JWT) draft -22 [@!I-D.ietf-oauth-selective-disclosure-jwt]
-* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -10 [@!I-D.ietf-oauth-sd-jwt-vc]
+* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -16 [@!I-D.ietf-oauth-sd-jwt-vc]
 * Fully-Specified Algorithms for JOSE and COSE draft -13 [@!I-D.ietf-jose-fully-specified-algorithms]
 
 While breaking changes to the specifications referenced in this specification are not expected, should they occur, OpenID4VP implementations should continue to use the specifically referenced versions above in preference to the final versions, unless updated by a profile or new version of this specification.
@@ -3276,7 +3276,7 @@ The following outlines a suggested non-normative set of processing steps for SD-
 
 ##### Step 1: SD-JWT VC Processing
 
-- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 4 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
+- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
 - If the `vct` value is associated with any SD-JWT VC Type Metadata, schema validation of the entire SD-JWT VCLD is performed, including the nested `ld` claim.
 - Additionally, trust framework rules are applied, such as ensuring the Credential Issuer is authorized to issue SD-JWT VCLDs for the specified `vct` value.
 


paulbastian (approved):

c2bo:
We need to touch that part anyway for 1.1 I believe (and check versions / remove some entries if there were no breaking changes)? Should this change be only scoped for 1.0 for the time being and we do a pass of the non-final specs part for 1.1 before final?

brentzundel:
Discussed in call. We may need to update to 16, but only after making sure there aren't normative inconsistencies.
There is also an SD-JWT reference that need avaluation.

jogu (changes requested):
There's still an inconsistency as the normative references section now references draft 16. We need to update the normative references to be draft 10, unless there's a later draft that has no normative changes in which change we can update both places to that later version.

brentzundel:
@Sakurann can you respond to the concerns raised by @jogu

Sakurann:
Does @jogu 's concern only applies to the following sentence and section 2 vs 4? I don't understand the concern. is it still there if now the PR points to -16 and not -10?

> A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.

Sakurann:
> We need to touch that part anyway for 1.1 I believe (and check versions / remove some entries if there were no breaking changes)? Should this change be only scoped for 1.0 for the time being and we do a pass of the non-final specs part for 1.1 before final?

@c2bo do you see any problems with merging this PR now and we can revisit 1.1 before publication again

c2bo:
> > We need to touch that part anyway for 1.1 I believe (and check versions / remove some entries if there were no breaking changes)? Should this change be only scoped for 1.0 for the time being and we do a pass of the non-final specs part for 1.1 before final?
> 
> @c2bo do you see any problems with merging this PR now and we can revisit 1.1 before publication again

fine by me and I believe Brent opened a new issue for that

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3211,7 +3211,7 @@ The following outlines a suggested non-normative set of processing steps for SD-
 
 ##### Step 1: SD-JWT VC Processing
 
-- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 4 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
+- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
```
awoie, line 3214:
```suggestion
- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2.4 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3276,7 +3276,7 @@ The following outlines a suggested non-normative set of processing steps for SD-
 
 ##### Step 1: SD-JWT VC Processing
 
-- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 4 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
+- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
```
awoie, line 3279:
```suggestion
- A receiver (holder or verifier) of an SD-JWT VCLD applies the processing rules outlined in Section 2.4 of [@!I-D.ietf-oauth-sd-jwt-vc], including verifying signatures, validity periods, status information, etc.
```

awoie (changes requested):
thanks. we should use 2.4 instead of 2.

paulbastian:
@javereec @fkj  volunteering to review this PR

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1762,7 +1762,7 @@ Implementers should be aware that this specification uses several specifications
 * OpenID Federation 1.0 draft -43 [@!OpenID.Federation]
 * SIOPv2 draft -13 [@!SIOPv2]
 * Selective Disclosure for JWTs (SD-JWT) draft -22 [@!I-D.ietf-oauth-selective-disclosure-jwt]
-* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -09 [@!I-D.ietf-oauth-sd-jwt-vc]
+* SD-JWT-based Verifiable Credentials (SD-JWT VC) draft -16 [@!I-D.ietf-oauth-sd-jwt-vc]
```
jogu, line 1765: Draft 16 contains at least one breaking change ( https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-16.html#name-document-history ):

> Change lang to locale. While lang is more accurate, locale is what has traditionally been used in OpenID Connect and later related specs.


so I don't think we can just update to -16, unless people can think of language we can add that could explain this as not really being a breaking change.

There may be other breaking changes too, I didn't review the full history.

The alternative is to figure out how to make the normative references section of the draft refer to draft 10 (or the latest version that doesn't contain breaking changes). I don't know how we get mmarkdown/xml2rfc to reference a specific IETF draft rather than the latest one - probably we would need to create a manual reference instead of relying on the automatic `[@!I-D.ietf-oauth-sd-jwt-vc]` syntax.

jogu (changes requested):

1.0/openid-4-verifiable-presentations-1_0.md
fkj: I agree that section 2.4 is the better reference. Section 2.4 also references the validation rules in section 2.5, so scoping it down like this makes it easier to find the rules without being too specific.

javereec:
Below are some notable differences between draft 10 and draft 16 of the SD-JWT VC spec that could potentially present an issue.

| description | draft 10 | draft 16 |
|-|-|-|
|`vct#integrity` claim | Not part of "Registered JWT Claims" | Part of "Registered JWT Claims". Because if it, this must always be disclosed. |
| `status` claim constraint | No constraint on content format | Addition of "the associated Status List Token MUST be in JWT format" in section 2.2.2.2 |
| Binary data encoding in claims | No requirement | Added requirement that: "Binary data in claims SHOULD be encoded as data URIs as defined in [[RFC2397](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-16.html#RFC2397)]"  in section 2.2.2.3 |
| Display data language tag property rename | Uses `lang` in section 8 | Uses `locale` in section 4.5 |
| SVG rendering method key change | Uses `svg_template` in section 8.1.2 | Uses `svg_templates` in section 4.5.1.2 |
| `vct` type metadata required | Not mentioned in type metadata in section 6.2 | Mentioned and required in Type metadata in section 4.2 |
| `schema` and `schema_uri` removed from type metadata | Mentioned in type metadata in section 6.2, but optional | Removed entirely from type metadata in section 4.2. |
| Type metadata glue documents (`vctm`) removed | Present | Removed |
| Introduction of `mandatory` claim in claim metadata | Not present | New `mandatory` boolean property. The claim itself is optional |
| Verification step when schema is present in type metadata removed | Contains "If a schema is provided in the Type Metadata, a recipient MUST validate the schema as defined in [Section 6.5](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-10.html#schema-type-metadata)." in section 3.4 | Verification step is removed in section 2.4 |

awoie:
> thanks. we should use 2.4 instead of 2.

I will also note that ETSI references -13. Because SD-JWT VC is probably most relevant for EUDI, and EUDI implementers will need to use -13, we should probably also use -13, or update to -16/-17 (next week) already. Between -13 and -16/-17 there shouldn't be any breaking changes.

brentzundel:
After discussion in the WG today, we don't like it, but we should point to the latest version of SD-JWT VC and move forward.
Mid to long term there should probably be a registry.

jogu:
> After discussion in the WG today, we don't like it, but we should point to the latest version of SD-JWT VC and move forward.

Relevant minutes for references: https://lists.openid.net/pipermail/openid-specs-digital-credentials-protocols/Week-of-Mon-20260713/001320.html


The WG can't just decide "we're going to make breaking changes in an errata".

As I mentioned here https://github.com/openid/OpenID4VP/pull/726#discussion_r3386792753 we really need to explain and justify why we're making a breaking change in an errata update (including making clear what behaviour we expect implementations with/without the errata update to have), and some of the explanation is likely to be guidance to implementers/ecosystems that needs to go into the specification text.

> Mid to long term there should probably be a registry.

Can you expand on this point please? A registry of what? [Actually I think that question is answered by the minutes - I don't think a registry solves any of this problem - I believe it's a registry of credential format identifiers like dc+sd-jwt etc - but the registry still has to stay 'dc+sd-jwt' means "SD-JWT VC draft 18" and can't change the meaning of that value ever.)

brentzundel:
Discussion in the group today: perhaps a breaking change in this circumstance might be necessary, and allowable under the need to address errata. If we go down this route, we need to be very explicit about the changes.
There seemed to be general support for this approach.
@Sakurann can you update this PR?

Sakurann:
WG discussion (chair hat on)

multiple options:
- option 1: use the same sd-jwt format identifier but update the version -10 to -16/final
  -  **add a text saying that if the wallet does not know the version of sd-jwt vc, return it.** many wallets do not know the version of the sd-jwt vc they hold. recommend final ver but not mandate.
- option 2: define format identifier per sd-jwt vc version (potentially a registry)
- option 3: define request parameter that allows verifier to specify sd-jwt version. if version is `any`, de facto same as option 1.

-> preferred direction seems to be option 1. 
  - for implementers who used VCI issuer metadata as opposed to sd-jwt vc metadata, same openid4vp code can be used. an email will be sent out asking implementers if anyone have strong objections updating from -10 to -16 (preferrably final ver of sd-jwt vc in vp 1.1 and haip 1.1 final)

what does conformance mean? need to provide clarity:
- strictly speaking, for vp 1.0, only returning sd-jwt vc -10 is compliant
- for 1.1, returning any sd-jwt vc version? returning an error is ok, but not non-compliant

brentzundel:
Discussed today, @Sakurann please adjust this PR according to option 1. Make sure to mention wallet and verifier requirements.

brentzundel:
Discussed today, still waiting for @Sakurann to update the PR

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3085,7 +3085,10 @@ If `require_cryptographic_holder_binding` is set to `false`, an SD-JWT without t
 
 ### Format Identifier
 
-The Credential Format Identifier is `dc+sd-jwt`.
+The Credential Format Identifier is `dc+sd-jwt`. 
+
+The `dc+sd-jwt` Credential Format Identifier identifies an IETF SD-JWT VC independently
+of the version of [@!I-D.ietf-oauth-sd-jwt-vc] used to issue it.
```
brentzundel: Adding this line confuses me. I don't think doing so is necessary

```suggestion
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3152,6 +3152,9 @@ If `require_cryptographic_holder_binding` is set to `false`, an SD-JWT without t
 
 The Credential Format Identifier is `dc+sd-jwt`.
 
+The `dc+sd-jwt` Credential Format Identifier identifies an IETF SD-JWT VC independently
+of the version of [@!I-D.ietf-oauth-sd-jwt-vc] used to issue it.
+
```
brentzundel: As above 
```suggestion
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3178,6 +3181,19 @@ The following requirements apply to the `nonce` and `aud` claims in the Key Bind
...
+A Wallet's inability to determine the version of [@!I-D.ietf-oauth-sd-jwt-vc]
+that the Issuer used to issue a Credential of the `dc+sd-jwt` Credential Format
+MUST NOT, on that basis alone, cause the Wallet to treat the Credential as
+non-matching or to withhold it from a response it would otherwise return to the
+Verifier. All other conditions for returning a Credential, including satisfying
+the DCQL query and obtaining the required End-User authorization apply.
+
+A Verifier MUST be prepared to receive a Credential of the `dc+sd-jwt` Credential
+Format that was issued under a version of [@!I-D.ietf-oauth-sd-jwt-vc] other than
+the one it expects. A Verifier MAY respond with an error if it cannot process the
```
jogu: This feels like it is probably too strong to me - for a given credential / set of trusted issuers there is probably a limited set of versions that actually exist, so I don't think there's a strong reason to require the verifier to support other versions.

1.0/openid-4-verifiable-presentations-1_0.md
jogu: I think saying something here in the format identifier section along the lines of "dc+sd-jwt is used to describe all versions of oauth-sd-jwt-vc" is a worthwhile statement. Possibly we might want to limit that to just the version VP 1.0 final references unto the latest.

jogu (commented):
I don't think the "we should be explicit that this is a potential breaking change compared to the first version of 1.0" part at https://github.com/openid/OpenID4VP/pull/726#issuecomment-5282797723 has been captured in the spec yet. Maybe something like 'Verifiers or Wallets compliant with older versions may not accept/correctly process credentials compliant with newer versions of SD-JWT VC." (that's not a very good wording but hopefully the intent is kind of clear).

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3178,6 +3181,19 @@ The following requirements apply to the `nonce` and `aud` claims in the Key Bind
...
+received version. Doing so is a local processing decision and does not render the
+Wallet's response non-conformant.
```
c2bo: Something like this to make it a note?
```suggestion
Note that a Verifier should be prepared to receive a Credential of the `dc+sd-jwt` Credential
Format that was issued under a version of [@!I-D.ietf-oauth-sd-jwt-vc] other than
the one it expects.
A Verifier can respond with an error if it cannot process the received version. Doing so is a local processing decision and does not render the Wallet's response non-conformant.
```

brentzundel:
Discussed in call. The group felt it would be better to avoid normative language and instead have guidance for the verifier: "You should ask for SD-JWT VC 1.0, but you may not get it, up to you if that's acceptable or not."
@Sakurann please adjust PR

PR #728, opened 2026-04-26
Intro text about communicating credential request mentions only dcql_query even though scope can be used too

### LINKED ISSUE #674: Intro text about communicating credential request mentions only dcql_query even though scope can be used too
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#name-authorization-request says:

> The Verifier articulates requirements of the Credential(s) that are requested using the dcql_query parameter.

However https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.5 allows the use of scopes as well.

So something like this would be better:

> The Verifier articulates requirements of the Credential(s) that are requested using either the dcql_query or scope parameters.

Also it seems like there's some missing words between 'DCQL Query" and 'Section 5" here:

> A VP Token is only returned if the corresponding Authorization Request contained a dcql_query parameter or a scope parameter representing a DCQL Query [Section 5](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#vp_token_request).

maybe:

> A VP Token is only returned if the corresponding Authorization Request contained a dcql_query parameter or a scope parameter representing a DCQL Query (as defined in [Section 5](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#vp_token_request)).

Sakurann:
editorial 
resolves #674

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -276,7 +276,7 @@ that signals to the Wallet that it can make an HTTP POST request to the Verifier
 endpoint with information about its capabilities as defined in (#request_uri_method_post). The Wallet MAY continue with JAR
 when it receives `request_uri_method` parameter with the value `post` but does not support this feature.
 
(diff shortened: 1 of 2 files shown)


paulbastian (approved):
I would actually argue to remove scope as it's yet another option that people barely use to my knowledge, but I guess that's not the right place to have this discussion.

fkj (approved):

jogu (approved):

Issue #642, opened 2025-06-17
Should we mandate the `use` field for keys in the JSON Web Key Set in client_metadata


QZHelen:
As an observation during an interop testing, the latest spec does not explicitly require use for each JSON Web Key from the request [client_metadata.jwks](https://openid.net/specs/openid-4-verifiable-presentations-1_0-29.html#section-5.1-5.2.2.1); therefore, test verifiers may not be setting this property their encryption public key. 

Given that the spec text suggests that the key may not just be used for encryption purpose, not having `use` could cause ambiguity when there are two keys of the same type, used for different purposes. The holder could pick a wrong key that is supposed to be used for different purpose to perform the response encryption.

It would be nice to clarify the expectation in the OpenID4VP spec to eliminate such ambiguity. Is it always true that the jwks keys shall always be used for encryption? Otherwise, can we require key `use` at least for the encryption key, or have some solution on the line to make sure that the holder will always know what key to pick for response encryption?

Sakurann:
the use of `use` parameter is not mandatory. PR #725. is further clarification needed?

Sakurann:
> Is it always true that the jwks keys shall always be used for encryption? 

yes.

in errata, to avoid breaking changes, we can only say that if the issuer is passing keys with multiple purposes in the client-metadata, it should clarify `use`. otherwise, keys are intended to be used for encryption

PR #730, opened 2026-04-26
Clarified intent_to_retain value when not present

### LINKED ISSUE #669: Clarify intent_to_retain
As discussed in WG10 meeting in Arizona, currently `intent_to_retain` is OPTIONAL but we don't say what value the wallet should use in this case. Suggest that we add "If not specified, the wallet shall use the value false".

Sakurann:
resolves #669 and potentially #642 - need to discuss if we want to add additional text to clarify usage of `use` is not mandatory

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -2808,7 +2808,7 @@ be a valid doctype identifier as defined in [@ISO.18013-5].
 The following are ISO mdoc specific parameters to be used in a Claims Query as defined in (#claims_query).
 
 `intent_to_retain`
-: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5].
+: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5]. If not specified, the Wallet uses the value `false`.
 
 ### Presentation Response
 
@@ -3567,7 +3567,8 @@ The technology described in this specification was made available from contribut
 -31
 
    * Clarify that `encrypted_response_enc_values_supported` applies only if JWE content encryption algorithm is used
-   * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used 
+   * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
+   * Clarified intent_to_retain value when not present
    
 -final
    
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -3593,7 +3593,6 @@ The technology described in this specification was made available from contribut
    * Clarify nonce entropy requirements
    * Clarify that state is recommended to match text from Section 14.3.2. Protection of the Response URI
    * Clarify that `encrypted_response_enc_values_supported` applies only if JWE content encryption algorithm is used
-   * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
    * Clarified intent_to_retain value when not present
    * Removed invalid_scope error guidance as duplicate of RFC6749
    * Clarified that Multi-RP-sig section means Verifier Info instead of attestations
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -2832,7 +2832,7 @@ be a valid doctype identifier as defined in [@ISO.18013-5].
 The following are ISO mdoc specific parameters to be used in a Claims Query as defined in (#claims_query).
 
 `intent_to_retain`
-: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5]. If not specified, the Wallet uses the value `false`.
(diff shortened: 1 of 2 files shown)


selfissued (changes requested):
We should not be creating a default interpretation of not specifying this parameter in a way that requires changes to existing wallets.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2808,7 +2808,7 @@ be a valid doctype identifier as defined in [@ISO.18013-5].
 The following are ISO mdoc specific parameters to be used in a Claims Query as defined in (#claims_query).
 
 `intent_to_retain`
-: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5].
+: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5]. If not specified, the Wallet uses the value `false`.
```
jogu: I can't find any place where we actually reached a conclusion on what we should do.

https://github.com/openid/OpenID4VP/issues/669#issuecomment-3281503177 says:

> explicitly stating that intent_to_retain being optional means that the verifier is not specifying if it will retain or not, which is arguably what the current spec says.

jogu (changes requested):
> resolves https://github.com/openid/OpenID4VP/issues/669 and potentially https://github.com/openid/OpenID4VP/issues/642 -

I don't think this touches #642.

brentzundel:
discussed in call today.  Some suggested changes to the text were made.

brentzundel:
Seemed to be consensus that if the verifier is not expressing anything, the wallet should indicate that the verifier hasn't expressed anything.

1.0/openid-4-verifiable-presentations-1_0.md
```suggestion
: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5]. If absent, the Verifier makes no statement on `IntentToRetain`.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2873,7 +2873,7 @@ be a valid doctype identifier as defined in [@ISO.18013-5].
 The following are ISO mdoc specific parameters to be used in a Claims Query as defined in (#claims_query).
 
 `intent_to_retain`
-: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5].
+: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5]. If not specified, the Wallet uses the value `false`.
```
```suggestion
: OPTIONAL. A boolean that is equivalent to `IntentToRetain` variable defined in Section 8.3.2.1.2.1 of [@ISO.18013-5]. If absent, the Verifier makes no statement on `IntentToRetain`.
```

selfissued (changes requested):
Please delete the statements "If not specified, the Wallet uses the value `false`." and replace them with statements that if not specified, the caller is making statement about whether it will retain the claims or not.

1.1/openid-4-verifiable-presentations-1_1.md
brentzundel: This suggested change matches my understanding of where the WG ended up during the last call.

@selfissued @jogu would this change address your requests for changes?

selfissued: Yes, please apply Christian's proposed changes.  Then I will approve.

brentzundel: @Sakurann looking for your input here as well, since this is your PR

c2bo (approved):

brentzundel (approved):
@jogu and @selfissued please make sure this addresses your concerns

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3661,6 +3661,7 @@ The technology described in this specification was made available from contribut
    * Clarify that state is recommended to match text from Section 14.3.2. Protection of the Response URI
    * Clarify that `encrypted_response_enc_values_supported` applies only if JWE content encryption algorithm is used; e.g., it does not apply to JOSE HPKE
    * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
+   * Clarified intent_to_retain value when not present
```
awoie:
```suggestion
   * Clarified `intent_to_retain` value when not present
```

awoie (approved):

paulbastian (approved):

fkj (approved):

paulbastian:
@jogu and @selfissued please review the latest changes on this PR. Otherwise we have sufficient approvals

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3593,6 +3593,8 @@ The technology described in this specification was made available from contribut
    * Clarify nonce entropy requirements
    * Clarify that state is recommended to match text from Section 14.3.2. Protection of the Response URI
    * Clarify that `encrypted_response_enc_values_supported` applies only if JWE content encryption algorithm is used
+   * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
```
jogu: Don't know why it was decided to add this here as it's not relevant to this issue - it's already present on line 3601 so shouldn't be added a second time.

```suggestion
```

jogu (changes requested):

c2bo:
@jogu, @selfissued could you please re-review/resolve the request for changes if everything is good now?

selfissued (approved):

jogu (approved):

PR #733, opened 2026-04-26
request_uri_method should be case sensitive

### LINKED ISSUE #675: request_uri_method should be case sensitive
https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.1 currently says:

> request_uri_method:
OPTIONAL. A string determining the HTTP method to be used when the request_uri parameter is included in the same request. Two case-sensitive valid values are defined in this specification

I'm dubious that it makes sense to have a parameter where some values can be case sensitive and some case insensitive. I think we should just define it as a case sensitive string.

Sakurann:
resolves #675


1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -304,7 +307,7 @@ This specification defines the following new request parameters:
...
+: OPTIONAL. A ase-sensitive string determining the HTTP method to be used when the `request_uri` parameter is included in the same request. Two case-sensitive valid values are defined in this specification: `get` and `post`. If `request_uri_method` value is `get`, the Wallet MUST send the request to retrieve the Request Object using the HTTP GET method, i.e., as defined in [@RFC9101]. If `request_uri_method` value is `post`, a supporting Wallet MUST send the request using the HTTP POST method as detailed in (#request_uri_method_post). If the `request_uri_method` parameter is not present, the Wallet MUST process the `request_uri` parameter as defined in [@RFC9101]. Wallets not supporting the `post` method will send a GET request to the Request URI (default behavior as defined in [@RFC9101]). `request_uri_method` parameter MUST NOT be present if a `request_uri` parameter is not present.
```
fkj:
```suggestion
: OPTIONAL. A case-sensitive string determining the HTTP method to be used when the `request_uri` parameter is included in the same request. Two case-sensitive valid values are defined in this specification: `get` and `post`. If `request_uri_method` value is `get`, the Wallet MUST send the request to retrieve the Request Object using the HTTP GET method, i.e., as defined in [@RFC9101]. If `request_uri_method` value is `post`, a supporting Wallet MUST send the request using the HTTP POST method as detailed in (#request_uri_method_post). If the `request_uri_method` parameter is not present, the Wallet MUST process the `request_uri` parameter as defined in [@RFC9101]. Wallets not supporting the `post` method will send a GET request to the Request URI (default behavior as defined in [@RFC9101]). `request_uri_method` parameter MUST NOT be present if a `request_uri` parameter is not present.
```

fkj (changes requested):

c2bo (approved):

jogu (approved):

fkj (approved):

PR #744, opened 2026-06-18
Add additional text clarifying how to match vct and doctype

### LINKED ISSUE #741: Clarification on checking inheritance of received `vct` (VP token validation)
Currently, Section 8.6 (VP Token Validation) does not explicitly state whether the Verifier is required to validate that the `vct` of a returned SD-JWT VC matches one of the `vct_values` requested in the DCQL query. This could become an issue if `vct` inheritance is used, e.g., if a root certificate is used to issue document signer certificates for different `vct` values.

Section B.3.5 defines `vct_values` as follows:

> `vct_values`: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [SD-JWT VC]. The Wallet MAY return Credentials that inherit from any of the specified types, following the inheritance logic defined in [SD-JWT VC].

Section 8.6 contains the following validation requirements:

> Check the individual Presentations according to the specific Credential Format requested:
>
> Validate that the returned Credential(s) meet all criteria defined in the query in the Authorization Request (e.g., Claims included in the presentation).
> Check that the set of Presentations returned satisfies all requirements defined in the Verifier's request as described in [Section 6.4](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#dcql_query_lang_processing_rules).
> If any of the checks related to an individual Presentation fail, the effected Presentation MUST be discarded. If any of the checks pertaining to the VP Token or the overall response fails, the VP Token MUST be rejected.

While the generic requirement to validate that returned credentials meet all query criteria could be interpreted as covering `vct_values`, there is currently no explicit statement that the Verifier must validate that the Credential's `vct` either:

1. matches one of the requested `vct_values`, or
2. is a subtype thereof according to the SD-JWT VC inheritance rules.

Given that `vct_values` is the primary mechanism for requesting specific SD-JWT VC types, it may be beneficial to make this validation step explicit in Section 8.6, rather than relying on the reader to infer it from the more generic query-validation requirements. @jogu @c2bo @fkj @paulbastian

GarethCOliver:
Resolves #741 by adding more explicit instructions on what allows a Credential to satisfy `vct_values`, and applies similar explicit text to  `doctype_value`

This makes use of the SD-JWT VC draft 15 which added `aka_vcts` for this purpose

### DIFF
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -2889,8 +2889,7 @@ The following is an ISO mdoc specific parameter in the `meta` parameter in a Cre
 
 `doctype_value`:
 : REQUIRED. String that specifies an allowed value for the
-doctype of the requested Verifiable Credential. It MUST
-be a valid doctype identifier as defined in [@ISO.18013-5].
+doctype of the requested Verifiable Credential. It MUST be a valid doctype identifier as defined in [@ISO.18013-5]. For a mdoc to satisfy the Credential Query, the value of the `docType` element in the `MobileSecurityObject` structure MUST be equal to the `doctype_value`.
 
 ### Parameter in the Claims Query {#mdocs_claims_query}
 
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter 
 The following is an SD-JWT VC specific parameter in the `meta` parameter in a Credential Query as defined in (#credential_query).
 
 `vct_values`:
-: REQUIRED. A non-empty array of strings that specifies allowed values for
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+A Credential's type can be determined as satifying the Credential Query as follows:
+
+1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
+1. If is is not, check if the Credential has an `extends` claim. If it does repeat this process for the Type metadata  specified by the `extends` value, until either a match is found or the `extends` claim is not present.
+1. If none of the above conditions are met, the Credential's type does not satisfy the Credential Query.
 
 ### Presentation Response
 
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3253,7 +3253,7 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 A Credential's type can be determined as satifying the Credential Query as follows:
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
-1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following `extends` claims, the Credential does not satisfy the Credential Query.
+1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
 1. If none of the above conditions are met, the Credential does not satisfy the Credential Query.
 
 ### Presentation Response
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -2889,7 +2889,7 @@ The following is an ISO mdoc specific parameter in the `meta` parameter in a Cre
 
 `doctype_value`:
 : REQUIRED. String that specifies an allowed value for the
-doctype of the requested Verifiable Credential. It MUST be a valid doctype identifier as defined in [@ISO.18013-5]. For a mdoc to satisfy the Credential Query, the value of the `docType` element in the `MobileSecurityObject` structure MUST be equal to the `doctype_value`.
+doctype of the requested Verifiable Credential. It MUST be a valid doctype identifier as defined in [@ISO.18013-5]. For an mdoc to satisfy the Credential Query, the value of the `docType` element in the `MobileSecurityObject` structure MUST be equal to the `doctype_value`.
 
 ### Parameter in the Claims Query {#mdocs_claims_query}
 
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3253,7 +3253,7 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 A Credential's type can be determined as satifying the Credential Query as follows:
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
-1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present.
+1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following `extends` claims, the Credential does not satisfy the Credential Query.
 1. If none of the above conditions are met, the Credential does not satisfy the Credential Query.
 
 ### Presentation Response
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3253,7 +3253,7 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 A Credential's type can be determined as satifying the Credential Query as follows:
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
-1. If is is not, check if the Credential has an `extends` claim. If it does repeat this process for the Type metadata  specified by the `extends` value, until either a match is found or the `extends` claim is not present.
+1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present.
 1. If none of the above conditions are met, the Credential's type does not satisfy the Credential Query.
 
 ### Presentation Response
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3250,7 +3250,7 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 `vct_values`:
 : REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
 
-A Credential's type can be determined as satifying the Credential Query as follows:
+The Wallet MUST determine whether a Credential's type satisfies a Credential Query by following these steps:
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
 1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3250,7 +3250,7 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 `vct_values`:
 : REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
 
-The Wallet MUST determine whether a Credential's type satisfies a Credential Query by following these steps:
+When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it MUST do so by following these steps:
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
 1. If the `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3253,7 +3253,7 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 The Wallet MUST determine whether a Credential's type satisfies a Credential Query by following these steps:
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
-1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
+1. If the `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
 1. If none of the above conditions are met, the Credential does not satisfy the Credential Query.
 
 ### Presentation Response
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3254,7 +3254,7 @@ A Credential's type can be determined as satifying the Credential Query as follo
 
 1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
 1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present.
-1. If none of the above conditions are met, the Credential's type does not satisfy the Credential Query.
+1. If none of the above conditions are met, the Credential does not satisfy the Credential Query.
 
 ### Presentation Response
 
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -3250,11 +3250,13 @@ The following is an SD-JWT VC specific parameter in the `meta` parameter in a Cr
 `vct_values`:
 : REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST be of a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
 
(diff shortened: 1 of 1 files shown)


1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+A Credential's type can be determined as satifying the Credential Query as follows:
+
+1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
+1. If is is not, check if the Credential has an `extends` claim. If it does repeat this process for the Type metadata  specified by the `extends` value, until either a match is found or the `extends` claim is not present.
```
fkj: typos, and I would be more explicit:
```suggestion
1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
+1. If none of the above conditions are met, the Credential's type does not satisfy the Credential Query.
```
fkj: to match other sentence structures:
```suggestion
1. If none of the above conditions are met, the Credential does not satisfy the Credential Query.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+A Credential's type can be determined as satifying the Credential Query as follows:
+
+1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
+1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present.
```
fkj: After reading SD-JWT-VC a bit more carefully, I think we should add some language about circular dependencies here
From https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-16.html#section-6.3-2:

> Consumers MUST detect such circular dependencies and reject the credential.

Maybe something like this, though it might be easier to read if we rewrite the entire paragraph to make the recursive process more explicit:
```suggestion
1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following `extends` claims, the Credential does not satisfy the Credential Query.
```

fkj (changes requested):

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+A Credential's type can be determined as satifying the Credential Query as follows:
+
+1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
+1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following `extends` claims, the Credential does not satisfy the Credential Query.
```
GarethCOliver:
```suggestion
1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+A Credential's type can be determined as satifying the Credential Query as follows:
+
+1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
+1. If `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
```
fkj: nit:

```suggestion
1. If the `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
```

fkj (changes requested):

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
 
 `vct_values`:
-: REQUIRED. A non-empty array of strings that specifies allowed values for
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+A Credential's type can be determined as satifying the Credential Query as follows:
```
fkj: Sorry to be annoying and do yet another round of suggestions, but I think this should be phrased as a normative requirement. Something like:
```suggestion
The Wallet MUST determine whether a Credential's type satisfies a Credential Query by following these steps:
```

fkj (changes requested):

fkj (approved):

c2bo (commented):
I guess this should also go into 1.0 errata?

#741 also asks about the RP/Verifier side -> Should we add a sentence that these checks must be verified by the Verifier?

Text looks good otherwise imho

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2889,8 +2889,7 @@ The following is an ISO mdoc specific parameter in the `meta` parameter in a Cre
 
 `doctype_value`:
 : REQUIRED. String that specifies an allowed value for the
-doctype of the requested Verifiable Credential. It MUST
-be a valid doctype identifier as defined in [@ISO.18013-5].
+doctype of the requested Verifiable Credential. It MUST be a valid doctype identifier as defined in [@ISO.18013-5]. For a mdoc to satisfy the Credential Query, the value of the `docType` element in the `MobileSecurityObject` structure MUST be equal to the `doctype_value`.
```
c2bo: nit
```suggestion
doctype of the requested Verifiable Credential. It MUST be a valid doctype identifier as defined in [@ISO.18013-5]. For an mdoc to satisfy the Credential Query, the value of the `docType` element in the `MobileSecurityObject` structure MUST be equal to the `doctype_value`.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
 
 `vct_values`:
-: REQUIRED. A non-empty array of strings that specifies allowed values for
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST either have or inherit from a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+The Wallet MUST determine whether a Credential's type satisfies a Credential Query by following these steps:
```
Sakurann:
```suggestion
Following steps are defined to determine whether a Credential's type satisfies a Credential Query:
```
incorporating Christian's suggestion and WG discussion to ensure these rules apply not just to the wallet but also the verifier

Sakurann:
WG discussion: makes sense to have this in errata, too and also not to limit validation rules to the wallet but also verifier.

1.1/openid-4-verifiable-presentations-1_1.md
fkj: I don't disagree with the suggestion, but I think the normative language should be preserved even when generalizing. Something like:
```suggestion
When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it MUST do so by following these steps:
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,13 @@ The following is a non-normative example of `client_metadata` request parameter
...
+
+1. Check if the `vct` in the Credential is contained in the `vct_values` array. If it is, the Credential satisfies the Credential Query.
+1. If the `vct` in the Credential is not contained in the `vct_values` array, check if the Credential has an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
```
awoie: The credential does not have an `extends` claim, it is the metadata that does. Additionally, metadata processing is optional.

awoie (changes requested):

1.1/openid-4-verifiable-presentations-1_1.md
Sakurann: can you suggest changes to the text so that it is more accurate?

brentzundel:
WG discussion suggests: Make sure it is clear this is an optional feature. @c2bo will propose text.

1.1/openid-4-verifiable-presentations-1_1.md
GarethCOliver: Maybe something like: 

```suggestion
1. If the `vct` in the Credential is not contained in the `vct_values` array, check if the Type Metadata has an `extends` claim. If it does, repeat this process for the Type Metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
```

I'm not really sure what to do with the whole 'metadata processing is optional'. If we don't process metadata how would you suggest evaluating what types it extends? Just not allow extends? That seems like a bit of an inter-op problem (if a Verifier requests a type on it's base type but a wallet only evaluates on the final type).

c2bo: Something like this?
```suggestion
1. If the `vct` in the Credential is not contained in the `vct_values` array and the Verifier supports Type metadata resolution, resolve the Type metadata and check if it contains an `extends` claim. If it does, repeat this process for the Type metadata specified by the `extends` value, until either a match is found or the `extends` claim is not present. If a circular dependency is detected while following the `extends` claims, the Credential does not satisfy the Credential Query.
```

fkj: I also don't think there's any way to allow extensions if you don't process the metadata.

GarethCOliver:
From conversation at IETF: SD-JWT-VC intends to add a claim to make it explicit what the super-types of a particular credential are (which will make this trivial and correct). 

Until that is in place we shouldn't update this text, and should update to reference it when it is ready. Note that the part of this PR related to mdoc can still go through.

1.1/openid-4-verifiable-presentations-1_1.md
GarethCOliver: conversation outdated: now explicitly just saying it matches the 'type' and we do that by seeing if it is `vct` or `aka_vcts`

GarethCOliver:
Updated text to reference the `aka_vcts` claim, making this nice and explicit.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,12 @@ The following is a non-normative example of `client_metadata` request parameter
...
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST be of a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it MUST do so by evaluating if at least one of the following true:
+
+1. The value of the `vct` claim in the Credential is contained in the `vct_values` array.
+1. The `aka_vcts` claim is present and has at least one element that is contained in the `aka_vcts` array.
```
fkj: Did you mean?

```suggestion
1. The `aka_vcts` claim is present and has at least one element that is contained in the `vct_values` array.
```

jogu (changes requested):
I think if we make this change we have to make it in 1.0 as well - otherwise it's a breaking change in 1.1 and/or 1.0 and 1.1 differ in core behaviour.

But we need to come to a conclusion on https://github.com/openid/OpenID4VP/pull/726 first really. If 1.0 is going to reference the latest SD-JWT spec then really 1.0 is going to gain this behaviour regardless, and if we agree on that then making it explicit in 1.0 is best.

(The actual changes in this PR do seem like a sensible way of addressing the problem.)

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,12 @@ The following is a non-normative example of `client_metadata` request parameter
...
 
 `vct_values`:
-: REQUIRED. A non-empty array of strings that specifies allowed values for
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST be of a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it MUST do so by evaluating if at least one of the following true:
```
jogu, line 3253: Note that this is tighter than the language in SD-JWT VC, which allows 'extends' to be considered too - https://drafts.oauth.net/oauth-sd-jwt-vc/draft-ietf-oauth-sd-jwt-vc.html#section-5.4

If the desired outcome is that only vct & aka_vcts matter, I think we should explicitly state that other data (like extends or any preexisting knowledge the wallet has of types that vct is equivalent to) MUST NOT be used.

However that feels like it would be a breaking change, so I think we need to say that Wallets may choose to return credentials with other vcts at a minimum in the case where aka_vcts is absent.

1.1/openid-4-verifiable-presentations-1_1.md
c2bo: Maybe something like this to still allow a verifier to add its own (or ecosystem defined) matching logic?

```suggestion
1. The `aka_vcts` claim is present and has at least one element that is contained in the `vct_values` array.
1. There exists out of band information that the received `vct` value is interpreted as a match with at least one element in the `vct_values` array.
```

1.1/openid-4-verifiable-presentations-1_1.md
Sakurann: see https://github.com/openid/OpenID4VP/pull/726#issuecomment-5478442979 
1.1 and 1.0 errata will reference sd-jwt vc

c2bo (approved):

1.1/openid-4-verifiable-presentations-1_1.md
GarethCOliver: Using just `aka_vcts` is by design. Using `extends` is highly problematic (both due to it being optional, and because it provides an inconsistent definition of extensibility between Wallet and Verifier). 

`aka_vcts` is introduced for that purposes to provide a consistent shared definition so that a Wallet does not return an SD-JWT that a Verifier does not understand. 

I think it would reasonable to provide it as a fallback when not present for backwards compatibility

fkj: I think that since using `extends` requires an out-of-band relationship anyway, we should not keep the fallback in the spec. Ecosystems that want to use the `extends` mechanism can still do that and the rest of us don't need to know about it. I think writing the "worse" way into the spec might make more people start doing it, which we should avoid.

brentzundel:
Discussed today, @GarethCOliver will adjust the text as in the meeting notes.

1.1/openid-4-verifiable-presentations-1_1.md
GarethCOliver: updated text to make this RECOMMENDED and added a note

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,15 @@ The following is a non-normative example of `client_metadata` request parameter
...
 
 `vct_values`:
-: REQUIRED. A non-empty array of strings that specifies allowed values for
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST be of a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it RECOMMENDED do so by evaluating if at least one of the following true:
```
javereec, line 3253:
```suggestion
When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it is RECOMMENDED do so by evaluating if at least one of the following is true:
```

javereec (approved):
Editorial remarks. Looks good otherwise.

Sakurann:
@jogu 1.0 is going to reference the latest SD-JWT spec. please re-review

awoie (approved):

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3249,11 +3248,15 @@ The following is a non-normative example of `client_metadata` request parameter
...
 
 `vct_values`:
-: REQUIRED. A non-empty array of strings that specifies allowed values for
-the type of the requested Verifiable Credential. All elements in the array MUST
-be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. The Wallet
-MAY return Credentials that inherit from any of the specified types, following
-the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
+: REQUIRED. A non-empty array of strings that specifies allowed values for the type of the requested Verifiable Credential. All elements in the array MUST be valid type identifiers as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. To satisfy the Credential Query, a Credential MUST be of a type that is included in the `vct_values` array as defined in [@!I-D.ietf-oauth-sd-jwt-vc]. 
+
+When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it is RECOMMENDED do so by evaluating if at least one of the following is true:
```
javereec, line 3253: Another small editorial change.

```suggestion
When a Wallet or Verifier needs to determine whether a Credential's type satisfies a Credential Query, it is RECOMMENDED to do so by evaluating if at least one of the following is true:
```

javereec (approved):
Editorial suggestion

brentzundel:
Discussed today, @jogu will re-review

brentzundel:
Discussed in call. Looking for @jogu re-review

PR #745, opened 2026-06-18
Clarify an empty object in a VP Token cannot be used to signify an error response

### LINKED ISSUE #743: Clarify error case when wallet returns empty vp_token vs error response vs no response
The spec allows the wallet to send a vp_token with an empty object, a vp_token with partial results, an error response, or no error response at all.

Do we want to provide implementation guidance when to send what?

@Sakurann @paulbastian @fkj @GarethCOliver @c2bo

awoie:
Fixes #743

- [x] needs errata? change 1.0?

I also improved the vp_token definition because it is a quite important part of the spec and the readability was not great.

I removed this editorially because I don't think it had any meaning anymore:
> The same rules as above apply for encoding the Presentations.


awoie:
We probably will need to apply this logic to DC API as well. I will need to double check if this is included with the current proposed text.

fkj (approved):
If people don't object to the rewriting of the definition, I agree that this makes it easier to understand what we actually mean.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1161,8 +1164,27 @@ The behavior with respect to the VP Token is unspecified for any other individua
...
+  * When `multiple` is omitted or set to `false`, the array MUST contain exactly
+    one Presentation.
+
+  * The object MUST NOT contain an entry for an optional Credential Query when
+    there are no matching Credentials for that Credential Query.
+
+  * Each Presentation MUST be encoded as a string or object according to
+    (#format_specific_parameters).
+
+  * The object MAY be empty only if the DCQL query can be satisfied without
```
c2bo: Something like this might make it more clear imho?

```suggestion
  * The object MUST NOT be empty unless the DCQL query can be satisfied without
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1161,8 +1164,27 @@ The behavior with respect to the VP Token is unspecified for any other individua
...
+  * Each key MUST be the `id` of a Credential Query in the DCQL query.
+
+  * Each value MUST be an array containing one or more Presentations matching
+    the corresponding Credential Query.
+
+  * When `multiple` is omitted or set to `false`, the array MUST contain exactly
+    one Presentation.
+
+  * The object MUST NOT contain an entry for an optional Credential Query when
+    there are no matching Credentials for that Credential Query.
```
paulbastian, line 1188: This is hard to understand for me, is this repeating things from DCQL rules? Why "optional"?

paulbastian: I think this line is rather confusing and repeating DCQL rules

Sakurann:
wg discussion, should be in the errata as well.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1161,8 +1164,27 @@ The behavior with respect to the VP Token is unspecified for any other individua
 
 When a VP Token is returned, the respective response includes the following parameters:
 
+
 `vp_token`:
-: REQUIRED. This is a JSON-encoded object containing entries where the key is the `id` value used for a Credential Query in the DCQL query and the value is an array of one or more Presentations that match the respective Credential Query. When `multiple` is omitted, or set to `false`, the array MUST contain only one Presentation. There MUST NOT be any entry in the JSON-encoded object for optional Credential Queries when there are no matching Credentials for the respective Credential Query. Each Presentation is represented as a string or object, depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Presentations.
+: REQUIRED. A JSON-encoded object subject to the following requirements:
+
+  * Each key MUST be the `id` of a Credential Query in the DCQL query.
+
```
paulbastian, line 1180: we usually don't use empty lines in enumerations

1.1/openid-4-verifiable-presentations-1_1.md
Sakurann:
```suggestion
  * The object MUST NOT be empty
```
i thought we discussed we don't want to allow empty

1.1/openid-4-verifiable-presentations-1_1.md
awoie: I know that but it was there previously. Happy to remove it but on the WG call on July 2, we decided to keep it.

1.1/openid-4-verifiable-presentations-1_1.md
awoie: But we do also in this spec doc and if I remember correctly, this won't render nicely if it was removed.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1161,8 +1164,27 @@ The behavior with respect to the VP Token is unspecified for any other individua
...
+  * The object MUST NOT contain an entry for an optional Credential Query when
+    there are no matching Credentials for that Credential Query.
+
+  * Each Presentation MUST be encoded as a string or object according to
+    (#format_specific_parameters).
+
+  * The object MUST NOT be empty
+    returning any Presentation according to
+    (#dcql_query_lang_processing_rules).
```
jogu: There's missing words or something here. This isn't a sentence that makes sense.

.DS_Store
jogu, line 1: we should remove .DS_Store

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1140,6 +1140,9 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
 
 A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query (as defined in #vp_token_request).
 
+The Wallet MUST return a VP Token only if the set of Presentations represented
+by the VP Token satisfies the requirements of the DCQL query according to (#dcql_query_lang_processing_rules).
```
jogu: There's a gap/ambiguity here I think. If all the requested credentials are set as optional, returning no credentials satisfies the DCQL query, so you can/must return a VP Token. But then the rules below leave no VP Token that is valid to return.

jogu (changes requested):

awoie:
Please review again @jogu @paulbastian @Sakurann . @jogu Note that if my memory is correct that the WG agreed to prohibit empty vp_token's even for the case where the DCQL query has only optional queries.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1144,6 +1144,17 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
...
 
+The Wallet MUST return a VP Token only if the set of Presentations returned in
+the VP Token satisfies the requirements of the DCQL query according to
+(#dcql_query_lang_processing_rules). If the Wallet does not return any
+Presentation, for example, because the End-User did not give consent, the
+requested Credentials are not available, or the DCQL query can be satisfied
+without returning any Presentation (i.e., all Credential Queries that cannot
+be fulfilled are optional), the Wallet MUST NOT return a VP Token; if the
+Wallet returns a response, it MUST be an error response as defined in
+(#error-response). In particular, an empty VP Token (a JSON object without
```
jogu, line 1154: I think my only remaining is that people might decide to return an error response without getting user consent first. I think there's existing text about this somewhere and we should probably reference that text from here?

brentzundel:
WG discussion: @awoie will make some changes to address @jogu remainning concern

brentzundel:
@paulbastian will review as well

1.0/openid-4-verifiable-presentations-1_0.md
c2bo: I guess we should point to https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_0-wg-draft.html#section-15.9 ?

awoie: Thank you. We should probably move some of the more generic content from 15.9.2 to 15.9 Error Response, and make sure the section applies to both, redirect and DC API flows. We would then need to update the reference to 15.9.2 in Annex A to include the generic and the more specific DC API-related section in the privacy considerations.

1.1/openid-4-verifiable-presentations-1_1.md
paulbastian: I've created an issue for it: #762

paulbastian (approved):

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2059,17 +2061,11 @@ Untrusted or unrecognized Request URI endpoints SHOULD be rejected or require En
 
 If the Wallet is acting within a trust framework that allows the Wallet to determine whether a Request URI belongs to a certain Client Identifier, the Wallet is RECOMMENDED to validate the Verifier's authenticity and authorization given by the Client Identifier and that the Request URI corresponds to this Verifier. If the link cannot be established in those cases, the Wallet MUST refuse the request.
 
-## Error Responses
+## Error Responses {#privacy-error-responses}
```
jogu: This might break links people have to the 1.0 spec so they no longer jump to the right section.

We could use the existing autogenerated id (`(#error-responses)` ) when linking to it instead?

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1166,7 +1179,25 @@ The behavior with respect to the VP Token is unspecified for any other individua
...
 `vp_token`:
-: REQUIRED. This is a JSON-encoded object containing entries where the key is the `id` value used for a Credential Query in the DCQL query and the value is an array of one or more Presentations that match the respective Credential Query. When `multiple` is omitted, or set to `false`, the array MUST contain only one Presentation. There MUST NOT be any entry in the JSON-encoded object for optional Credential Queries when there are no matching Credentials for the respective Credential Query. Each Presentation is represented as a string or object, depending on the format as defined in (#format_specific_parameters). The same rules as above apply for encoding the Presentations.
+: REQUIRED. A JSON-encoded object subject to the following requirements:
+
+  * Each key MUST be the `id` of a Credential Query in the DCQL query.
+
+  * Each value MUST be an array containing one or more Presentations matching
+    the corresponding Credential Query.
+
+  * When `multiple` is omitted or set to `false`, the array MUST contain exactly
```
jogu: This might make it clearer what's been talked about:

```suggestion
  * When `multiple` was omitted in the DCQL query or set to `false`, the array MUST contain exactly
```

1.0/openid-4-verifiable-presentations-1_0.md
jogu: I think this might actually be specific to DC API? e.g. 'A request for a single document type will reveal the End-User is in possession of that Credential' can I think only happen for DC API.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1144,6 +1144,19 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
 
 A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query (as defined in #vp_token_request).
 
+The Wallet MUST return a VP Token only if the set of Presentations returned in
+the VP Token satisfies the requirements of the DCQL query according to
```
jogu: Maybe this reads slightly better? I can live with existing wording though.

```suggestion
The Wallet MUST return a VP Token only if it contains a set of
  Presentations that satisfy the requirements of the DCQL query according to
```

brentzundel: @jogu I think you've got an extra 'only' in there

jogu: I do - fixed, thanks!

awoie:
@jogu You're right, `{#privacy-error-responses}` would have changed the published anchor and broken existing links. So I dropped the explicit anchor and used `(#error-responses)` for the three cross-references instead.

DC API specificity: Partly agreed. The underlying consideration isn't unique to the DC API. A Wallet in a redirect-based flow that returns an error depending on whether it can match the request leaks the same information, which is what the two SHOULD NOT paragraphs below the list already address. But you're right that the bullets were phrased as unconditional, and that absolute only holds for the DC API. Rather than moving them back under the DC API heading, I scoped the lead-in:

> Where the fact that an error response is returned, or the content of that error response, depends on the Wallet's ability to satisfy the request, the narrower a request is, the more information is revealed:

and added a note after the list:

> Note that when the Digital Credentials API is used, this can be the case even if the Wallet behaves identically in all cases, as described in (#privacy-dc-api-error).

The bullets themselves are unchanged. Let me know if you'd rather have them back in the DC API subsection.

Also synced the same changes plus your two earlier suggestions into 1.1, which had drifted from 1.0.

brentzundel:
Discussed in DCP WG call. Changes have been made and @jogu needs to re-review.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1143,6 +1143,19 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
 
 A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query (as defined in #vp_token_request).
```
jogu: Unrelated to this PR, but if I raise a separate PR it would conflict - this renders incorrectly (the #vp_token_request doesn't become a link):

Image

```suggestion
A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query, as defined in (#vp_token_request).
```

jogu (approved):

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1139,6 +1139,19 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
 
 A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query (as defined in #vp_token_request).
```
jogu: Same rendering fix for 1.1:

```suggestion
A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query, as defined in (#vp_token_request).
```

jogu (approved):
Looks good to me I think - spotted to unrelated rendering issues that would be good to include.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1143,6 +1143,19 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
...
 
+The Wallet MUST return a VP Token only if it contains a set of
+  Presentations that satisfy the requirements of the DCQL query according to
```
```suggestion
Presentations that satisfy the requirements of the DCQL query according to
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2105,27 +2136,31 @@ If the Wallet is acting within a trust framework that allows the Wallet to deter
...
+
+In the event that another component is invoked instead of the Wallet, the End-User SHOULD be informed and give consent before the invoked component returns the `wallet_unavailable` Authorization Error Response to the Verifier.
+
+### Digital Credential API Error Responses {#privacy-dc-api-error}
+
+Returning any OpenID4VP protocol error, regardless of content, can reveal additional information about the End-User’s underlying Credentials or Wallet in a way that is unique to the Digital Credentials API since reaching the Wallet can be dependent on a Wallet's ability to satisfy the request. For example, platform implementations could only allow Wallets to be selected that satisfy the request. In this case, OpenID4VP protocol error responses can only be returned by a selected Wallet and would therefore reveal that the End-User is in possession of Credentials that satisfy the request. This is in contrast to other engagement methods, in which the Wallet receives the request before learning if it can be fulfilled. What is revealed by a Wallet in those cases depends on how each individual Wallet processes the request.
```
fkj:
```suggestion
Returning any OpenID4VP protocol error, regardless of content, can reveal additional information about the End-User’s underlying Credentials or Wallet in a way that is unique to the Digital Credentials API since reaching the Wallet can be dependent on a Wallet's ability to satisfy the request. For example, platform implementations could allow only Wallets that satisfy the request to be selected. In this case, OpenID4VP protocol error responses can only be returned by a selected Wallet and would therefore reveal that the End-User is in possession of Credentials that satisfy the request. This is in contrast to other engagement methods, in which the Wallet receives the request before learning if it can be fulfilled. What is revealed by a Wallet in those cases depends on how each individual Wallet processes the request.
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -2042,27 +2073,31 @@ If the Wallet is acting within a trust framework that allows the Wallet to deter
...
+
+In the event that another component is invoked instead of the Wallet, the End-User SHOULD be informed and give consent before the invoked component returns the `wallet_unavailable` Authorization Error Response to the Verifier.
+
+### Digital Credential API Error Responses {#privacy-dc-api-error}
+
+Returning any OpenID4VP protocol error, regardless of content, can reveal additional information about the End-User’s underlying Credentials or Wallet in a way that is unique to the Digital Credentials API since reaching the Wallet can be dependent on a Wallet's ability to satisfy the request. For example, platform implementations could only allow Wallets to be selected that satisfy the request. In this case, OpenID4VP protocol error responses can only be returned by a selected Wallet and would therefore reveal that the End-User is in possession of Credentials that satisfy the request. This is in contrast to other engagement methods, in which the Wallet receives the request before learning if it can be fulfilled. What is revealed by a Wallet in those cases depends on how each individual Wallet processes the request.
```
fkj:
```suggestion
Returning any OpenID4VP protocol error, regardless of content, can reveal additional information about the End-User’s underlying Credentials or Wallet in a way that is unique to the Digital Credentials API since reaching the Wallet can be dependent on a Wallet's ability to satisfy the request. For example, platform implementations could allow only Wallets that satisfy the request to be selected. In this case, OpenID4VP protocol error responses can only be returned by a selected Wallet and would therefore reveal that the End-User is in possession of Credentials that satisfy the request. This is in contrast to other engagement methods, in which the Wallet receives the request before learning if it can be fulfilled. What is revealed by a Wallet in those cases depends on how each individual Wallet processes the request.
```

fkj (commented):
Minor editorial suggestions, but otherwise looks fine to me.

awoie:
@dastoikov @brentzundel @Sakurann this might be good to merge now.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1137,7 +1137,20 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
 
 # Response {#response}
 
-A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query (as defined in #vp_token_request).
+A VP Token is only returned if the corresponding Authorization Request contained a `dcql_query` parameter or a `scope` parameter representing a DCQL Query, as defined in (#vp_token_request).
+
+The Wallet MUST return a VP Token only if it contains a set of
+  Presentations that satisfy the requirements of the DCQL query according to
```
```suggestion
Presentations that satisfy the requirements of the DCQL query according to
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1141,7 +1141,20 @@ Additional, more complex examples can be found in (#more_dcql_query_examples).
...
+(#dcql_query_lang_processing_rules). If the Wallet does not return any
+Presentation, for example, because the End-User did not give consent, the
+requested Credentials are not available, or the DCQL query can be satisfied
+without returning any Presentation (i.e., all Credential Queries that cannot
+be fulfilled are optional), the Wallet MUST NOT return a VP Token; if the
+Wallet returns a response, it MUST be an error response as defined in
+(#error-response). In particular, an empty VP Token (a JSON object without
+any entries) MUST NOT be used to signify an error. Privacy considerations that
+apply when returning an error response, in particular with regard to End-User
+consent, are defined in (#error-responses).
```
c2bo, line 1157: Should we say something about the error to use here? I'd say `access_denied` is the natural choice.

dastoikov:
> @dastoikov @brentzundel @Sakurann this might be good to merge now.

I would need to catch up in order to properly review the PR. That said, the intent of the text is clear and the PR seems to be aligned with the WG decision.

1.0/openid-4-verifiable-presentations-1_0.md
dastoikov: > Should we say something about the error to use here? I'd say `access_denied` is the natural choice.

IMO, there is no need for that.

dpostnikov (approved):

c2bo (approved):

PR #750, opened 2026-06-29
723 duplicate claims entry

### LINKED ISSUE #723: Clarify duplicate `claims` entry
This issue was raised by the European Commission conformance team.

OID4VP says:

> claims: OPTIONAL. A non-empty array of objects as defined in [Section 6.3](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#claims_query) that specifies claims in the requested Credential. Verifiers MUST NOT point to the same claim more than once in a single query. Wallets SHOULD ignore such duplicate claim queries.

- It is unclear what exactly is meant by "point to the same claim more than once".
- It is also unclear what it means in practice that duplicate claim queries "SHOULD be ignored".

OID4VP further says:

> Each entry in claims MUST be an object with the following properties:
> id: REQUIRED if claim_sets is present in the Credential Query; OPTIONAL otherwise. A string identifying the particular claim. The value MUST be a non-empty string consisting of alphanumeric, underscore (_), or hyphen (-) characters. Within the particular claims array, the same id MUST NOT be present more than once.
> path: REQUIRED The value MUST be a non-empty array representing a claims path pointer that specifies the path to a claim within the Credential, as defined in [Section 7](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#claims_path_pointer).
> values: OPTIONAL A non-empty array of strings, integers or boolean values that specifies the expected values of the claim. If the values property is present, the Wallet SHOULD return the claim only if the type and value of the claim both match exactly for at least one of the elements in the array. Details of the processing rules are defined in [Section 6.4.1](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#selecting_claims).

My interpretation is that "point to the same claim" refers to claims with the same path value, since path is the actual pointer to the claim. Would you agree? I think OID4VP 1.1, and potentially errata as well, should clarify what exactly "point to" means in this context.

Following that interpretation, I would assume that the following two entries are duplicates, even though their `id` and `values` are different:

```
claims = [
  { id = 1, path = [ name ], values = [ oliver ] }, 
  { id = 2, path = [ name ], values = [ colette ] },
  { id = 3, path = [ foo ] }, 
  { id = 4, path = [ bar ] }
],
claim_sets = [ [ 1, 3 ], [ 2, 4 ] ]
```

If that is correct, should both duplicate entries be ignored, or only every occurrence after the first one?

More generally, the specification should clarify whether:
- all duplicate claims are to be ignored, or
- only subsequent occurrences after the first are to be ignored.

I am also not sure why this is specified as a SHOULD rather than a MUST. In which situations would it be acceptable, or even necessary, to process duplicate claim queries instead of ignoring them?

@danielfett @c2bo @paulbastian @mickrau @jogu (cc @brentzundel, potentially for 1.1?)

c2bo:
Closes #723

Open question from the issue. SHOULD -> MUST?

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -780,8 +780,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim more than
-once in a single query. Wallets SHOULD ignore such duplicate claim queries.
+claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
+path pointer more than once in a single query. Wallets SHOULD ignore such duplicate claim queries (ignore the
+duplicate, but properly process to the first occurence).
 
 `claim_sets`:
 : OPTIONAL. A non-empty array containing arrays of identifiers for
@@ -3599,6 +3600,7 @@ The technology described in this specification was made available from contribut
    * Updated origin examples to remove trailing slash
    * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
    * Clarified that request_uri_method is a case-sensitive string
+   * Clarify duplicate claims entries
    
 -final
    
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -3611,7 +3611,7 @@ The technology described in this specification was made available from contribut
    * Updated origin examples to remove trailing slash
    * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
    * Clarified that request_uri_method is a case-sensitive string
-   * Clarify duplicate claims entries
+   * Remove requirements for duplicate claim entries
    
 -final
    
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -781,8 +781,8 @@ proof.
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
 claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
-path pointer more than once in a single query. Wallets SHOULD ignore such duplicate claim queries (ignore the
-duplicate, but properly process to the first occurence).
+path pointer more than once in a single query. Wallets SHOULD discard such duplicate claim queries (ignore the
+duplicate, but properly process the first occurrence).
 
 `claim_sets`:
 : OPTIONAL. A non-empty array containing arrays of identifiers for
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -780,9 +780,7 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT include queries pointing to the same claim as identified
-by the claims path pointer more than once, unless they are requesting different values for that claim.
-If such duplicate claim queries are detected, Wallets SHOULD process the first occurrence and ignore duplicates.
+claims in the requested Credential.
 
 `claim_sets`:
 : OPTIONAL. A non-empty array containing arrays of identifiers for
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -780,9 +780,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
-path pointer more than once in a single query. Wallets SHOULD discard such duplicate claim queries (ignore the
-duplicate, but properly process the first occurrence).
+claims in the requested Credential. Verifiers MUST NOT include queries pointing to the same claim as identified
+by the claims path pointer more than once, unless they are requesting different values for that claim.
+If such duplicate claim queries are detected, Wallets SHOULD process the first occurrence and ignore duplicates.
 
 `claim_sets`:
 : OPTIONAL. A non-empty array containing arrays of identifiers for
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -776,9 +776,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
-path pointer more than once in a single query. Wallets SHOULD ignore such duplicate claim queries (ignore the
-duplicate, but properly process to the first occurence).
+claims in the requested Credential. Verifiers MUST NOT include queries pointing to the same claim as identified
+by the claims path pointer more than once, unless they are requesting different values for that claim.
+If such duplicate claim queries are detected, Wallets SHOULD process the first occurrence and ignore duplicates.
 
 `claim_sets`:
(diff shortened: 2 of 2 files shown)


martijnharing:
Alternatively to the solution proposed in this PR, what if we remove the restrictions on the RP and requirements on the wallet to check anything related to duplicates?
It feels like there isn't a direct security or privacy problem without those requirements / recommendations and it would mean we don't need to define what we mean exactly with duplicate.

fkj:
I agree with @martijnharing that it might be better to just remove this text. Is there a specific argument for this text to exist?

paulbastian:
> Alternatively to the solution proposed in this PR, what if we remove the restrictions on the RP and requirements on the wallet to check anything related to duplicates? It feels like there isn't a direct security or privacy problem without those requirements / recommendations and it would mean we don't need to define what we mean exactly with duplicate.

Despite any personal opinion, removing the entire text about duplicate entries is likely a breaking change.

paulbastian (approved):
I think it adds clarity while avoiding a breaking change.

fkj (approved):
I'm OK with doing this to clarify without a breaking change, though it's still not clear to me why this text exists in the first place.

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -780,8 +780,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim more than
-once in a single query. Wallets SHOULD ignore such duplicate claim queries.
+claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
+path pointer more than once in a single query. Wallets SHOULD ignore such duplicate claim queries (ignore the
+duplicate, but properly process to the first occurence).
```
jogu: typos:


```suggestion
duplicate, but properly process the first occurrence).
```

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -780,8 +780,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim more than
-once in a single query. Wallets SHOULD ignore such duplicate claim queries.
+claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
+path pointer more than once in a single query. Wallets SHOULD ignore such duplicate claim queries (ignore the
```
jogu: I don't think fully resolves the example Oliver raised in the issue:

```
claims = [
  { id = 1, path = [ name ], values = [ oliver ] }, 
  { id = 2, path = [ name ], values = [ colette ] },
  { id = 3, path = [ foo ] }, 
  { id = 4, path = [ bar ] }
],
claim_sets = [ [ 1, 3 ], [ 2, 4 ] ]
```

It's now clear that you ignore '2' in claims. But what do you do about [2, 4] in the claim_sets? I think it would be sensible to ignore too, but someone might make the argument '2' should be treated as '1'.

Possible this is clearer:



```suggestion
claims in the requested Credential. Verifiers MUST NOT point to the same claim as identified by the claims
path pointer more than once in a single query. Wallets SHOULD discard such duplicate claim queries (ignore the
```


I'm also not exactly clear on the new language, does it mean:

```
claims = [
  { id = 1, path = [ "nationalities", 0 ] }, 
  { id = 2, path = [ "nationalities", null ] }
}
```

is treated as a duplicate, i.e. '2' is discarded?

jogu (changes requested):

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -776,8 +776,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim more than
-once in a single query. Wallets SHOULD ignore such duplicate claim queries.
+claims in the requested Credential. Verifiers MUST NOT include queries pointing to the same claim as identified
+by the claims path pointer more than once, unless they are requesting different values for that claim.
+If such duplicate claim queries are detected, Wallets SHOULD process the first occurrence and ignore duplicates.
```
jogu: I'm still not sure what this means for this case, does it mean:

```
claims = [
  { id = 1, path = [ "nationalities", 0 ] }, 
  { id = 2, path = [ "nationalities", null ] }
}
```

is treated as a duplicate, i.e. '2' is discarded?

The latest change causes a change in the behaviour for this query:

```
claims = [
  {  path = [ name ], values = [ oliver ] }, 
  {  path = [ name ], values = [ colette ] }
]
```

I think it's gone from 'ignore the 'collete' line' (in 1.0 final) to 'process that line but the result is a query that can never match a credential'.


I kind of feel like we're going the wrong direction down a rabbit hole, but I'm not sure what the right direction is.

My suspicion is the original text was really only trying to say that the second line here has no meaning:

```
claims = [
  {  path = [ name ] }, 
  {  path = [ name ] }
]
```

martijnharing:
> > Alternatively to the solution proposed in this PR, what if we remove the restrictions on the RP and requirements on the wallet to check anything related to duplicates? It feels like there isn't a direct security or privacy problem without those requirements / recommendations and it would mean we don't need to define what we mean exactly with duplicate.
> 
> Despite any personal opinion, removing the entire text about duplicate entries is likely a breaking change.

I'm not sure if it's actually a breaking change.

It would allow an RP to include a duplicate in the request, but I don't think that actually breaks anything. 
If we consider that duplicate entry referred to exact duplicates (e.g. two path strings that are equal and no value elements that are different) then whether you ignore one of them or not doesn't make a difference.

brentzundel:
WG discussion: let's try removing the lines as recommended by @martijnharing

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -780,8 +780,9 @@ proof.
 
 `claims`:
 : OPTIONAL. A non-empty array of objects as defined in (#claims_query) that specifies
-claims in the requested Credential. Verifiers MUST NOT point to the same claim more than
-once in a single query. Wallets SHOULD ignore such duplicate claim queries.
+claims in the requested Credential. Verifiers MUST NOT include queries pointing to the same claim as identified
+by the claims path pointer more than once, unless they are requesting different values for that claim.
+If such duplicate claim queries are detected, Wallets SHOULD process the first occurrence and ignore duplicates.
```
```suggestion
claims in the requested Credential.
```

1.1/openid-4-verifiable-presentations-1_1.md
```suggestion
claims in the requested Credential.
```

fkj (approved):

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -3612,6 +3611,7 @@ The technology described in this specification was made available from contribut
    * Updated origin examples to remove trailing slash
    * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
    * Clarified that request_uri_method is a case-sensitive string
+   * Clarify duplicate claims entries
```
martijnharing:
```suggestion
   * Remove requirements for duplicate claim entries
```

It's a bit of a nit, but strictly speaking we aren't really clarifying anything. For someone not reading this issue, this log text is more helpful.

martijnharing (approved):
Proposed changing the changelog text, but not gating approval on that.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -3688,4 +3687,5 @@ The technology described in this specification was made available from contribut
    * Clarified that Multi-RP-sig section means Verifier Info instead of attestations
    * Updated origin examples to remove trailing slash
    * Clarified that request_uri_method is a case-sensitive string
+   * Clarify duplicate claims entries
```
```suggestion
   * Remove requirements for duplicate claim entries
```

brentzundel:
@jogu the PR has been adjusted to match the direction the group felt was best during our last call. 
Please review again.

brentzundel (approved):

Sakurann:
@jogu please re-review text you commented on got removed with the new direction

jogu (approved):

PR #752, opened 2026-07-03
clarify encryption keys use field

### LINKED ISSUE #642: Should we mandate the `use` field for keys in the JSON Web Key Set in client_metadata

c2bo:
Closes #642

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -1316,7 +1316,7 @@ To encrypt the Authorization Response, implementations MUST use an unsigned, enc
 
 To obtain the Verifier's public key to which to encrypt the Authorization Response, the Wallet uses JWKs from client metadata (such as the `jwks` member within the `client_metadata` request parameter or other mechanisms as allowed by the given Client Identifier Prefix).
 Using what it supports and its preferences, the Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. 
-JWKs whose `use` parameter is absent or has the value `enc` are for encryption of the Authorization Response. The Wallet MUST NOT select a key whose `use` parameter indicates a different purpose (e.g., `sig`).
+JWKs whose `use` parameter is absent or has the value `enc` are for encryption of the Authorization Response. The Wallet SHOULD NOT select a key whose `use` parameter indicates a different purpose (e.g., `sig`).
 The `alg` parameter MUST be present in the JWKs.
 The JWE `alg` algorithm used MUST be equal to the `alg` value of the chosen `jwk`.
 If the selected public key contains a `kid` parameter, the JWE MUST include the same value in the `kid` JWE Header Parameter (as defined in [@!RFC7516, Section 4.1.6]) of the encrypted response. This enables the Verifier to easily identify the specific public key that was used to encrypt the response.
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -302,7 +302,7 @@ This specification defines the following new request parameters:
 `client_metadata`:
 : OPTIONAL. A JSON object containing the Verifier metadata values. It MUST be UTF-8 encoded. The following metadata parameters MAY be used:
 
-    * `jwks`: OPTIONAL. A JSON Web Key Set, as defined in [@!RFC7591], that contains one or more public keys, such as those used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#response_encryption)), or where the Wallet will require the public key of the Verifier to generate a Verifiable Presentation. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests. Each JWK in the set MUST have a `kid` (Key ID) parameter that uniquely identifies the key within the context of the request.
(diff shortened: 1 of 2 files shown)


Sakurann (approved):

c2bo:
@QZHelen could i ask for your review as well? Couldn't select you for reviewers (probably because you didn't joint the gh org?).

fkj (commented):
The requirement that the Wallet MUST NOT select a key with the wrong `use` is new normative text, and surprisingly RFC 7517 does not contain any normative text saying what to actually do with a `use` parameter (e.g. that you MUST or SHOULD use a `use=enc` key only for encryption).

So this is technically a breaking change. However, I think any implementation that ignores `use` will never actually work anyway. So in this case I think we can make an exception. But this should be discussed in the WG. One option is to weaken the MUST NOT to a SHOULD NOT.

QZHelen:
Looks good to me, many thanks for the work to clarify this!

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -302,7 +302,7 @@ This specification defines the following new request parameters:
...
+    * `jwks`: OPTIONAL. A JSON Web Key Set, as defined in [@!RFC7591], that contains one or more public keys, such as those used by the Wallet as an input to a key agreement that may be used for encryption of the Authorization Response (see (#response_encryption)), or where the Wallet will require the public key of the Verifier to generate a Verifiable Presentation. This allows the Verifier to pass ephemeral keys specific to this Authorization Request. JWKs in this set that do not contain a `use` parameter are intended to be used for encryption of the Authorization Response. If the Verifier includes keys intended for different purposes, it SHOULD set the `use` parameter in each key of the set to avoid ambiguity about which key to use for which purpose. Public keys included in this parameter MUST NOT be used to verify the signature of signed Authorization Requests. Each JWK in the set MUST have a `kid` (Key ID) parameter that uniquely identifies the key within the context of the request.
```
paulbastian, line 305: Is this a breaking change?

paulbastian (changes requested):
My opinion is that "enc" actually fits well for the Credential Response encryption, while ECDH+MAC-based cryptographic holder binding should actually use `sig` value or register a new value for the `use` parameter. I believe the current language is a breaking change which we wanted to avoid.

c2bo:
> My opinion is that "enc" actually fits well for the Credential Response encryption, while ECDH+MAC-based cryptographic holder binding should actually use `sig` value or register a new value for the `use` parameter. I believe the current language is a breaking change which we wanted to avoid.

That is pretty much what the PR says? It just adds a simplification that if `use` isn't present, then encryption should be assumed as a simplification / leniency clause (since we've already seen implementations do that and it's strictly speaking not required right now).

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1316,6 +1316,7 @@ To encrypt the Authorization Response, implementations MUST use an unsigned, enc
 
 To obtain the Verifier's public key to which to encrypt the Authorization Response, the Wallet uses JWKs from client metadata (such as the `jwks` member within the `client_metadata` request parameter or other mechanisms as allowed by the given Client Identifier Prefix).
 Using what it supports and its preferences, the Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. 
+JWKs whose `use` parameter is absent or has the value `enc` are for encryption of the Authorization Response. The Wallet MUST NOT select a key whose `use` parameter indicates a different purpose (e.g., `sig`).
```
```suggestion
JWKs whose `use` parameter is absent or has the value `enc` are for encryption of the Authorization Response. The Wallet SHOULD NOT select a key whose `use` parameter indicates a different purpose (e.g., `sig`).
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1312,6 +1312,7 @@ To encrypt the Authorization Response, implementations MUST use an unsigned, enc
 
 To obtain the Verifier's public key to which to encrypt the Authorization Response, the Wallet uses JWKs from client metadata (such as the `jwks` member within the `client_metadata` request parameter or other mechanisms as allowed by the given Client Identifier Prefix).
 Using what it supports and its preferences, the Wallet selects the public key to encrypt the Authorization Response based on information about each key, such as the `kty` (Key Type), `use` (Public Key Use), `alg` (Algorithm), and other JWK parameters. 
+JWKs whose `use` parameter is absent or has the value `enc` are for encryption of the Authorization Response. The Wallet MUST NOT select a key whose `use` parameter indicates a different purpose (e.g., `sig`).
```
```suggestion
JWKs whose `use` parameter is absent or has the value `enc` are for encryption of the Authorization Response. The Wallet SHOULD NOT select a key whose `use` parameter indicates a different purpose (e.g., `sig`).
```

paulbastian (approved):
approved after proper re-reading

awoie (approved):

1.0/openid-4-verifiable-presentations-1_0.md
brentzundel: reduction in normative requirements would not be a breaking change

brentzundel:
Multiple positive reviews, sufficient review period, requests for changes all addressed, merging

PR #748, opened 2026-06-22
Security considerations on untrusted input

c2bo:
Fixes the concerns raised in #647 and broadens the scope to untrusted JSON input

Should this also go in 1.0?

PR #759, opened 2026-07-13
Add text on untrusted input to 1.0

c2bo:
Mirrors the changes of #748 for 1.0 errata.

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -1919,6 +1919,15 @@ these constraints. The Wallet is not controlled by the Verifier and the Verifier
 MUST perform its own security checks on the returned Credentials and
 Presentations.
 
+## Parsing of untrusted inputs
+
+Wallets MUST treat all incoming requests as untrusted input. To mitigate injection and resource exhaustion attacks, Wallets MUST implement input validation on the Authorization Request and its enclosed DCQL query.
+
+Wallets SHOULD implement the following steps:
+
+* Enforce input validation: Verify that the Authorization Request and DCQL query contain no malformed properties. Unknown parameters MUST be ignored.
+* Apply resource limits: Enforce maximum length restrictions on strings, maximum depths for nested objects, and maximum item counts for arrays within the query.
+
 # Privacy Considerations {#privacy-considerations}
 
 Many privacy considerations are specific to the Credential format and associated proof type used in a particular Presentation.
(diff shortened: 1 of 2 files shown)


fkj (approved):

paulbastian (approved):

awoie (approved):

martijnharing:
Given this text, is the following an accurate interpretation?

In the following cases any wallet is mandated to abort processing if:
- There is a credential query id field that contains the character ~
- There is a nonce field that contains the character ±

If the wallet supports encrypted response, it is mandated to abort processing if:
- There is an encrypted_response_enc_values_supported field that contains the character ±

If the wallet supports the authority key identifier, it is mandated to abort processing if:
- There is an authority key identifier that's not valid base64url

If the wallet supports signed requests and the wallet supports the x5c header, it is mandated to abort processing if:
-  The x5c header contains an element that's not a X.509 certificate

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1934,6 +1934,15 @@ these constraints. The Wallet is not controlled by the Verifier and the Verifier
 MUST perform its own security checks on the returned Credentials and
 Presentations.
 
+## Parsing of untrusted inputs
+
+Wallets MUST treat all incoming requests as untrusted input. To mitigate injection and resource exhaustion attacks, Wallets MUST implement input validation on the Authorization Request and its enclosed DCQL query.
+
+Wallets SHOULD implement the following steps:
```
lj-raidiam:
```suggestion
Wallets SHOULD implement at least the following steps:
```

c2bo: I am ok with that change (although I don't believe it changes a lot), but then we should change the 1.1 text accordingly (which was already merged in a previous PR)

fkj:
> Given this text, is the following an accurate interpretation?

@martijnharing After thinking through this a bit more, I think the pedantic answer is no.

This says that a Wallet MUST do input validation, but the specific steps are only a SHOULD. So if you are a very pedantic Wallet, you probably could argue in the direction of doing "input validation" but not actually doing any of the things we consider to be input validation. I think that would be a case of malicious compliance that only hurts the Wallet itself.

Orthogonally, this proposed text also does not say that the Wallet MUST do anything specific if the validation fails. If you want to strengthen it, we could add some text like we have in other places: "The Wallet MUST abort processing and return an error if any input did not pass the validation." I guess the error should be an `invalid_request`.

1.0/openid-4-verifiable-presentations-1_0.md
Sakurann: WG discussion. Christian to push this update to both files

Sakurann:
wg discussion
if after the discussions following your question, you believe there is a change needed to the spec, @martijnharing please open an issue about wallet behavior when input validation fails. if spec text stays the current way, wallet is not forced to abort a request whose validation failed.

PR #761, opened 2026-07-19
Add security guidance for platform-specific Origins and DC API.

### LINKED ISSUE #646: Handling of `origin` in Native Mobile Platform App to App flows for DC API

fkj:
I have only added the guidance to 1.1 so far, but I guess we should also add it to 1.0 as an errata.
This is purely clarification of existing expectations of the platforms.

This resolves #646.


1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -117,7 +117,7 @@ Issuer-Holder-Verifier Model:
...
+: An identifier for the calling website or native application, asserted by the underlying web or application platform. For Web-based callers, the Origin is an opaque `origin` or `tuple origin` as defined in the "Origins" section of [@!whatwg.html]. For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The corresponding web Origin is https://verify.example.com, where https is the scheme, verify.example.com is the host, and the port is not explicitly included because 443 is the default port for https. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0. The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string (but see #dc_api_origin_security_considerations for the expected properties of platform-specific application Origins).
```
```suggestion
: An identifier for the calling website or native application, asserted by the underlying web or application platform. For Web-based callers, the Origin is an opaque `origin` or `tuple origin` as defined in the "Origins" section of [@!whatwg.html]. For native application callers, the Origin follows a platform-specific convention. Some platforms use a linked web origin, while others use a platform-specific application origin. For example, the Verifier for the organization MyExampleOrg is served from https://verify.example.com. The corresponding web Origin is https://verify.example.com, where https is the scheme, verify.example.com is the host, and the port is not explicitly included because 443 is the default port for https. On some platforms, a native application uses the same Origin, while on others it uses a platform-specific value such as platform:pkg-key-hash:Z4OFzVVSZrzTRa3eg79hUuHy12MVW0vzPDf4q4zaPs0. The Origin provided by the underlying platform is outside the scope of this specification and MUST be treated as an opaque string (see (#dc_api_origin_security_considerations) for the expected properties of platform-specific application Origins).
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2677,6 +2677,15 @@ The following security considerations from OpenID4VP apply:
 * Security Checks on the Returned Credentials and Presentations as described in (#dcql_query_security).
 * DCQL Value Matching as described in (#dcql-value-matching).
 
+### Use of platform-specific application origins {#dc_api_origin_security_considerations}
+
+Some platforms use platform-specific application Origins with the Digital Credentials API.
```
c2bo, line 2682: It's strictly speaking not DC API and we should probably make that clear: the OS layer credential management APIs are not governed by the W3C specification, but they might choose to allow local (app-to-app) requests that re-use DC API definitions.

fkj: Is the Origin not transported through the DC API? I think this might fall under the general clause in the beginning of the overall section that says that by DC API we mean also other things that look like the DC API. Or can you suggest some alternative text to clarify what you mean?

c2bo: I didn't remember the terminology section. I think it is fine with the terminology

c2bo (approved):

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2677,6 +2677,15 @@ The following security considerations from OpenID4VP apply:
...
+While the Origin values used by platforms and the transport of the request and Origin to the Wallet are outside the scope of this specification, the security of the protocol rests on the following properties:
+
+* The platform must provide collision-resistant values when assigning Origins.
+* The platform must ensure that each application using the platform is assigned a unique Origin.
+* The platform must ensure that it transports the Origin value to the Wallet unmodified, such that values can be compared to the `expected_origins` using exact string comparison.
```
```suggestion
* The platform provides collision-resistant values when assigning Origins.
* The platform ensures that each application using the platform is assigned a unique Origin.
* The platform ensures that it transports the Origin value to the Wallet unmodified, such that values can be compared to the `expected_origins` using exact string comparison.
```

Sakurann:
WG call:
the goal is to make sure this PR is not factually wrong.
@martijnharing to review text in this PR to ensure assumptions about platform behavior is not off. (there is no iOS implementation to verify what is in this PR.)
@QZHelen would appreciate if you can approve this PR in addition to issue comment.
@fkj to open another issue to revise text about platforms to make it generic and future-proofing

QZHelen (approved):

Sakurann (approved):

dpostnikov:
Outstanding task: @martijnharing to review text in this PR to ensure assumptions about platform behavior is not off. (there is no iOS implementation to verify what is in this PR.)

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -2677,6 +2677,15 @@ The following security considerations from OpenID4VP apply:
...
 * DCQL Value Matching as described in (#dcql-value-matching).
 
+### Use of platform-specific application origins {#dc_api_origin_security_considerations}
+
+Some platforms use platform-specific application Origins with the Digital Credentials API.
+While the Origin values used by platforms and the transport of the request and Origin to the Wallet are outside the scope of this specification, the security of the protocol rests on the following properties:
+
+* The platform provides collision-resistant values when assigning Origins.
+* The platform ensures that each application using the platform is assigned a unique Origin.
+* The platform ensures that it transports the Origin value to the Wallet unmodified, such that values can be compared to the `expected_origins` using exact string comparison.
```
martijnharing: This last one only applies when it's not the platform generating the origin correct? If so, do we need to say that?

When any origin is generated (including by the platform), the generation and/or transportation of origin must be done in such a way that both the Wallet and the RP must be able to generate the (exact) same value for the origin.

fkj: Yes, or if the platform generates any kind of "seed" that is then used to generate the origin. The difficult part is writing this in a way that does not assume anything about the way the origin is obtained by/provided to the wallet/verifier.
You're right that this is more of a "it will not work" consideration than a security consideration, so we could consider just dropping it.

brentzundel:
Discussed in DCP WG call. Proposal is to remove the last line, as it add more confusion than guidance.

1.1/openid-4-verifiable-presentations-1_1.md
fkj: I dropped it in 3cf9b59.

dpostnikov (approved):

PR #775, opened 2026-08-08
fix cross-reference for the origin Client Identifier Prefix

marcoscaceres:
The `origin` Client Identifier Prefix says it is defined in Appendix A.2, Request, which does not mention it; the rule is in Appendix A.4, Response, which three other cross-references in this document already point at. Changed in both 1.0 and 1.1.


c2bo (approved):

jogu (approved):

marcoscaceres:
(not sure what the process is, but I guess someone will merge this for me 🙏 )

fkj (approved):

dpostnikov (approved):

PR #776, opened 2026-08-08
complete the https to openid_federation prefix rename in the invalid_request example

marcoscaceres:
#401 renamed the `https` Client Identifier Prefix to `openid_federation`, including in the sentence immediately below the prefix list, but this occurrence in the `invalid_request` example was missed. Changed in both 1.0 and 1.1.


jogu (approved):
Looks correct to me, thanks Marcos.

c2bo (approved):
To be fair, using `https` as a client-id Prefix is a valid example for `invalid_request` (since it doesn't exist anymore) :sweat_smile:

fkj (approved):

dpostnikov (approved):

PR #782, opened 2026-08-20
Normatively define how an SD-JWT VC is returned in vp_token

### LINKED ISSUE #780: It's not normatively defined how to return an SD-JWT VC
When defining how credential are returned in vp_token, we say in https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#name-response-parameters:

> Each Presentation is represented as a string or object, depending on the format as defined in Appendix B

However https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#name-ietf-sd-jwt-vc doesn't actually define this for SD-JWT VC.

The obvious answer seems to be to say it needs to be a string containing the compact serialisation, and this is mostly backed up by https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-18.html#section-2.2 which says:

> A presentation of an SD-JWT VC MUST be encoded as an SD-JWT or as an SD-JWT+KB defined in Section 4 or Section 8 of [[RFC9901](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-18.html#RFC9901)]. By default, the format defined in Section 4 of [[RFC9901](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-18.html#RFC9901)] is used, whereas support for the JWS JSON Serialization in Section 8 of [[RFC9901](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-18.html#RFC9901)] is OPTIONAL.

(Tangentially related to https://github.com/openid/OpenID4VC-HAIP/issues/380 which suggests that the JSON serialisation is not actually usable in practice in OpenID4VP currently.)

jogu:
Specify that each SD-JWT VC presentation is a string containing an SD-JWT or SD-JWT+KB in the compact serialized format from Section 4 of the SD-JWT specification, and that the JWS JSON Serialization must not be used, as agreed in the working group discussion today.

closes #780

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -3226,6 +3226,8 @@ the inheritance logic defined in [@!I-D.ietf-oauth-sd-jwt-vc].
 
 ### Presentation Response
 
+Each Presentation is represented as a string containing an SD-JWT or SD-JWT+KB in the compact serialized format defined in Section 4 of [@!I-D.ietf-oauth-selective-disclosure-jwt]. The JWS JSON Serialization defined in Section 8 of [@!I-D.ietf-oauth-selective-disclosure-jwt] MUST NOT be used with the `dc+sd-jwt` Credential Format Identifier.
+
 A non-normative example DCQL query using the SD-JWT VC format is shown in (#dcql_query_example).
 The respective response is shown in (#response_dcql_query).
 
@@ -3639,6 +3641,7 @@ The technology described in this specification was made available from contribut
    * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
    * Clarified that request_uri_method is a case-sensitive string
    * Remove requirements for duplicate claim entries
+   * Clarify `dc+sd-jwt` presentations use the compact serialized SD-JWT format
    
 -final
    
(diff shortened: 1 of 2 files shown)


c2bo (approved):

dpostnikov:
Discussed on the APAC call today:
- Feedback from the WG was requested.
- @jogu, to update for 1.0 as well?
- "The JWS JSON Serialization defined in Section 8 of [@!I-D.ietf-oauth-selective-disclosure-jwt] MUST NOT be used." need to be further qualified.

c2bo:
> * @jogu, to update for 1.0 as well?

Agreed, this should imho go into 1.0 errata as well

jogu:
Both points resolved. I did actually mean to put it into 1.0 errata but forget. Then remembered and forgot again. So I agree it should be there.

brentzundel (approved):

brentzundel:
Discussion in group: We need to see if this is needed after #726. It may work to land this one anyway.

jogu:
> Discussion in group: We need to see if this is needed after #726. It may work to land this one anyway.

Even if we reference a version of SD-JWT VC that includes the changes in https://github.com/oauth-wg/oauth-sd-jwt-vc/pull/425/changes I think we still need this text to make clear that, even if an ecosystem decides to fully define a JSON serialisation, that it needs to have a different credential format identifier than the compact serialisation.

Sakurann (approved):

PR #788, opened 2026-08-27
Clarify Request URI response can be unsigned for prefixes requiring unsigned requests

### LINKED ISSUE #756: Clarification on Request URI response
## Description

Requesting clarification on whether the `request_uri` response can be an unsigned JWS or if it is required to be signed.

## Use Case

For clients using the `redirect_uri` prefix, signed requests are not supported. In this scenario, is reference mode using `request_uri` achievable with an unsigned `request_uri` response?

Example JWT header:

```json
{
  "alg": "none"
}
```

Example JWS:

```
eyJhbGciOiJub25lIn0./*Payload in B64 encoded form*/.
```

## References

The following references indicate that the `request_uri` response is expected to be signed:

**OpenID4VP — Request URI Response**

> The Request URI response MUST be an HTTP response with the content type `application/oauth-authz-req+jwt` and the body being a signed, optionally encrypted ([[OpenID4VP ref](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.10.1-1)](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.10.1-1)).

**RFC 9101 — Authorization Request Object**

> The Authorization Request Object MUST be one of the following:
> (a) JWS signed
> (b) JWS signed and JWE encrypted
> ([[RFC 9101 ref](https://www.rfc-editor.org/info/rfc9101/#section-5-7)](https://www.rfc-editor.org/info/rfc9101/#section-5-7))

Based on these requirements, clarification is requested on whether an unsigned `request_uri` response is permitted or if signing is mandatory.

jogu:
The Request URI Response section said the body must be a signed request object, which conflicts with the redirect_uri Client Identifier Prefix rule that requests cannot be signed. Clarify that the request object is signed unless the Client Identifier Prefix in use requires an unsigned request, in which case the JWT uses alg none.

Fixes #756

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -660,7 +660,7 @@ wallet_nonce=qPmxiNFCR3QTm19POc8u
 
 ### Request URI Response
 
-The Request URI response MUST be an HTTP response with the content type `application/oauth-authz-req+jwt` and the body being a signed, optionally encrypted, request object as defined in [@RFC9101]. The request object MUST fulfill the requirements as defined in (#vp_token_request).
+The Request URI response MUST be an HTTP response with the content type `application/oauth-authz-req+jwt` and the body being a request object as defined in [@RFC9101], optionally encrypted. The request object is signed unless the Client Identifier Prefix in use requires an unsigned request (e.g., `redirect_uri`), in which case the JWT uses the `alg` value `none`. The request object MUST fulfill the requirements as defined in (#vp_token_request).
 
 The following is a non-normative example of a payload for a request object:
 
@@ -3670,6 +3670,7 @@ The technology described in this specification was made available from contribut
    * Add assumption that the platform that provides the Origin is trusted
    * Clarified intent_to_retain value when not present
    * Removed invalid_scope error guidance as duplicate of RFC6749
+   * Clarify that the Request URI response is unsigned (JWT with `alg` value `none`) when the Client Identifier Prefix in use requires unsigned requests
    * Clarified that Multi-RP-sig section means Verifier Info instead of attestations
    * Updated origin examples to remove trailing slash
    * Clarify that `aud` corresponds to `issuer` Wallet Metadata paremeter if Dynamic Discovery is used
--- a/1.1/openid-4-verifiable-presentations-1_1.md
+++ b/1.1/openid-4-verifiable-presentations-1_1.md
@@ -656,7 +656,7 @@ wallet_nonce=qPmxiNFCR3QTm19POc8u
 
 ### Request URI Response
 
-The Request URI response MUST be an HTTP response with the content type `application/oauth-authz-req+jwt` and the body being a signed, optionally encrypted, request object as defined in [@RFC9101]. The request object MUST fulfill the requirements as defined in (#vp_token_request).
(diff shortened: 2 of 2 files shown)


charsleysa (approved):

Sakurann (approved):

paulbastian (changes requested):
I'm not sure RFC9101 allows that. There is text saying `It MUST also reject the request if
   the Request Object uses an "alg" value of "none".`

jogu:
@paulbastian 

> I'm not sure RFC9101 allows that. There is text saying `It MUST also reject the request if the Request Object uses an "alg" value of "none".`

That's in the section where 'require_signed_request_object' is in the client metadata as 'true':

Screenshot 2026-08-31 at 15 36 05

so I don't think applies here. none is generally permitted in RFC9101.

paulbastian:
But the next paragraph says: 
```
   When the value of it as server metadata is "true", then the server
   MUST reject the authorization request from any client that does not
   conform to this specification.  It MUST also reject the request if
   the Request Object uses an "alg" value of "none".  If omitted, the
   default value is "false".
```
The text isn't a 100% clear if that sentence is conditional to the first sentence.

fkj:
RFC 9101 also says (in section 10.1, which is Security Considerations):
> When sending the Authorization Request Object through the request parameter, it MUST be either signed using [JWS](https://www.rfc-editor.org/info/rfc9101/#RFC7515) [[RFC7515](https://www.rfc-editor.org/info/rfc9101/#RFC7515)] or signed and then encrypted using [JWS](https://www.rfc-editor.org/info/rfc9101/#RFC7515) [[RFC7515](https://www.rfc-editor.org/info/rfc9101/#RFC7515)] and [JWE](https://www.rfc-editor.org/info/rfc9101/#RFC7516) [[RFC7516](https://www.rfc-editor.org/info/rfc9101/#RFC7516)], respectively, with algorithms considered appropriate at the time.

jogu:
This is where we would all like if it RFCs had a nicer errata process :-)

> The text isn't a 100% clear if that sentence is conditional to the first sentence.

My reading is:

require_signed_request_object: true => reject alg none
require_signed_request_object: false => accept alg none

If we instead assume alg: none is disallowed in both cases, what would the meanings of  require_signed_request_object true/false be?

c2bo:
> This is where we would all like if it RFCs had a nicer errata process :-)
> 
> > The text isn't a 100% clear if that sentence is conditional to the first sentence.
> 
> My reading is:
> 
> require_signed_request_object: true => reject alg none require_signed_request_object: false => accept alg none
> 
> If we instead assume alg: none is disallowed in both cases, what would the meanings of require_signed_request_object true/false be?

That is my interpretation as well and if I recall correctly that is also what we ended up with initially when discussing this?

c2bo (approved):

paulbastian:
require_signed_request_object means whether it is required to send authorization request as RFC9101 or whether RFC6749 style is still allowed.

paulbastian (approved):
After re-reading I'm fine

1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -660,7 +660,7 @@ wallet_nonce=qPmxiNFCR3QTm19POc8u
 
 ### Request URI Response
 
-The Request URI response MUST be an HTTP response with the content type `application/oauth-authz-req+jwt` and the body being a signed, optionally encrypted, request object as defined in [@RFC9101]. The request object MUST fulfill the requirements as defined in (#vp_token_request).
+The Request URI response MUST be an HTTP response with the content type `application/oauth-authz-req+jwt` and the body being a request object as defined in [@RFC9101], optionally encrypted. The request object is signed unless the Client Identifier Prefix in use requires an unsigned request (e.g., `redirect_uri`), in which case the JWT uses the `alg` value `none`. The request object MUST fulfill the requirements as defined in (#vp_token_request).
```
dpostnikov: Does "as defined in (#vp_token_request)" imply signing in any way?

PR #790, opened 2026-08-27
Clarify terminate/reject/return error behaviours including for unsupported transaction_data

### LINKED ISSUE #454: Clarify unsupported transaction_data error
Clause [8.4. ](https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#section-8.4)[Transaction Data](https://openid.net/specs/openid-4-verifiable-presentations-1_0-24.html#name-transaction-data) states: "If the wallet does not support transaction_data parameter, it MUST return an error."

Clarify what is the error that needs to be returned when `transaction_data `is not supported.

### LINKED ISSUE #757: Clarification Request: Interpretation of terminate, reject, and return an error in OpenID4VP
The specification uses the terms **terminate**, **reject**, and **return an error** in different contexts. Clarification is requested on the expected behavior and distinction between these terms.

### 1. `terminate`

Reference: [[OpenID4VP Section 5.10.1](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.10.1-5)](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5.10.1-5)

> The Wallet MUST extract the set of Authorization Request parameters from the Request Object. The Wallet MUST only use the parameters in this Request Object, even if the same parameter was provided in an Authorization Request query parameter. The Client Identifier value in the client_id Authorization Request parameter and the Request Object client_id claim value MUST be identical, including the Client Identifier Prefix. If any of these conditions are not met, the Wallet MUST terminate request processing.

Our understanding is that **terminate** represents a hard stop in request processing. In this case, the Wallet does not continue the VP flow and does not notify the Verifier.

### 2. `reject`

Reference: [[OpenID4VP Section 5.8](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5-8)](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-5-8)

> The Wallet MUST ignore any unrecognized parameters, other than the transaction_data parameter. One exception to this rule is the transaction_data parameter. Wallets that do not support this parameter MUST reject requests that contain it.

### 3. `return an error`

Reference: [[OpenID4VP Section 8.4](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-8.4-3)](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html#section-8.4-3)

> If the Wallet does not support transaction_data parameter, it MUST return an error upon receiving a request that includes it.

## Clarification Required

For `terminate`, the expected behavior appears clear: the Wallet stops processing immediately without notifying the Verifier.

However, there is ambiguity regarding the difference between **reject** and **return an error**.

Specifically, for a Wallet that does not support the `transaction_data` parameter:

* Does **reject** mean that the Wallet should stop processing the request locally without notifying the Verifier, or should the Wallet notify the Verifier before ending the VP flow?
* Does **return an error** mean that the Wallet should notify the Verifier through the available error reporting mechanism before ending the VP flow?

For example, if the VP request contains `transaction_data` and uses `direct_post` response mode with a `response_uri`, should the Wallet notify to Verifier:

```http
POST /response_uri

{
  "error": "invalid_transaction_data",
  "error_description": "Transaction data feature is not supported by the Wallet"
}
```

Could you please clarify:

1. The intended distinction between **terminate**, **reject**, and **return an error**.
2. Whether **reject** and **return an error** result in different Wallet behaviors.

### LINKED ISSUE #795: Editorial - Repetition - Needs to be cleaned up.
The Wallet MUST ignore any unrecognized parameters**, other than the transaction_data parameter. One exception to this rule is the transaction_data parameter.** Wallets that do not support this parameter MUST reject requests that contain it.

jogu:
The spec used "terminate", "reject", and "return an error" without defining how they differ, and stated the unsupported transaction_data requirement twice in different words (sections 5.8 and 8.4).

- Define that terminating request processing means no response is returned to the Verifier, as no authentic request (and hence no trusted response endpoint) was obtained.
- Harmonize the unsupported transaction_data wording: section 5 now defers to the Transaction Data section, which specifies that the Wallet must not return a VP Token and that any response returned is an invalid_transaction_data error response. Aborting without a response remains possible per the privacy considerations, resolving the tension between the previous unconditional "MUST return an error" and the SHOULD NOT in the Error Responses privacy section.
- List the unsupported parameter case under invalid_transaction_data.

Fixes #454
Fixes #757
Fixes #795

### DIFF
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -283,8 +283,8 @@ The Verifier communicates a Client Identifier Prefix that indicates how the Wall
 Depending on the Client Identifier Prefix, the Verifier can communicate a JSON object with its metadata using the `client_metadata` parameter which contains name/value pairs.
 
 Additional request parameters, other than those defined in this section, MAY be defined and used, as described in [@!RFC6749].
-The Wallet MUST ignore any unrecognized parameters, other than the `transaction_data` parameter.
-One exception to this rule is the `transaction_data` parameter. Wallets that do not support this parameter MUST reject requests that contain it.
+The Wallet MUST ignore any unrecognized parameters.
+One exception to this rule is the `transaction_data` parameter: a Wallet that does not support this parameter MUST reject requests that contain it, as defined in (#transaction_data).
 
 ## New Parameters {#new_parameters}
 This specification defines the following new request parameters:
@@ -680,6 +680,8 @@ The following is a non-normative example of a payload for a request object:
 The Wallet MUST process the request as defined in [@RFC9101]. The Wallet SHOULD NOT follow HTTP redirects and SHOULD treat them as an error response (see (#http_redirects)). Additionally, if the Wallet passed a `wallet_nonce` in the POST request, the Wallet MUST validate whether the request object contains the respective nonce value in a `wallet_nonce` claim. If it does not, the Wallet MUST terminate request processing.
 
 The Wallet MUST extract the set of Authorization Request parameters from the Request Object. The Wallet MUST only use the parameters in this Request Object, even if the same parameter was provided in an Authorization Request query parameter. The Client Identifier value in the `client_id` Authorization Request parameter and the Request Object `client_id` claim value MUST be identical, including the Client Identifier Prefix. If any of these conditions are not met, the Wallet MUST terminate request processing.
+
+When this specification requires the Wallet to terminate request processing (or terminate the process), the Wallet stops processing the request and does not return an Authorization Response or Authorization Error Response to the Verifier. In these situations, the Wallet was unable to obtain an authentic Authorization Request, so there is no trusted endpoint to which an Authorization Error Response could be sent.
 
 The Wallet then validates the request as specified in OAuth 2.0 [@RFC6749].
 
@@ -1490,7 +1492,7 @@ The transaction data mechanism enables a binding between the user's identificati
 
 The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective Credential presentation. How this is done is transaction data type specific. Credential Formats can give recommendations of how to handle transaction data, such as those in (#format_specific_parameters).
 
-If the Wallet does not support `transaction_data` parameter, it MUST return an error upon receiving a request that includes it.
+If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead abort processing without returning a response to the Verifier.
 
 ## Error Response {#error-response}
 
@@ -1526,6 +1528,7 @@ This document also defines the following additional error codes and error descri
 
 `invalid_transaction_data`:
 
+- the Wallet does not support the `transaction_data` parameter, or
 - any of the following is true for at least one object in the `transaction_data` structure:
   - contains an unknown or unsupported transaction data type value,
   - is an object of a known type but containing unknown fields,
@@ -3660,6 +3663,8 @@ The technology described in this specification was made available from contribut
 
 -31
 
+   * Clarify that a Wallet terminating request processing does not return a response to the Verifier
+   * Clarify that a Wallet rejecting a request containing an unsupported `transaction_data` parameter uses the `invalid_transaction_data` error code if it returns an error response
    * add security considerations on untrusted input
    * Clarify that the Wallet does not follow HTTP redirects
    * Clarify jwks use parameter
--- a/1.0/openid-4-verifiable-presentations-1_0.md
+++ b/1.0/openid-4-verifiable-presentations-1_0.md
@@ -1492,7 +1492,7 @@ The transaction data mechanism enables a binding between the user's identificati
 
 The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective Credential presentation. How this is done is transaction data type specific. Credential Formats can give recommendations of how to handle transaction data, such as those in (#format_specific_parameters).
 
-If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead abort processing without returning a response to the Verifier.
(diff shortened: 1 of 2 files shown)


1.0/openid-4-verifiable-presentations-1_0.md
```diff
@@ -1490,7 +1492,7 @@ The transaction data mechanism enables a binding between the user's identificati
 
 The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective Credential presentation. How this is done is transaction data type specific. Credential Formats can give recommendations of how to handle transaction data, such as those in (#format_specific_parameters).
 
-If the Wallet does not support `transaction_data` parameter, it MUST return an error upon receiving a request that includes it.
+If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead abort processing without returning a response to the Verifier.
```
javereec: Nitpicking here, but should you use terminate in the last sentence to make it consistent with what you wrote earlier?


```suggestion
If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead terminate the request processing.
```

javereec (commented):
Looks good, just one small suggestion.

brentzundel:
discussed today: needs review.

1.0/openid-4-verifiable-presentations-1_0.md
jogu: I think you're right that my suggested wording isn't fully resolving the issue - but terminate is defined quite specifically to when there's no trusted endpoint to send a response to. So it's probably better to align with the 'can opt to cancel the flow' style wording we have in https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_0-wg-draft.html#section-15.9-7


```suggestion
If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead cancel the flow without returning a response to the Verifier.
```

I'll apply this for now but let me know if you think terminate is better.

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1550,7 +1552,7 @@ The transaction data mechanism enables a binding between the user's identificati
 
 The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective Credential presentation. How this is done is transaction data type specific. Credential Formats can give recommendations of how to handle transaction data, such as those in (#format_specific_parameters).
 
-If the Wallet does not support `transaction_data` parameter, it MUST return an error upon receiving a request that includes it.
+If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead abort processing without returning a response to the Verifier.
```
jogu: Same 'cancel the flow' wording  as 1.0 suggestion:

```suggestion
If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead cancel the flow without returning a response to the Verifier.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -1550,7 +1552,7 @@ The transaction data mechanism enables a binding between the user's identificati
 
 The Wallet that received the `transaction_data` parameter in the request MUST include a representation or reference to the data in the respective Credential presentation. How this is done is transaction data type specific. Credential Formats can give recommendations of how to handle transaction data, such as those in (#format_specific_parameters).
 
-If the Wallet does not support `transaction_data` parameter, it MUST return an error upon receiving a request that includes it.
+If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet can instead cancel the flow without returning a response to the Verifier.
```
fkj, line 1555: I think this wording is actually normative, so we might as well make it explicit:

```suggestion
If the Wallet does not support the `transaction_data` parameter, it MUST reject a request that includes it: the Wallet MUST NOT return a VP Token for such a request, and any response returned MUST be an error response using the error code `invalid_transaction_data` (see (#error-response)). As described in (#error-responses), the Wallet MAY instead cancel the flow without returning a response to the Verifier.
```

1.1/openid-4-verifiable-presentations-1_1.md
```diff
@@ -677,6 +677,8 @@ The Wallet MUST process the request as defined in [@RFC9101]. The Wallet SHOULD
 
 The Wallet MUST extract the set of Authorization Request parameters from the Request Object. The Wallet MUST only use the parameters in this Request Object, even if the same parameter was provided in an Authorization Request query parameter. The Client Identifier value in the `client_id` Authorization Request parameter and the Request Object `client_id` claim value MUST be identical, including the Client Identifier Prefix. If any of these conditions are not met, the Wallet MUST terminate request processing.
 
+When this specification requires the Wallet to terminate request processing (or terminate the process), the Wallet stops processing the request and does not return an Authorization Response or Authorization Error Response to the Verifier. In these situations, the Wallet was unable to obtain an authentic Authorization Request, so there is no trusted endpoint to which an Authorization Error Response could be sent.
```
fkj, line 680: I think this language is actually normative, so might as well make it explicit:
```suggestion
When this specification requires the Wallet to terminate request processing (or terminate the process), the Wallet stops processing the request and MUST NOT return an Authorization Response or Authorization Error Response to the Verifier. In these situations, the Wallet was unable to obtain an authentic Authorization Request, so there is no trusted endpoint to which an Authorization Error Response could be sent.
```

fkj (commented):
Looks fine except it plays a little loose with normative wording. My suggestions should obviously also be fixed in the errata if they are accepted.

brentzundel:
Discussed today. @jogu will respond too @fkj , @fkj will then re-view. @c2bo and @GarethCOliver and Jim Richards will review

brentzundel:
Discussed in call. Ready for review from @c2bo @fkj @GarethCOliver

PR #798, opened 2026-09-17
fix JAR fallback text

### LINKED ISSUE #779: "Wallet MAY continue with JAR" is contradicted later in the spec and intended to be a MUST
https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#name-authorization-request says:

> The Wallet MAY continue with JAR when it receives request_uri_method parameter with the value post but does not support this feature.

I don't think this is correct; this is actually a MUST I believe - https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#name-new-parameters says:

> Wallets not supporting the post method will send a GET request to the Request URI (default behavior as defined in [[RFC9101](https://openid.github.io/OpenID4VP/openid-4-verifiable-presentations-1_1-wg-draft.html#RFC9101)]).

I'd be tempted to put that 'will' as a MUST too, but I think it's probably clear as is. Deleting 'will' would work too.

I suggest changing the 'MAY continue with JAR' to a 'MUST' (or possibly removing the 'MAY' from the sentence, or just removing the sentence entirely.)

The certification tests treat the 'MAY' as a 'MUST'.

c2bo:
Fixes #779 

Added it to the 1.0 errata as well


brentzundel:
Discussed to day. @jogu  and @awoie will review

jogu (approved):
There's a little bit of a breaking change around treating unknown values as GET that's a little bit dicey (in terms of whether we can included it in the conformance tests for wallets that are following 1.0 without errata) but overall I think it's probably best to make the change.

awoie (approved):
LGTM, although the change could be considered breaking. However, it is probably what wallets have implemented already, so it should be fine.

dpostnikov (approved):

brentzundel:
Sufficient review time, approved, no requests for changes, merging

